PR #1, opened 2018-05-18
Strip leading null byte from public key when unwrapping key

frasertweedale:
NSS identifies RSA private keys by setting the PKCS #11 CKA_ID
attribute to the SHA-1 digest of the public key (modulus).

PK11KeyWrapper can receive this modulus value (a byte[]) with a
leading null byte.  This results in the digest (CKA_ID) not matching
what the rest of NSS expects, e.g. when adding the corresponding
certificate, NSS fails to associate it with the private key, because
it is looking for a different CKA_ID.

This results in Dogtag lightweight CA key replication failures.

Apparently the problem did not occur with the old DB backend, only
with the new SQL backend.  Or there was some other change in NSS
that landed in Fedora 28, which prompted this issue.  In either
case, the resolve the problem by dropping the leading null byte from
the modulus byte[].

Fixes: https://pagure.io/jss/issue/5
Change-Id: I4685d7e091b0adc72d5ca067f2d65c3c068c8f7a

### DIFF
--- a/org/mozilla/jss/pkcs11/PK11KeyWrapper.java
+++ b/org/mozilla/jss/pkcs11/PK11KeyWrapper.java
@@ -4,6 +4,8 @@
 
 package org.mozilla.jss.pkcs11;
 
+import java.util.Arrays;
+
 import org.mozilla.jss.crypto.*;
 import java.security.spec.AlgorithmParameterSpec;
 import java.security.InvalidAlgorithmParameterException;
@@ -394,6 +396,11 @@ final class PK11KeyWrapper implements KeyWrapper {
         }
 
         byte[] publicValue = extractPublicValue(publicKey, type);
+        /* If first byte is null, omit it.
+         * It can be null due to how BigInteger.toByteArray() is specified. */
+        if (publicValue.length > 0 && publicValue[0] == 0) {
+            publicValue = Arrays.copyOfRange(publicValue, 1, publicValue.length);
+        }
 
         if( symKey != null ) {
             Assert._assert(pubKey==null && privKey==null);


tiran:
Is it always safe to remove a leading NULL byte? What happens if the ``CKA_ID`` is already correct but coincidentally starts with a NULL byte? The case can occur one out of 256 times. Or does NSS always remove the first leading byte no matter what?

frasertweedale:
@tiran this function strips the leading NULL byte from the modulus *input value* (which I think is DER-encoded?), not from the resulting digest which is used as the CKA_ID.

tiran:
* ``extractPublicValue()`` calls ``((RSAPublicKey)publicKey).getModulus().toByteArray()``
* ``getModulus`` is inherited from ``RSAKey`` and returns a ``BigInteger``, 
https://docs.oracle.com/javase/7/docs/api/java/security/interfaces/RSAKey.html#getModulus()
* ``toByteArray()`` returns a big endian representation, https://docs.oracle.com/javase/7/docs/api/java/math/BigInteger.html#toByteArray()

The patch may chop off the least significant byte. That looks ... strange.

frasertweedale:
Ah yeah, that's right, big-endian int.  I believe it is a quirk of `BigInteger.toByteArray()`.  One question I still have (yet to verify) is if it is a change in Java `BigInteger` behaviour that occurred in f28, or if (more likely) the Java behaviour did not change but instead, NSS when using old DB backend, and/or NSS pre-f28, automatically skipped/ignored a leading null byte when computing the digest of the modulus, and now (as of f28/SQL NSSDB) it does not.

I think I need to more thoroughly investigate this...

tiran:
aaah, I double reversed the polarity of the big endian representation. It's fine and safe to remove a leading NULL byte. Is it always just one or could there be more NULL bytes?

frasertweedale:
@tiran I'll read the openjdk source code and find out :)

frasertweedale:
@tiran more info:

First, it is definitely a discrepancy between dbm and sql backend behavoiur.
In brief, the dbm backend recomputes the CKA_ID using the modulus of the key that
was actually imported, and the SQL backend does not.  More details in BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1583027

Second, OpenJDK's `BigInteger.toByteArray()` behaviour: a *single* leading zero byte is included
for non-negative integers whose topmost bit in the big-endian bytearray of the two's complement representation of the number is set.  This is to distinguish e.g. `{0, 255}` = 255 from `{255}` = -1.

Because the modulus values should always be positive and there is always at most one leading byte based on the implementation of `BigInteger`, the current fix (strip leading zero-byte if present) seems adequate.

PR #5, opened 2018-05-31
Updated README instructions for git

cipherboy:
README instructions still used mercurial for updating the repository and tagging new releases. It appears that the new standard for tags/branches is `v<major>.<minor>.<patch>`, so perhaps all of section 7 should be updated. I left the ids/versions/tag messages untouched to minimize changes.

### DIFF
--- a/README
+++ b/README
@@ -17,7 +17,7 @@
         (There is no need to clone every time. For additional builds,
          simply use:
           cd nspr; hg pull -u -v; cd ..; cd nss; hg pull -u -v; cd ..;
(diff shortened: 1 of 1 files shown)


edewata:
Thanks for the submission!

Just one thing, according to the man page the -u option for git pull would be redundant, but it wouldn't hurt either, so it's ACKed.

We are indeed using a new tag format based on the examples in this page:
https://git-scm.com/book/en/v2/Git-Basics-Tagging
So yes, we should update the documentation later.

PR #6, opened 2018-06-01
 Updated README

edewata:


mharmsen99:
NACK

Removing the numbered sections (1) .. (8) messes up the instructions which reference these sections by their old numerical designations.

edewata:
Patch updated. Please take a look. Thanks.

cipherboy:
@edewata Two thoughts:

I like that you largely left the README alone and did minimal changes. However, this resulted in nearly everything being a code block due to the four spaces indentation. I'd recommend pulling at least the large text block in section 8 out of the code block. (I recognize that pulling more out can damage the flow of the if/else statements in sections 2, 4, and 6, but 8 and some of the longer notes could be unindented).  

Section 1 links to issues being on Pagure; I'm not sure the intent of having issues on Pagure instead of GH, but I'd suggest disabling GH issues for simplicity.

edewata:
@cipherboy, thanks for the feedback! I've made some additional changes. Please take a look at the new patch.

When fedoraproject.org was about to be retired, there were discussions about where to host PKI-related projects, their issue trackers, how to do code review, etc. Due to limited experience with either system, we ended up with the current hybrid solution with some in GH and some in Pagure. But this configuration is not final, we can certainly make some changes (e.g. moving the tickets to GH) if there are strong arguments for it, especially after using it for some time. We can discuss this further on the mailing list or IRC.

cipherboy:
@edewata Thanks! LGTM :)

edewata:
Thanks! It has been merged.

PR #7, opened 2018-06-12
Added PK11RSAPrivateKey.getModulus()

edewata:
The PK11RSAPrivateKey.getModulus() has been modified to get the
modulus from the public key, which was converted from the private
key using SECKEY_ConvertToPublicKey().

A new PK11Exception has been added as a generic exception for all
PKCS #11-related errors.

https://pagure.io/jss/issue/9


org/mozilla/jss/pkcs11/PK11PrivKey.c
```diff
@@ -665,3 +665,64 @@ Java_org_mozilla_jss_pkcs11_PK11PrivKey_getDSAParamsNative
...
+    // check return code
+    if (rv != PR_SUCCESS) {
+        char *message = PR_smprintf("Unable to get RSA private key (rc: %d)", rv);
+        JSS_throwMsg(env, PK11_EXCEPTION, message);
+        PR_smprintf_free(message);
+        goto finish;
+    }
+
+    // get slot from private key
+    slot = PK11_GetSlotFromPrivateKey(privateKey);
```
jmagne: I think you need to free that slot object: See here: https://dxr.mozilla.org/security/source/mozilla/security/nss/lib/ssl/ssl3con.c#5359

PR #8, opened 2018-06-13
Fixed JSSKeyStoreSpi.engineLoad()

edewata:
The JSSKeyStoreSpi.engineLoad() has been modified to store the
token provided via JSSLoadStoreParameter in the keystore. The
token later will be used to execute various operations in the
keystore (i.e. retrieving certs and keys from the token).

### DIFF
--- a/org/mozilla/jss/provider/java/security/JSSKeyStoreSpi.java
+++ b/org/mozilla/jss/provider/java/security/JSSKeyStoreSpi.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.security.Key;
+import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.cert.Certificate;
 import java.security.cert.CertificateEncodingException;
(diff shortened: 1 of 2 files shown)


cipherboy:
@edewata / @jmagne -- As I get familiar with this project, is the `engineLoad(InputStream stream, char[] password)` function call correct? It looks like it ignores both of its parameters; I personally find this surprising behavior, considering that `engineLoad(KeyStore.LoadStoreParameter param)` extracts and saves a token parameter.

edewata:
@jmagne, thanks!

@cipherboy, here is some documentation:
http://www.dogtagpki.org/wiki/JSS_KeyStore

Here is a code in PKI that will use the JSS code:
https://github.com/edewata/pki/blob/pkcs11/base/java-tools/src/com/netscape/cmstools/pkcs11/PKCS11CertFindCLI.java#L74

The KeyStore class supports 2 different load() methods. The one you mentioned is for loading the keystore from an input stream which might be protected with a password, which might be appropriate for PKCS #12 keystore. For PKCS #11, we're not loading the certs/keys from HSM token, but we're only telling the keystore which token to find the certs/keys. I'll have a patch for that later.

edewata:
@cipherboy, thanks!

Merged.

PR #9, opened 2018-06-14
Fixing JSSKeyStoreSpi

edewata:
The JSSKeyStoreSpi.engineAliases() has been modified to return
all aliases from all tokens if the keystore token is not set,
and only aliases from the keystore token if it's set.


edewata:
Thanks! Merged.

PR #12, opened 2018-06-22
Fixed JSSKeyStoreSpi.engineGetKey()

edewata:
The JSSKeyStoreSpi.engineGetKey() has been rewritten in Java to
search for a private key by its cert. If that is not found, it
will search for a private key by its key ID. If that is not found
either, it will search for a symmetric key by its nickname.


edewata:
Thanks!

Merged.

PR #17, opened 2018-07-12
Updated JSSKeyStoreSpi.engineDeleteEntry()

edewata:
Updated JSSKeyStoreSpi.engineDeleteEntry()

The JSSKeyStoreSpi.engineDeleteEntry() has been modified such that:
* if the given alias is a cert nickname, it will delete the cert without deleting the public and private keys
* if the given alias is a key ID, it will delete the public and private keys without deleting the cert

### DIFF
--- a/org/mozilla/jss/pkcs11/PK11Store.c
+++ b/org/mozilla/jss/pkcs11/PK11Store.c
@@ -436,6 +436,47 @@ finish:
 }
 
 /**********************************************************************
+ * PK11Store.deletePublicKey
+ */
+JNIEXPORT void JNICALL
+Java_org_mozilla_jss_pkcs11_PK11Store_deletePublicKey
+    (JNIEnv *env, jobject this, jobject publicKeyObj)
+{
+    PK11SlotInfo *slot;
+    SECKEYPublicKey *publicKey;
+
+    PR_ASSERT(env != NULL && this != NULL);
+
+    if (publicKeyObj == NULL) {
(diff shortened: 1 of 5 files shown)


org/mozilla/jss/pkcs11/PK11Store.c
```diff
@@ -435,6 +435,47 @@ Java_org_mozilla_jss_pkcs11_PK11Store_deletePrivateKey
...
+JNIEXPORT void JNICALL
+Java_org_mozilla_jss_pkcs11_PK11Store_deletePublicKey
+    (JNIEnv *env, jobject this, jobject publicKeyObj)
+{
+    PK11SlotInfo *slot;
+    SECKEYPublicKey *publicKey;
+
+    PR_ASSERT(env != NULL && this != NULL);
+
+    if (publicKeyObj == NULL) {
```
jmagne: Question. NO big deal, but is that really the exception to use? Technically it's true that "nothing" doesn't exist on on the token just wondering if we have any argument based exceptions?

edewata: This code was based on Java_org_mozilla_jss_pkcs11_PK11Store_deletePrivateKey() which uses the same exception for null private key (see line 401). We could certainly change this. Would NPE be better? I don't think exception argument is necessary since the argument will be null.

jmagne: I don't think this is important enough to mess with. I will approve, since it's based on precedent.

edewata:
Thanks!

PR #23, opened 2018-08-16
Complete deprecation annotations already started

emaldona:
This is for https://pagure.io/jss/issue/19.
It affects many files but only in the comments as the actual code isn't touched at all.
Given the size it may be wise to wait until other simpler pull requests have been dealt with before reviewing this one.


emaldona:
Failure due to incorrect annotation in two files, need to change @deprecation to @deprecated.

PR #27, opened 2018-09-18
Reflect dependencies in README

cipherboy:
I noticed that dependency information was missing from the README and thought it was useful to add.

On Fedora, there's also the `dnf builddep jss`, but I'm not sure if that'd be desired instead of the explicit command.

Interestingly, `dnf builddep jss` fails to pull in `zlib-devel`, as it is missing from the spec file. However, without `zlib-devel`, `nss` fails to build. This is only required if building with (1)(a) instructions though.


cipherboy:
(I updated this to reflect `java-devel` dependency and change formatting)

cipherboy:
As this does not affect the code, I'm merging.

PR #28, opened 2018-09-18
Align jss with make standard targets

cipherboy:
A common pattern across projects is to use `make check` to execute
tests, `make dist` to build a release, and `make distclean` to
clean the tree after a release. This adds targets to `rules.mk`
which support these targets and moves a target from the `Makefile`
to `rules.mk`.

This help aligns the project with the GNU coding standards: https://www.gnu.org/prep/standards/html_node/Standard-Targets.html

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,8 +34,6 @@ include $(CORE_DEPTH)/coreconf/config.mk
 # built first
 all:: buildJava
 
-test_jss: testJava
-
 include $(CORE_DEPTH)/coreconf/rules.mk
 
 #######################################################################
--- a/rules.mk
+++ b/rules.mk
@@ -3,8 +3,14 @@
 .PHONY: releaseJava
 .PHONY: testJava
 
+distclean:: cleanJava
 clean:: cleanJava
+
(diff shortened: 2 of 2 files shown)


cipherboy:
@jmagne Any preferences on this?

jmagne:
This all looks good. I just want to confirm that the previous targets are still there in case there is stuff out there using them, thus we are just adding targets to comply with the standards.

cipherboy:
Yep, they're all still there. I moved `test_jss` from `Makefile -> rules.mk` (which is imported from `Makefile`), and then moved `clean` down 4 lines to group it a little better, but I might revert that and keep the order the same. 

No other changes were made besides adding 3 new targets (`dist`, `distclean`, and `check`).

There is no `installcheck` target as we don't have rules for testing an installed JSS. (`distcheck` isn't a standard target).

cipherboy:
Okay this is a nicer diff and better ordering (`clean -> build -> test`). Pending CI, I'll merge.

Thanks for the comments, @jmagne!

PR #22, opened 2018-08-16
duplicate name in manifest warnings

emaldona:
For https://pagure.io/jss/issue/18

PR #30, opened 2018-09-19
build_java.pl improvements

cipherboy:
A few improvements to the `build_java.pl` script: 

 - Linting it with Perl `-Mstrict` flag
    - New tool introduced in `tools/check_perl_style.sh` to validate this.
 - Remove the unnecessary `eval` just to call a function inside the script.
 - Removed the logic around appending to manifest.
 - Simplified the logic around the test suite to avoid nested `if` statements.

(The `B::Lint` module was giving what I believe to be a false positive and wouldn't install cleanly from `cpan` so I decided against it). 

The second change introduces a hash to map `name->subroutine`; this would need to be kept in sync if more commands are introduced, but I don't think this is likely. Secondly, it drops passing `ARGV` into the function. This would be easy to add, but I don't believe any of the functions use this behavior.

### DIFF
--- a/build_java.pl
+++ b/build_java.pl
@@ -474,26 +474,24 @@ sub test {
     our $nss_bin_dir;
     our $nss_lib_dir;
 
-    if( $os eq 'Linux' || $os eq 'Darwin' ) {
-        # Test JSS presuming that it has already been built
-
-        if(( -d $dist_dir )  &&
-           ( -d $jss_objdir || -l $jss_objdir )) {
-            my $cmd = "cd $jss_dir/org/mozilla/jss/tests;"
-                    . "perl all.pl dist \"$dist_dir\" \"$nss_bin_dir\" \"$nss_lib_dir\" \"$jss_lib_dir\";"
-                    . "cd $jss_dir";
-
-            print("#######################\n" .
-                  "# BEGIN:  Testing JSS #\n" .
-                  "#######################\n");
-            print_do($cmd);
-            print("#####################\n" .
-                  "# END:  Testing JSS #\n" .
-                  "#####################\n");
-        } else {
-            die "JSS builds are not available at $jss_objdir.";
-        }
-    } else {
+    if( $os ne 'Linux' && $os eq 'Darwin' ) {
         die "make test_jss is only available on Linux and MacOS platforms.";
     }
+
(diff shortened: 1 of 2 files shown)


cipherboy:
I'm marking this WIP for the time being until I fix the issue discovered in #22.

tools/test_perl_style.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
```
cipherboy: @jmagne or @edewata - Could I get review on this?

In particular: I'm introducing a new directory, `tools`. Is there anything special I need to do with respect to the spec file? I verified that it doesn't show up in the built RPM but not that it didn't get added to the jar archive.

cipherboy:
Removing WIP flag as it is ready for review.

build_java.pl
```diff
@@ -138,60 +148,59 @@ sub setup_vars {
...
     if( $ENV{BUILD_OPT} ) {
         $javac_opt_flag = "-O";
-    } else {
-        $javac_opt_flag = "-g";
     }
-    $jni_header_dir = "$dist_dir/private/jss/_jni";
 
+    our $jni_header_dir = "$dist_dir/private/jss/_jni";
+
+    our$jarFiles = "/usr/share/java/slf4j/slf4j-api.jar:/usr/share/java/commons-codec.jar";
```
jmagne: Do we care that there is no space between our and the var?

cipherboy: Heh good catch. It looks like Perl doesn't care about spacing around scalars:

```perl
my$var="testing\n";
print$var;
```

prints:
```
testing
```

So go figure. I'll update this when I get time.

cipherboy:
Thanks for the comments @jmagne! I'll review this on Wednesday when I get back and double check this. I'll validate the jar as well. I'll also try binary diffing the contents as well to make sure the built artifacts didn't change as a result too.

cipherboy:
Interestingly, Travis failed to push status back, but it shows as all green on the website:

https://travis-ci.org/dogtagpki/jss/builds/433545305?utm_source=github_status&utm_medium=notification

screenshot from 2018-09-26 10-44-24

cipherboy:
Thanks @jmagne! :) Merging...

PR #31, opened 2018-10-01
Cleanup all.pl to address test_perl_style.sh tool messages

emaldona:
- Address https://pagure.io/jss/issue/23

- perl -Mstrict -Mdiagnostics -cw org/mozilla/jss/tests/all.pl
- now shows org/mozilla/jss/tests/all.pl syntax OK


cipherboy:
Okay, I have added another commit that adds this to the `tools/test_perl_style.sh` script. Simpler than creating another PR or having you update it. :)

Pending CI, I'll merge.

I've validated:
 - That this still runs the tests and doesn't affect their output
 - That intentionally making the tests fail does indeed cause failure in the resulting script.
 - That the changes look sane.


Thanks @emaldona! :)

PR #32, opened 2018-10-03
Remove dependency on javax.xml.bind.DatatypeConverter

cipherboy:
This is no longer present in JDK 9 and greater; it has been moved
to Java EE instead. Remove this dependency and rely on existing
code instead.


cipherboy:
+1 Travis... This has passed on Travis and addresses the issues from @edewata.

PR #34, opened 2018-10-04
Update travis.yml to check style

cipherboy:
Two improvements:
 - Validate style before building RPMs.
 - Build RPMs in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,25 +3,44 @@
 # All rights reserved.
 # END COPYRIGHT BLOCK
 
+language: java
+
 services:
   - docker
 
-env:
-  - FEDORA=27
-  - FEDORA=28
+stages:
+  - validate
+  - build_rpm
 
-install:
-  - docker pull registry.fedoraproject.org/fedora:$FEDORA
-  - docker run
-      --name=container
-      --detach
-      -i
-      -v $(pwd):/root/jss
-      registry.fedoraproject.org/fedora:$FEDORA
-  - docker exec container dnf install -y dnf-plugins-core gcc make rpm-build
-  - docker exec container dnf copr -y enable ${JSS_4_5_REPO:-@pki/10.6}
-  - docker exec container dnf builddep -y --spec /root/jss/jss.spec
-  - docker exec container /root/jss/build.sh --with-timestamp --with-commit-id rpm
+jobs:
+  include:
+    # Validate style of scripts
+    - stage: validate
+      install:
+        - docker pull registry.fedoraproject.org/fedora:28
+        - docker run --name=validate --detach -i -v $(pwd):/root/jss registry.fedoraproject.org/fedora:28
+        - docker exec validate dnf install -y perl ShellCheck
+      script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -3,25 +3,62 @@
...
-script:
-  - docker exec container rpm -Uvh /root/build/jss/RPMS/*
+stages:
+  - validate
+  - build_rpm
+
+jobs:
+  include:
+    # Validate style of scripts
+    - stage: validate
```
SilleBille: How about moving this stage to existing scripts and keeping the matrix? -- The matrix is to avoid rewriting the same set of scripts :)

.travis.yml
```diff
@@ -3,25 +3,44 @@
...
-      registry.fedoraproject.org/fedora:$FEDORA
-  - docker exec container dnf install -y dnf-plugins-core gcc make rpm-build
-  - docker exec container dnf copr -y enable ${JSS_4_5_REPO:-@pki/10.6}
-  - docker exec container dnf builddep -y --spec /root/jss/jss.spec
-  - docker exec container /root/jss/build.sh --with-timestamp --with-commit-id rpm
+    # Build RPMs on Fedora
+    - stage: build_rpm
+      env:
+        - FEDORA=27
+        - FEDORA=28
```
cipherboy: So it looks like this approach doesn't work as it first sets F27 and then F28 and then runs the build (i.e., for the last one set which is F28)... So that's sad. :/

cipherboy:
I'll open a new PR with Dockerfiles and perhaps build scripts; I'd like to support non-RHEL-like builds upstream as well.

SilleBille:
@cipherboy  here's my2c: We had maintained docker images earlier for our project. The problem was, you need to maintain them quite intensively. They get outdated very soon (ie) few components included in the docker image will become old. That's the reason we moved to vanilla Docker images -- Use the Fedora images from official fedora registry and install all the deps required on runtime. This would save valuable time of yours and maintainer on cost of increased travis build time , which is ~2 mins (not bad, right?)

cipherboy:
@SilleBille On the other hand, maintaining Dockerfiles is equivalent work to maintaining the `.travis.yml` as it accomplishes the same purpose. You can run Dockerfiles from Travis without having to pull remote non-Fedora images; see @ComplianceAsCode's `content` repository for example:

Traivs.yml is much simplified compared to ours:
 - https://github.com/ComplianceAsCode/content/blob/master/.travis.yml 

And a Dockerfile contains all of the logic and easily allows a developer to run Travis checks locally:
 - https://github.com/ComplianceAsCode/content/blob/master/Dockerfiles/fedora_python3
 - It uses the Fedora images and installs all the deps on container build/run. 

No maintaining docker images on a Dockerhub-like-platform necessary. :-)

SilleBille:
@cipherboy  ah! That makes sense. Yeah, seems like a better approach. I thought something like building images on Dockerhub and pulling them. Maybe we can start with JSS before we try to implement in our pki repo! :) Thanks for the idea! :+1:

PR #35, opened 2018-10-09
Update travis.yml to check style and use stages

cipherboy:
- Adds style check / validator stage to check perl and
     script styles
 - Uses Dockerfiles to perform all tests
 - Adds check to ensure Ubuntu build is successful

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,25 +3,31 @@
 # All rights reserved.
 # END COPYRIGHT BLOCK
 
+language: java
+
 services:
   - docker
 
-env:
-  - FEDORA=27
-  - FEDORA=28
+stages:
+  - validate
+  - test
 
-install:
-  - docker pull registry.fedoraproject.org/fedora:$FEDORA
-  - docker run
-      --name=container
-      --detach
-      -i
-      -v $(pwd):/root/jss
-      registry.fedoraproject.org/fedora:$FEDORA
-  - docker exec container dnf install -y dnf-plugins-core gcc make rpm-build
-  - docker exec container dnf copr -y enable ${JSS_4_5_REPO:-@pki/10.6}
-  - docker exec container dnf builddep -y --spec /root/jss/jss.spec
-  - docker exec container /root/jss/build.sh --with-timestamp --with-commit-id rpm
+env:
+  - BASE_IMAGE="fedora_27"
+  - BASE_IMAGE="fedora_28"
+  - BASE_IMAGE="ubuntu_jdk8"
+
+before_install:
+  - docker build --no-cache --tag jss_$BASE_IMAGE:latest -f tools/Dockerfiles/$BASE_IMAGE .
 
 script:
-  - docker exec container rpm -Uvh /root/build/jss/RPMS/*
+  - docker run jss_$BASE_IMAGE:latest
+
+matrix:
+  fast_finish: true
+  include:
+    - stage: validate
+      env: BASE_IMAGE="stylecheck"
+  allow_failures:
+    - env: BASE_IMAGE="ubuntu_jdk8"
+
--- a/tools/Dockerfiles/stylecheck
+++ b/tools/Dockerfiles/stylecheck
@@ -1,0 +1,22 @@
+FROM fedora:latest
+
+# Install generic dependencies to check style
+RUN true \
+        && dnf update -y --refresh \
+        && dnf install -y perl ShellCheck \
+        && mkdir -p /home/sandbox \
+        && dnf clean -y all \
+        && rm -rf /usr/share/doc /usr/share/doc-base \
+                  /usr/share/man /usr/share/locale /usr/share/zoneinfo \
+        && true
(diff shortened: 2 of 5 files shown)


cipherboy:
@SilleBille Hmmm the Ubuntu build takes forever on Travis and is quicker (but still slower than the RPM build because it has to build NSS & NSPR).

Thoughts? I typically use `ccache` to speed up compilation locally... I think you could do that with Docker and Travis but it could be a bit of work.

tools/Dockerfiles/ubuntu_jdk8
```diff
@@ -0,0 +1,37 @@
+FROM ubuntu:latest
```
SilleBille: Do you want to have a ubuntu image pulled? Travis runs on Ubuntu 14.04 (stable) version. They usually run their infra on a LTS version. Why not use the Travis' underlying ubuntu directly?

tools/Dockerfiles/stylecheck
```diff
@@ -0,0 +1,22 @@
...
+
+# Link in the current version of jss from the git repository
+WORKDIR /home/sandbox
+COPY . /home/sandbox/jss
+
+# Perform the style checks
+WORKDIR /home/sandbox/jss
+CMD true \
+        && bash ./tools/test_perl_style.sh \
+        && bash ./tools/test_shell_style.sh \
```
SilleBille: Is there no building and installing of JSS rpms in latest Fedora images?

.travis.yml
```diff
@@ -3,25 +3,47 @@
...
+stages:
+  - validate
+  - build
+
+jobs:
+  include:
+    # Validate style of scripts
+    - stage: validate
+      env:
+        - BASE_IMAGE="stylecheck"
```
SilleBille: Is it possible to use the "Build stage Matrix Expansion" feature? https://docs.travis-ci.com/user/build-stages/matrix-expansion/ and https://docs.travis-ci.com/user/build-stages#build-stages-and-build-matrix-expansion

I haven't myself implemented it in the `pki` repo, but I would love to. This feature was released recently. Since different jobs in the `build` stage have the same exact script, how about trying this out and keeping the .travis.yml short and easy maintenance?

**Note:** If you decide to implement this feature, I'm not sure whether the first stage can be named as `validate`. It defaults to `Test` as mentioned in the examples

tools/Dockerfiles/stylecheck
```diff
@@ -0,0 +1,22 @@
+FROM fedora:latest
```
SilleBille: Do you wanna run the style check from rawhide -- some other unrelated packages might be broken? I'm not sure whether `latest` refers to 28 or 29 or rawhide.

tools/Dockerfiles/ubuntu_jdk8
cipherboy: Because Ubuntu LTS 14.04 is 4 years out of date at this point. 16.04 and 18.04 are both LTS versions that are newer than 14.04. Also, 14.04.5 LTS EOLs on April 2019. `ubuntu:latest` tracks the latest LTS version of Ubuntu (18.04 -> 20.04 in 2 years, etc). 

This allows:
- Us to test locally w/o travis on a common platform (so if branch `xyz` fails this Ubuntu test on Travis, you can run with the exact same environment locally -- this is what containers gives us).
- Us to change the test (or add a new test) to support JDK 11 which won't ship on 14.04.
- Us to test other (fixed) versions of Ubuntu if we want. (18.10 when/before it ships, 19.04, etc. if we want to track testing versions of Ubuntu).

cipherboy: (See https://hub.docker.com/_/ubuntu/ for more information)

tools/Dockerfiles/stylecheck
cipherboy: No, this script is only for checking style. `latest` refers to the latest release by Fedora project, so this is (currently) 28. See https://hub.docker.com/_/fedora/ for more information about Docker's Fedora images. Since our style checks don't really care about the version of the checkers, and it'd be nice to not update them every time a new release happens, I'm fine using `latest`. And no, hopefully 28 won't have broken packages and (hopefully) when 29 ships, it won't have broken packages either. :-)


`rawhide` is a different tag.

.travis.yml
cipherboy: I'll have to take a look at that and get back to you on that. But looks interesting. :)

cipherboy:
@SilleBille updated with build matrix support. There's one item in the `validate` group and then three items in the `test` group which it places after the validate group. I think we're good to go on this PR then.

cipherboy:
Okay, I added fast finish and allowed to fail on the Ubuntu job; this means that PR status is updated based on Fedora jobs and that the Ubuntu job will still run but won't block status updates.

cipherboy:
Thanks for the review, @SilleBille!

PR #36, opened 2018-10-09
Add documentation about JSS

cipherboy:
This is the initial step in improving documentation about JSS. In particular, this PR documents two aspects:

 - The build system and the various parts of it. It will be expanded as more information is acquired.
 - Using JSS in an external project. This documents the CLASSPATH dependencies of JSS.

The latter is important as [389-console](https://pagure.io/389-console) appears to be using JSS [without including `slf4j-api.jar` in the classpath](https://pagure.io/389-console/blob/master/f/build.xml#_39).

### DIFF
--- a/docs/build_system.md
+++ b/docs/build_system.md
@@ -1,0 +1,72 @@
+# Build System
+
+There are three interfaces to the build system:
+
+ - `build_java.pl`
+ - `build.sh`
+ - `Makefile`
+
+Each of these are described in further detail below.
+
+
+## `build_java.pl`
+
+This is the primary script that drive the build system. It contains all logic
+for compiling the Java files and producing the output artifacts. It supports
+five primary operations:
+
+ - `clean` -- removes build artifacts from the directory, including `dist`
+   directory.
+ - `build` -- builds JSS by finding java files, compiling them, and creating
+    JNI headers
+ - `release` -- creates a release with all classes in one location.
+ - `javadoc` -- build documentation from the in-source javadocs
+   This creates an HTML version in `sandbox/dist/jssdoc`.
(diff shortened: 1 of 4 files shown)


docs/CI.md
```diff
@@ -0,0 +1,24 @@
+# Containers and Continuous Integration
+
+JSS uses [Travis CI](https://travis-ci.com/about) for PR gating. We use
```
SilleBille: How about adding the Travis link to repo instead of about? https://travis-ci.org/dogtagpki/jss

docs/CI.md
```diff
@@ -0,0 +1,24 @@
...
+buildah bud --tag jss_$BASE_IMAGE:latest -f tools/Dockerfiles/$BASE_IMAGE . 
+podman run jss_$BASE_IMAGE:latest
+```
+
+For more extensive documentation, please refer to the
+[Buildah](https://github.com/containers/buildah/blob/master/docs/) and
+[Podman docs](https://github.com/containers/libpod/tree/master/docs).
+
+To skip running CI for a given commit (e.g., for updating documentation),
+append `[skip ci]` to the commit summary.
```
SilleBille: How about adding a note something like "Even though test runs against `ubuntu`, the final build results are considered `Success` even if `ubuntu` build fails"? This will avoid the user from assuming that the CI has bugs when a user sees a X mark against ubuntu but still his/her PR passes. Your thoughts?

cipherboy:
This now includes documentation on the new `gh-pages` branch for https://pagure.io/jss/issue/24.

Merging as it is now complete and has been spellchecked. Further documentation can be added or revised later.

PR #41, opened 2018-10-11
PKCS11 Constants

cipherboy:
This PR introduces a new interface, `org.mozilla.jss.pkcs11.PKCS11Constants` to replace the old `sun.security.pkcs11.wrapper.PKCS11Constants` interface. 

#### Background

In general, the `sun.*` interface shouldn't have been used; this was documented in the [Java FAQ](https://www.oracle.com/technetwork/java/faq-sun-packages-142232.html) but never strictly enforced prior to JDK 9. With JDK 9 came a new organization system, ["modules"](https://www.oracle.com/corporate/features/understanding-java-9-modules.html) and as a result, any package not explicitly exported was considered fully private. Because PKCS11Constants resides within (the intended) private Sun API space, and this space was not explicitly exported, it is no longer accessible to classes outside the JDK for JDK 9+. Further, there appears to be no way to access the constants. 

However, NSS (and in turn JSS) depends on these constants for a few calls and hence it is important to expose them not only internally, but also externally to our users. Thus we faced two choices, discussed further in an RFC ([jss issue #26](https://pagure.io/jss/issue/26)):

1. Copy the source code of the Sun/Java implementation (`PKCS11Constants`).
2. Translate the source code of the NSS implementation (`pkcs11t.h`).

The former is of dubious copyright legality and the latter is more technically challenging. However, as most of our use cases for these constants involve calling into NSS, it makes sense to translate the NSS copy in a way that preserves the original Java calls.

#### Overview

This PR implements the second choice. This includes five commits:

1. A Python script to convert and test the NSS `pkcs11t.h` header file.
2. Small tests for style for Python code. Keeps us honest. :)
3. Check in the resulting Java file.
4. CI test to ensure we'll be notified if the NSS file ever changes and to make sure that Python 2 / Python 3 are handled well. 
5. A JSS test which validates the content of this file in JDK8 only. 

#### Expected Workflow

The expected workflow will be:

1. The resulting file is checked in; this way we don't introduce another build step and another hard dependency. Although admittedly, Python is already required on most systems we build on and Java and a C compiler are expected to exist for building this project anyways.
2. If/when NSS updates, we can regenerate the file and check it in again.
3. Consider adding a warning to the build if the contents of the generated file differ from the contents of the NSS header.

#### Outstanding Questions

The following questions have been addressed in later comments. 

~A few questions that I'd like to answer are:~

1. ~Should we add comments to the generated output saying which line we got a particular constant from?~ Yes, added.
2. ~Should we expand the header at the beginning of the generated file?~ Yes, copyright. Added.
3. ~In what release should we include this change? The next incremental release (4.5.x?), the next minor release (4.x?), or the next major release (5.0?).~ 4.5.1 according to discussion with @jmagne and @edewata.
4. ~Does the answer to 3 differ based on whether or not we include the transition in the test suite to the new functionality?~ Nope, everything is backwards compatible and we can include the test suite change in a later PR: #47. 

#### WIP

This PR is no longer WIP. 

It was WIP because:

1. ~I want to build a JDK 8 check that verifies the contents of the new and old file are roughly equivalent and report any differences.~ Done.
2. ~I need to validate that these changes work and preserve functionality. Currently that work is in a separate branch so this PR is technically backwards compatible: it exposes a new interface that we'll support in future versions, but doesn't update anything (i.e., the tests) to rely on it yet.~ Done, both in `cipherboy/jss:jdk11-full` branch (PR #47) and in a branch on PKI.
3. ~Documentation on the script, better comments, etc.~ Deferred for other doc updates in a later PR. There are plenty of comments in the script and in the output.


jmagne:
This looks like some really nice stuff. Going with nss seems reasonable.  I guess the biggest concern is if the rest of pki compiles using this thing in case there is any divergence between the original and these values. For instance I seem to recall in my deep recesses there may have been examples where a constant didn't appear in the sun class, and it had to come from somewhere else. As for the testing questions, I guess we just need to make sure all our software still works as expected using this class vs the old class.

cipherboy:
(I'll be updating the PR description with more information later to better document this).

I've created two branches, [one against JSS](https://github.com/cipherboy/jss/commits/pkcs11-full) and [one against PKI](https://github.com/cipherboy/pki/commits/pkcs11-jss-new) that use this new interface. I then used [`ansible-vms`](https://github.com/cipherboy/ansible-vms) to build JSS/TomcatJSS/PKI and [install a CA instance](https://github.com/cipherboy/ansible-vms/blob/master/pki-integration.yml). I'm guessing that the build system would've alerted me to missing symbols in PKI (as that's where the most symbols are used).

The other thing I want to do is create a test case which uses reflection to test the interfaces exposed by both and create a list of symbols that differ between the two and compare the value of all symbols. Since I created this from the NSS library, I know any calls into NSS will be fine since they use the same value, but it'd be nice to have assurances about anyone using this with Java, and compatibility with old code still using the Sun interface.

emaldona:
On the CK_ prefixed constants these are okay to add
Field: FALSE - only Sun
Field: NULL_PTR - only Sun
After all pkcs11t.h has
#define CK_TRUE 1
#define CK_FALSE 0
#define CK_NULL_PTR 0

On these:
Field: CK_EFFECTIVELY_INFINITE - only Sun
Field: CK_INVALID_HANDLE - only Sun             
Field: CK_UNAVAILABLE_INFORMATION - only Sun

I am not sure and will have to see it's usage

emaldona:
I like the TestPKSS11Contants.java. 
It uses reflceltion so it compiles wether the constants are visible or not. 
It won't run on JDK 11 but in all.pl I see
+if ($java =~ /1.8/i) {
+    # Only run the PKCS11Constants test on JDK 8. Newer versions do not
which takes care of that.
Based on this I created a patch to apply locally to my fork on the JDK_11_WORK_BRANCH  and things work fine.  I tested with 'make clean all', 'make testJava', and 'make javadoc' and it went. Covered JDK 8, 10, and 11 on fedora 28, and f29 beta. Also did debian 'buster', unstable branch for upcoming debian 10, and results as good as in fedora. There I could also switch among JDK versions.

emaldona:
Sorry, that bold and large test wasn't intended.

cipherboy:
@emaldona The issue with the `CK_` prefixed constants is that not all of them are `long`-like. 

---

- `CK_NULL_PTR` and `CK_PTR` are well, pointer types (`CK_NULL_PTR` appears to only used [with pointers](https://gist.github.com/cipherboy/48e55741d051d1218954096c49b44504) and is thus kinda useless in Java unless we create classes for everything and define `CK_NULL_PTR = null`, but I'm inclined to leave `null` as `null` and not a pretty name for `null`). So including those are not necessary I agree. There are a bunch more PTR types 

- Then there's the `CK_CALLBACK_FUNCTION`, `CK_DECLARE_FUNCTION`, and `CK_DECLARE_FUNCTION_POINTER` defines which seem interesting but are harder to replicate in Java. Java doesn't have functions as first-class citizens and lacks a macro system, so emulating C at this part would be a little much. Plus they're mostly constants used to define other things. Definitely not int types.

- `CK_INVALID_SESSION` appears to be an alias for a [null pointer](https://gist.github.com/cipherboy/00e670e7a1dca92378d6acd6fd312589) and again, we should just use `null` directly. Especially since I'd imagine JNI translates `(void *)(0)` to/from `null` in Java anyways. Looks to be a `null` / ptr type, not a `long` type. 

- `CK_INVALID_HANDLE` seems to have a similarly [limited utility](https://gist.github.com/cipherboy/9fba3e1292541b4b9d3b78f41e1983af).

- `CK_UNAVAILABLE_INFORMATION` is weird because it is the negation of 0 (i.e., `0xffffff...` to 64-bit size). It's used in a [few places](https://gist.github.com/cipherboy/dfda5e5afd3b8b017fabcd015a4fc737), but I'd have to dig deeper into that.

- `CK_EFFECTIVELY_INFINITE` is another one of those weird values since its value is 0. Which seems awfully finite for me. :-) But it is used in [very few places](https://gist.github.com/cipherboy/68ffdabf51e217a738f593ed7c7835ca). Go figure? I'll audit these more closely.

---

In short, I'd be more inclined to parse the `_NETSCAPE_` constants from `pkcs11n.h` rather than deal with the weird types on the `CK_` and add more fine-grained white listing.

cipherboy:
Okay, so after adding `pkcs11n.h` as a input header, I get the following results for `only Sun`:

```
contents.txt:51:Field: CKA_NETSCAPE_BASE - only Sun
contents.txt:52:Field: CKA_NETSCAPE_CERT_MD5_HASH - only Sun
contents.txt:53:Field: CKA_NETSCAPE_CERT_SHA1_HASH - only Sun
contents.txt:68:Field: CKA_NETSCAPE_TRUST_BASE - only Sun
contents.txt:69:Field: CKA_NETSCAPE_TRUST_CLIENT_AUTH - only Sun
contents.txt:70:Field: CKA_NETSCAPE_TRUST_CODE_SIGNING - only Sun
contents.txt:71:Field: CKA_NETSCAPE_TRUST_EMAIL_PROTECTION - only Sun
contents.txt:72:Field: CKA_NETSCAPE_TRUST_SERVER_AUTH - only Sun
contents.txt:436:Field: CKM_NSS_TLS_PRF_GENERAL - only Sun
contents.txt:720:Field: CK_EFFECTIVELY_INFINITE - only Sun
contents.txt:721:Field: CK_INVALID_HANDLE - only Sun
contents.txt:722:Field: CK_UNAVAILABLE_INFORMATION - only Sun
contents.txt:723:Field: FALSE - only Sun
contents.txt:725:Field: NULL_PTR - only Sun
contents.txt:726:Field: PCKK_ANY - only Sun
contents.txt:727:Field: PCKK_HMAC - only Sun
contents.txt:728:Field: PCKK_SSLMAC - only Sun
contents.txt:729:Field: PCKK_TLSMASTER - only Sun
contents.txt:730:Field: PCKK_TLSPREMASTER - only Sun
contents.txt:731:Field: PCKK_TLSRSAPREMASTER - only Sun
contents.txt:732:Field: PCKM_KEYSTORE - only Sun
contents.txt:733:Field: PCKM_SECURERANDOM - only Sun
contents.txt:734:Field: PCKO_ANY - only Sun
contents.txt:739:Field: TRUE - only Sun
```

(This output can be duplicated by running the test suite on a JDK 8 box).

Except for the `CK_` constants & friends, the other constants are missing from the NSS code, so I'm going to guess they are not needed. Pending CI pass (which does a system check to ensure the values of these constants are correct), this PR is good to go.

emaldona:
> On Oct 15, 2018, at 8:25 AM, Elio Maldonado <elio.maldonado.batiz@gmail.com> wrote:
> 
> Adding Bob in case he may wish to chime in
> 
Now I am doing it

emaldona:
Attempt to answer Outstanding Questions that @cipherboy posed.

> 1. Should we add comments to the generated output saying which line we got a particular constant from?

That might require manual labor or an expansion of the functionality of the Python script.
I would say only sparsely, if at all. Notice that the pkcs11t.h comments are by section and basically state in which PKCS #11 version a group of constants was introduced. Don't think is worth the effort going beyond that.

> 2. Should we expand the header at the beginning of the generated file?

The header at the beginning where it has the Copyright notice should be expanded
to include the Copright notice(s) of the file upon which this one is based orfrom which is generated.    NSS does that, See https://dxr.mozilla.org/mozilla-central/source/security/nss/lib/util/pkcs11t.h
where the Mozilla Copyright notice is followed by the one for RSA. I think  this is standard practice.

2. In what release should we include this change? The next incremental release (4.5.x?), 
the next minor release (4.x?), or the next major release (5.0?).

A 4.5.x certainly would be to soon, 4.6.0 (jdk9+ changes I think), maybe the
earliest and reduces the amount of work for 5.0 which, if I'm not mistake, will
be full jss migration to JDK 11. Others who know better than me the scope of
would be best qualified to answer.
    
4. Does the answer to 3 differ based on whether or not we include the transition
in the test suite to the new functionality?

If I'm not mistaken you have already wrap test 34 on a condition, this could be expanded
to additional tests you add. By the time JSS is getting tested on JDK 11 all tests would be activated.

cipherboy:
Regarding point 1 above... This has actually been added in the latest version of the script. The comments are a little vague, but can be made verbose with the `verbose` flag. In which case they show:
 - The complete derivation of the constant values which helps ensure correctness if you're ever reading it (How'd we get that magical value? _Oh_, it came from this file on this line, which eventually got expanded to that.. Okay)
 - The output of running the `c <-> parsed value` interoperability checks). 

--- 

Regarding the copyright notice, this has been addressed in the latest pushed changes. In particular I now include the sourced headers initial comments which are the copyright notices. The complete header on our file now looks like:

```java
/**
 * PKCS11Constants.java - part of the JSS project
 * Copyright(C) 2018 Red Hat, Inc.
 *
 * This file is automatically generated (tools/build_pkcs11_constants.py)
 * from the contents of NSS's pkcs11t.h and pkcs11n.h headers. This
 * replaces the sun.security.pkcs11.wrapper.PKCS11Constants interface
 * which was removed from public visibility in JDK 9 due to the
 * introduction of modules.
 *
 * For more information, please see the documentation under
 * jss/docs/pkcs11_constants.md.
 *
 * Below are the copyright statements for the sourced files:
**/
/* Copyright statements from /usr/include/nss3/pkcs11t.h */
/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
/* License to copy and use this software is granted provided that it is
 * identified as "RSA Security Inc. PKCS #11 Cryptographic Token Interface
 * (Cryptoki)" in all material mentioning or referencing this software.
 * License is also granted to make and use derivative works provided that
 * such works are identified as "derived from the RSA Security Inc. PKCS #11
 * Cryptographic Token Interface (Cryptoki)" in all material mentioning or
 * referencing the derived work.
 * RSA Security Inc. makes no representations concerning either the
 * merchantability of this software or the suitability of this software for
 * any particular purpose. It is provided "as is" without express or implied
 * warranty of any kind.
 */

/* Copyright statements from /usr/include/nss3/pkcs11n.h */
/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```

So I think that should suffice.

--- 

Regarding versions &c... I think migrating JSS test suite to the new functionality is good to go but I'll include that commit in a separate PR. That'll work since JSS is tightly coupled to its test suite, and was made possible by testing the values of the constants across C and Java. These changes are backwards-compatible in that they introduce a new API and don't remove any existing APIs. So I don't think there's anything breaking that prevents us from putting it in a minor or point release, but I'll leave this tagged for 4.6.x for the time being.

cipherboy:
Per discussion with @edewata and @jmagne, we'll be targeting 4.5.1 with these changes since we can support JDK8+ with the same code base and without breaking backwards compatibility.

emaldona:
No hacks needed, it works fine.

tools/build_pkcs11_constants.py
```diff
@@ -0,0 +1,567 @@
...
+        self.stdout = proc.stdout.read()
+        if not isinstance(self.stdout, str):
+            self.stdout = self.stdout.decode('utf8')
+
+        self.stderr = proc.stderr.read()
+        if not isinstance(self.stderr, str):
+            self.stderr = self.stderr.decode('utf8')
+
+        if self.stdout:
+            print("When checking symbol '%s': " + self.stdout)
```
cipherboy: Need to fill in the format string here.

cipherboy:
Okay, I added extended commentary in the utility which generates this based on discussion with @jmagne. Also as part of that, we decided to order `PKCS11Constants.java` by the order of the header files it is generated from (`pkcs11t.h` and then `pkcs11n.h`).

cipherboy:
And pushed an update... Looks like `.copy()` on a `list` was added in Python 3, so it doesn't work on Python 2. `list[:]` works on both though, so updated.

org/mozilla/jss/pkcs11/PKCS11Constants.java
```diff
@@ -0,0 +1,8623 @@
...
+     *
+     * Content automatically generated; see NSS documentation for more information.
+     *
+     * Source file: /usr/include/nss3/pkcs11t.h
+     * Line number: 103
+     * Line: #define CKN_SURRENDER 0
+     * Parsed name: CKN_SURRENDER
+     * Parsed value: 0
+     *
+    **/
```
edewata: Generally the comment block should end with just */ instead of **/ and the * should align with the previous line. See also: https://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#examples

This type of comment block will be used by Javadoc tool to generate documentation content.

org/mozilla/jss/pkcs11/PKCS11Constants.java
```diff
@@ -0,0 +1,8623 @@
...
+
+/* Copyright statements from /usr/include/nss3/pkcs11n.h */
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.jss.pkcs11;
+
+public interface PKCS11Constants {
+    /**
```
edewata: Probably better to add a space before /**.

org/mozilla/jss/pkcs11/PKCS11Constants.java
```diff
@@ -0,0 +1,8623 @@
...
+     *
```
edewata: The initial blank line is unnecessary.

org/mozilla/jss/pkcs11/PKCS11Constants.java
```diff
@@ -0,0 +1,8623 @@
...
+     * Content automatically generated; see NSS documentation for more information.
+     *
+     * Source file: /usr/include/nss3/pkcs11t.h
+     * Line number: 103
```
edewata: Not sure if line number is necessary to include here. The constant identifier should be sufficient.

org/mozilla/jss/pkcs11/PKCS11Constants.java
```diff
@@ -0,0 +1,8623 @@
...
+     * Line: #define CKN_SURRENDER 0
+     * Parsed name: CKN_SURRENDER
+     * Parsed value: 0
+     *
```
edewata: The final blank line is unnecessary.

org/mozilla/jss/tests/TestPKCS11Constants.java
```diff
@@ -0,0 +1,71 @@
...
+                    assert(sun_field.getType() == long.class);
+                    assert(jss_field.getLong(null) == sun_field.getLong(null));
+                    System.out.println("Field: " + key + " - OK");
+                } else if (jss_map.containsKey(key)) {
+                    System.err.println("Field: " + key + " - only JSS");
+                } else if (sun_map.containsKey(key)) {
+                    System.err.println("Field: " + key + " - only Sun");
+                }
+            }
+        } catch (Exception e) {
```
edewata: We should avoid this type of catch block since it doesn't handle anything and it hides the actual stack trace. I'd suggest adding "throws Exception" in the main() method definition, then remove the try-catch block completely. If an error occurs, java CLI will display the stack trace so we can troubleshoot the problem.

org/mozilla/jss/pkcs11/PKCS11Constants.java
cipherboy: Per discussion with @edewata, this was about a newline before this line, which has been fixed.

org/mozilla/jss/tests/TestPKCS11Constants.java
cipherboy: Fair point. I had originally done that as without the `throws Exception` or `try`+`catch`, Java refuses to compile the test case:

```java
./org/mozilla/jss/tests/TestPKCS11Constants.java:16: error: unreported exception ClassNotFoundException; must be caught or declared to be thrown
        Class jss = Class.forName("org.mozilla.jss.pkcs11.PKCS11Constants");
                                 ^
./org/mozilla/jss/tests/TestPKCS11Constants.java:17: error: unreported exception ClassNotFoundException; must be caught or declared to be thrown
        Class sun = Class.forName("sun.security.pkcs11.wrapper.PKCS11Constants");
                                 ^
./org/mozilla/jss/tests/TestPKCS11Constants.java:57: error: unreported exception IllegalAccessException; must be caught or declared to be thrown
                assert(jss_field.getLong(null) == sun_field.getLong(null));
                                        ^
./org/mozilla/jss/tests/TestPKCS11Constants.java:57: error: unreported exception IllegalAccessException; must be caught or declared to be thrown
                assert(jss_field.getLong(null) == sun_field.getLong(null));
                                                                   ^
```

cipherboy:
We got green on the CI and two ACKs so I'm merging this. Thanks @edewata, @jmagne, and @emaldona for the reviews!

PR #42, opened 2018-10-15
Add -notimestamp option to javadoc creation

cipherboy:
This prevents reproducible builds for the `jss-javadoc` sub-package; in particular, any compilation at a slightly different time will produce javadocs which differ in contents by a hidden comment. This is undesirable and thus should be prevented with the `-notimestamp` feature. 


cipherboy:
Merging. I verified that this build the javadoc with the timestamp comment removed.

PR #43, opened 2018-10-15
lib/Common.pm - Common library for build_java.pl and tests/all.pl

cipherboy:
This PR contains work by @emaldona that he has agreed to let me separate out from his JDK11-related work.

This introduces a `lib/Common.pm` Perl module for use by both `build_java.pl` and `tests/all.pl`, along with various small refactoring. This removes the call to `javah`, which has been removed in JDK9+, and replaces it to a single call to `javac` with the `-h` flag.

I'll be reviewing and cleaning up these two commits in this PR.

### DIFF
--- a/build_java.pl
+++ b/build_java.pl
@@ -1,4 +1,6 @@
-#use strict;
+use strict;
+use warnings;
+
 use Cwd qw(abs_path cwd);
 use File::Find;
 use File::Compare;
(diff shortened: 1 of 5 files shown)


build_java.pl
```diff
@@ -1,63 +1,75 @@
...
+use warnings;
 use Cwd qw(abs_path cwd);
 use File::Find;
 use File::Compare;
 use File::Basename;
 use File::stat;
 use File::Copy;
 
+use File::Basename qw(dirname);
+use Cwd  qw(abs_path);
```
cipherboy: Duplicate include.

build_java.pl
```diff
@@ -1,63 +1,75 @@
...
-org.mozilla.jss.pkcs11.PK11ECPublicKey
-org.mozilla.jss.pkcs11.PK11SecureRandom 
-org.mozilla.jss.provider.java.security.JSSKeyStoreSpi
-org.mozilla.jss.SecretDecoderRing.KeyManager
-org.mozilla.jss.ssl.SSLSocket 
-org.mozilla.jss.ssl.SSLServerSocket 
-org.mozilla.jss.ssl.SocketBase 
-org.mozilla.jss.util.Password       
+# source list equivalent to the old @javah_classes
+my @wanted_sources = qw(
```
cipherboy: Need to see if these need to be extended.

build_java.pl
```diff
@@ -169,9 +180,8 @@ sub setup_vars {
     our $jni_header_dir = "$dist_dir/private/jss/_jni";
 
     our $jarFiles = "/usr/share/java/slf4j/slf4j-api.jar:/usr/share/java/commons-codec.jar";
-    if( $ENV{DEBIAN_BUILD} ) {
-        $jarFiles = "/usr/share/java/slf4j-api.jar:/usr/share/java/commons-codec.jar";
-    }
+
+    $jarFiles = Common::get_jar_files;
```
cipherboy: Make this `our $jarFiles` and remove line 182.

lib/Common.pm
```diff
@@ -0,0 +1,30 @@
...
+        # then add another elfsif section to handle it.
+    } else {
+        $jarFiles = "/usr/share/java/slf4j/api.jar:/usr/share/java/commons-codec.jar:/usr/share/java/jaxb-api.jar";
+        $jarFiles = "$jarFiles:/usr/share/java/commons-lang.jar";
+    }
+
+    return $jarFiles;
+}
+
+1;
```
cipherboy: Why the `1;`? Investigate.

org/mozilla/jss/crypto/Algorithm.c
```diff
@@ -13,7 +13,6 @@
 
 #include <jssutil.h>
 
-#include "_jni/org_mozilla_jss_crypto_Algorithm.h"
```
cipherboy: Did the location perhaps change?

org/mozilla/jss/tests/all.pl
```diff
@@ -9,6 +9,16 @@
 use Cwd 'abs_path';
 use POSIX 'uname';
 
+use strict;
```
cipherboy: Move to top of file?

org/mozilla/jss/tests/all.pl
```diff
@@ -9,6 +9,16 @@
...
+use warnings;
+ 
+use File::Basename qw(dirname);
+use Cwd  qw(abs_path);
```
cipherboy: Already included.

org/mozilla/jss/tests/all.pl
```diff
@@ -207,7 +217,7 @@ sub setup_vars {
     if ($osname =~ /Darwin/) {
         $java = "$ENV{JAVA_HOME}/bin/java";
```
cipherboy: We could perhaps refactor some of this logic as well.

org/mozilla/jss/tests/all.pl
```diff
@@ -455,14 +461,9 @@ sub print_case_result {
...
-$jarFiles = "";
-
-if( $ENV{DEBIAN_BUILD} ) {
-    $jarFiles = "/usr/share/java/slf4j-api.jar:/usr/share/java/slf4j-jdk14.jar:/usr/share/java/commons-lang.jar";
-} else {
-    $jarFiles = "/usr/share/java/slf4j/slf4j-api.jar:/usr/share/java/slf4j/slf4j-jdk14.jar:/usr/share/java/commons-lang.jar";
-}
-
+$jarFiles = Common::get_jar_files;
+$jarFiles = "$jarFiles:/usr/share/java/slf4j/jdk14.jar";
```
cipherboy: Perhaps verify this location across Debian and Fedora. I'd like a `get_jar_file` which takes an argument and searches a few locations for the jar in question; should be better for builds.

org/mozilla/jss/util/jssutil.c
```diff
@@ -12,7 +12,6 @@
 #include "jss_exceptions.h"
 #include "java_ids.h"
 
-#include "_jni/org_mozilla_jss_util_Password.h"
```
cipherboy: Verify this include as well.

build_java.pl
emaldona: Yes,  keep the one at the top.

lib/Common.pm
emaldona: Likely a copy and paste typo.

org/mozilla/jss/tests/all.pl
emaldona: Yes

build_java.pl
emaldona: Yes, I tried it and it works.

org/mozilla/jss/tests/all.pl
emaldona: I see ensure_dir_exists so perhaps also ensure_file_exists to check the jar file exists. You'll know how best to do do it.

org/mozilla/jss/tests/all.pl
emaldona: Perhaps but notice that setup_vars on all.pl is quite different from the one in build_java.pl

org/mozilla/jss/util/jssutil.c
cipherboy: Yep, I can confirm no native functions.

org/mozilla/jss/crypto/Algorithm.c
cipherboy: Yep, I can confirm no native functions.

emaldona:
Two of the three patches would not apply cleanly. Perhaps there was another commit after they were posted. The last commit I had when I tried applying them is:
commit ff28a6cf6c403d68f89c3fea850eb8d88564dab2 (HEAD -> master, origin/master, origin/HEAD)
Author: Alexander Scheel <ascheel@redhat.com>
Date:   Thu Oct 11 12:34:04 2018 -0400

I fixed the problems to make it work,  make clean all javadoc and testJava were fine.
Tested on java-1.8.0-openjdk, attached is the revised combined patch I used.
[pr43revised.patch.txt](https://githubtestJava'.com/dogtagpki/jss/files/2481004/pr43revised.patch.txt)
I had to add the .txt suffix otherwise it woudn't let me attach.

emaldona:
Won't let me do it, I'll email @cipherboy the patch then.

cipherboy:
So the order I see is:

5037ecfd075a9b4015d7e9ef80d8332e4df95272
db574ea935ac246537304f128e04eab64957c0e4
5bf20fc8c5295fa6dbb46720658601b6da4963b9

emaldona:
Aha, order  whichI tried them was
5037ecf
db574e
5bf20fc
backwards, maybe that's the reason for the problems I had.

emaldona:
Yes, the order you gave works fine.

emaldona:
This is good to go, approved.

emaldona:
A clarification is in order. As it is now ,this works on jdk 8 only.
when I tried jdk 9 I got  /org/mozilla/jss/tests/SymKeyDeriving.java:43: error: package sun.security.pkcs11.wrapper is not visible
1 error
The patch does not include the PKCS11Constants work currently under discussion at https://github.com/dogtagpki/jss/pull/41

cipherboy:
Correct; I'd be pushing this work independently of the work on PKCS11 Constants. Since these build changes work on JDK8+, I'm fine including them and testing them while we're still working on finalizing JDK11 support through the other work. Also, since build changes aren't really `jss4.jar`-consumer facing, I think collecting as many of them together in one small point release will be good. :)

cipherboy:
Okay, I tested this again on Ubuntu and pulled this series of commits, the series of commits from PR #41, and the two commits necessary to create a working JDK11 build into my branch [`cipherboy:lcp-pc`](https://github.com/cipherboy/jss/commits/lcp-pc) and tested this on a Ubuntu 18.04 machine. I'll test on a Debian Sid machine and then merge.

Thanks @emaldona for, code, help, and suggestions!

PR #44, opened 2018-10-15
Create script to automatically detect exports for build

cipherboy:
This introduces a new script, `tools/autoenv.sh` that should hope to plaster over most of the differences across Debian/SUSE/Fedora and allow the user to `source tools/autoenv.sh && make clean all check` to build the project easier.

In particular, we:
 - Make an attempt to detect a JDK 8 version.
 - Make an attempt to detect 64-bit mode.
 - Check if we need to export a build-env specific flag
 - Check if we need to use the system NSS/NSPR or if we're building in a sandbox with NSS/NSPR.

### DIFF
--- a/tools/autoenv.sh
+++ b/tools/autoenv.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+# This script attempts to detect the correct environmental variables required
+# to build JSS from source.
+
+# Load OS information.
+if [ -f /etc/os-release ]; then
+    . /etc/os-release
+elif [ -f /usr/lib/os-release ];then
+    . /usr/lib/os-release
+else
+    echo "Cannot determine OS information. Exiting..." 1>&2
+    exit 1
+fi
+
+# Try and detect a JDK installation.
+if [ -d "/etc/alternatives/java_sdk" ]; then
+    export JAVA_HOME="/etc/alternatives/java_sdk"
+else
+    java8_jdk="$(find /usr/lib/jvm -maxdepth 1 -mindepth 1 -type d | grep '\(\-8\-\|-1\.8\.\)' | sort | tail -n 1)"
+    if [ -d "$java8_jdk" ]; then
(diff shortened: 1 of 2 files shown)


tools/autoenv.sh
```diff
@@ -0,0 +1,58 @@
...
+else
+    echo "Cannot determine OS information. Exiting..." 1>&2
+    exit 1
+fi
+
+# Try and detect a JDK installation.
+if [ -d "/etc/alternatives/java_sdk" ]; then
+    export JAVA_HOME="/etc/alternatives/java_sdk"
+else
+    java8_jdk="$(find /usr/lib/jvm -maxdepth 1 -mindepth 1 -type d | grep '\(\-8\-\|-1\.8\.\)' | sort | tail -n 1)"
```
cipherboy: Okay, so I changed this logic to try and detect `jdk8` automatically. This will probably be updated eventually with JDK-11 related stuff at which point we can do a more complete feature detection (`javac`, `java`, `jar`, etc. in the same bin directory), but it suffices for now on Fedora and Ubuntu.

tools/autoenv.sh
```diff
@@ -0,0 +1,58 @@
...
+    NSPR_LIB_DIR="$(pkg-config --libs-only-L nspr | sed 's/-L//')"
+    NSS_INCLUDE_DIR="$(pkg-config --cflags-only-I nss | sed 's/-I//')"
+    NSS_LIB_DIR="$(pkg-config --libs-only-L nss | sed 's/-L//')"
+    export NSPR_INCLUDE_DIR
+    export NSPR_LIB_DIR
+    export NSS_INCLUDE_DIR
+    export NSS_LIB_DIR
+    export XCFLAGS="-g"
+else
+    export USE_INSTALLED_NSPR=""
```
cipherboy: This is important in case you switch from `system -> sandbox` builds of NSS/NSPR.

PR #47, opened 2018-10-16
JDK9+ support

cipherboy:
#### Dependencies

- #41: PKCS11Constants -- Those commits are duplicated here for CI / testing efforts. When that is merged, this PR can be rebased.
- #48: Perl improvements -- Those commits are duplicated here for CI / testing efforts. When that is merged, this PR can be rebased. 

#### Description

This PR introduces two new commits:

1. Switching to the new PKCS11 Constants interface exposed in #41.
2. Add a Debian Testing / JDK11 test to the Travis matrix + rework the Travis matrix.

The second commit adds a Dockerfile for `debian_jdk11` which installs the necessary dependencies and runs the test against JDK11. Reworking the Travis matrix at the same time to allow easy view of what's required and what isn't:

screenshot from 2018-10-17 13-27-55

I've opted to make `debian_jdk11` optional as it _is_ Debian Testing and might be unstable. Also, Debian is the only distribution thus far to ship a real JDK11; `rawhide` only supports JDK 10, but ships it as 11.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,11 @@ services:
 stages:
   - validate
   - test
+  - extra
 
 env:
   - BASE_IMAGE="fedora_27"
   - BASE_IMAGE="fedora_28"
-  - BASE_IMAGE="ubuntu_jdk8"
-  - BASE_IMAGE="pkcs11check"
 
 script:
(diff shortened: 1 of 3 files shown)


cipherboy:
As this is a minor change + CI, I've decided to go ahead and push it as CI is green :)

PR #48, opened 2018-10-17
Various Perl improvements

cipherboy:
This PR includes several Perl improvements split off from #47. In particular:

 - Do jar detection versus using ENV vars to set the location of jars in `lib/Common.pm`
 - Use fatal warnings in `build_java.pl` and `tests/all.pl` -- this found a number of bugs and more importantly, stops the script in the event of failure
 - Uses the revised autoenv.sh in the Ubuntu build; this should allow us to test it more frequently.

Of particular note in the `tests/all.pl` revisions are the following:

 - An error in the tests now results in an error in the build. Previously the exit status was still zero in the event of one or more test failures.
 - `NATIVE_FLAG` has been removed. There appears to have been a [discussion with wtc](https://bugzilla.mozilla.org/show_bug.cgi?id=302550) in which nobody stated its purpose.
 - Various cleanup in the handling of the classpath and detecting libraries.

### DIFF
--- a/lib/Common.pm
+++ b/lib/Common.pm
@@ -2,32 +2,50 @@ package Common;
 
 use strict;
 use warnings;
+use warnings FATAL => 'all';
 
 use Exporter qw(import);
 
 # The below methods are used by build_java.pl and tests/all.pl; if the
 # location of Common.pm ever changes, update the includes in those files
 # to point to the new location.
-our @EXPORT_OK = qw(get_jar_files);
+our @EXPORT_OK = qw(get_jar_files detect_jar_file);
+
+sub detect_jar_file {
+    # Detect the correct path to a jar file.
+    # Die if the jar is missing.
+
+    my @locations = qw(/usr/share/java /usr/lib/java);
+    for my $location (@locations) {
+        for my $candidate (@_) {
+            my $path="$location/$candidate";
+            if (-e $path) {
+                return $path;
+            }
+        }
+    }
+
+    die "Can't find any jars for $_[1]";
+}
 
 sub get_jar_files {
     # Return a list of JAR files of the dependencies of JSS; in particular,
     # handle Debian vs OpenSUSE vs Fedora builds and the locations of relevant
     # jars on those platforms.
-    our $jarFiles = "";
 
-    if( $ENV{DEBIAN_BUILD} ) {
-        $jarFiles = "/usr/share/java/slf4j-api.jar:/usr/share/java/commons-codec.jar:/usr/share/java/jaxb-api.jar";
-        $jarFiles = "$jarFiles:/usr/share/java/commons-lang.jar"
-    } elsif( $ENV{OPENSUSE_BUILD} ) {
-        $jarFiles = "/usr/share/java/slf4j/api.jar:/usr/share/java/apache-commons-codec.jar:/usr/share/java/apache-commons-lang.jar";
-        $jarFiles = "$jarFiles:/usr/share/java/jaxb-api.jar:/usr/share/java/apache-commons-lang.jar"
-        # If some distro doesn't match the cases that we have identified
-        # then add another elfsif section to handle it.
-    } else {
-        $jarFiles = "/usr/share/java/slf4j/api.jar:/usr/share/java/commons-codec.jar:/usr/share/java/jaxb-api.jar";
-        $jarFiles = "$jarFiles:/usr/share/java/commons-lang.jar";
+    our @jarFiles = ();
+    my $slf4j = detect_jar_file("slf4j-api.jar", "slf4j/api.jar");
+    my $codec = detect_jar_file("apache-commons-codec.jar", "commons-codec.jar");
+    my $lang = detect_jar_file("apache-commons-lang.jar", "commons-lang.jar");
+    push(@jarFiles, $slf4j);
+    push(@jarFiles, $codec);
+    push(@jarFiles, $lang);
+
+    if (defined $ENV{JDK9_BUILD} and $ENV{JDK9_BUILD}) {
+        my $jaxb = detect_jar_file("jaxb-api.jar", "jboss-jaxb-2.2-api.jar");
+        push(@jarFiles, $jaxb);
     }
 
-    return $jarFiles;
+
+    return join(':', @jarFiles);
 }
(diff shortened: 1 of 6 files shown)


lib/Common.pm
```diff
@@ -2,32 +2,50 @@ package Common;
...
+            if (-e $path) {
+                return $path;
+            }
+        }
+    }
+
+    die "Can't find any jars for $_[1]";
+}
 
 sub get_jar_files {
```
cipherboy: In the future it would be nice to have this accept an array of additional jars to find / search for. But it'd likely have to be an array of arrays which I'm not wanting to introduce that complexity right now. Instead I exported `detect_jar_file` so others callers (`tests/all.pl`) can use that.

org/mozilla/jss/tests/all.pl
```diff
@@ -92,7 +93,7 @@ sub checkPort {
 my $run_shell    = "";
 my $pwfile       = "";
 my $nss_bin_dir  = "";
-my $jarFiles     = "";
+my $classpath    = "";
```
cipherboy: No need to expose jarFiles; handle this in `setup_vars` and let everything use `classpath`.

org/mozilla/jss/tests/all.pl
```diff
@@ -131,7 +132,6 @@ sub setup_vars {
         $scriptext = "sh";
     }
 
-    $ENV{CLASSPATH}  = "";
```
cipherboy: Since we have `$classpath` and explicitly pass it on all `java` calls, this variable can disappear.

org/mozilla/jss/tests/all.pl
```diff
@@ -317,11 +317,14 @@ sub outputEnv {
 
    print "*****ENVIRONMENT*****\n";
    print "java=$java\n";
-   print "NATIVE_FLAG=$ENV{NATIVE_FLAG}\n";
    print "$ld_lib_path=$ENV{$ld_lib_path}\n";
-   print "CLASSPATH=$ENV{CLASSPATH}\n";
-   print "BUILD_OPT=$ENV{BUILD_OPT}\n";
-   print "USE_64=$ENV{USE_64}\n";
+   print "CLASSPATH=$classpath\n";
+   if (defined $ENV{BUILD_OPT}) {
```
cipherboy: I wasn't sure the purpose of printing this at the test stage, but figured it'd be OK to leave it in.

org/mozilla/jss/tests/all.pl
```diff
@@ -711,14 +709,14 @@ sub print_case_result {
 print "\n================= Test Results\n";
 print "JSSTEST_SUITE: $testpass / $testrun\n";
 my $rate = $testpass / $testrun * 100;
-printf "JSSTEST_RATE: %.0f %\n",$rate;
+printf "JSSTEST_RATE: %.0f %%\n",$rate;
```
cipherboy: Double percent required to print a percent... `man printf` :)

org/mozilla/jss/tests/all.pl
```diff
@@ -711,14 +709,14 @@ sub print_case_result {
...
 
 if ($testpass ne $testrun) {
     printf "Test Status: FAILURE\n";
-    system("false");
     printf "to test failed tests set the classpath and run the command(s)\n";
     outputEnv();
+    exit 1;
```
cipherboy: I wasn't sure the purpose of using `system("false")` so I changed it to `exit 1`... Perhaps it wasn't used due to [portability reasons](https://perldoc.perl.org/perlport.html#exit) but those platforms aren't our main candidates.

tools/autoenv.sh
```diff
@@ -1,38 +1,46 @@
-#!/bin/bash
+#!/bin/sh
```
cipherboy: Change required for use from Dockerfiles... Can't use non-standard `==` either.

tools/autoenv.sh
```diff
@@ -1,38 +1,46 @@
...
-    exit 1
-fi
-
 # Try and detect a JDK installation.
-if [ -d "/etc/alternatives/java_sdk" ]; then
+if [ "x$JAVA_HOME" = "x" ] && [ -e "/etc/alternatives/java_sdk" ]; then
     export JAVA_HOME="/etc/alternatives/java_sdk"
-else
+fi
+if [ "x$JAVA_HOME" = "x" ]; then
```
cipherboy: I decided to split these into separate `if ... ; then ... else` blocks versus chaining them.

tools/autoenv.sh
```diff
@@ -1,38 +1,46 @@
...
+        export JAVA_HOME="$system_java"
+    elif [ -e "$dirname_system_java/bin/javac" ]; then
+	export JAVA_HOME="$dirname_system_java"
+    fi
+fi
+
+# Validate that JAVA_HOME was set; otherwise, exit.
+if [ "x$JAVA_HOME" = "x" ]; then
+    echo "Unable to detect JAVA_HOME installation automatically!" 1>&2
+    exit 1
```
cipherboy: This should be a fatal error such that the user is notified sooner rather than when they go to run `make` or `build_java.pl`.

lib/Common.pm
```diff
@@ -2,32 +2,50 @@ package Common;
...
 # location of Common.pm ever changes, update the includes in those files
 # to point to the new location.
-our @EXPORT_OK = qw(get_jar_files);
+our @EXPORT_OK = qw(get_jar_files detect_jar_file);
+
+sub detect_jar_file {
+    # Detect the correct path to a jar file.
+    # Die if the jar is missing.
+
+    my @locations = qw(/usr/share/java /usr/lib/java);
```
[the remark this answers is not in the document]
emaldona: In case in the future you main need it, it got this:
$ ls -l /usr/bin/java
/usr/bin/java -> /System/Library/Frameworks/JavaVM.framework/Versions/Current/Commands/java
/usr/bin/javac -> /System/Library/Frameworks/JavaVM.framework/Versions/Current/Commands/javac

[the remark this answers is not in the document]
cipherboy: Hmmm and the locations of the required Jars? I'm guessing they're not shipped with Mac and we'd have to include Brew/MacPorts installation paths... Hmmmm

PR #49, opened 2018-10-17
Fix JDK9+ check

cipherboy:
As reported by @emaldona...

This check fails to correctly detect Java 11 because `1.11` < `1.9`... Let's do feature detection instead. This change was introduced in JDK9 and is what we're really looking for here. 

### DIFF
--- a/tools/autoenv.sh
+++ b/tools/autoenv.sh
@@ -32,9 +32,31 @@ if [ "x$JAVA_HOME" = "x" ]; then
     exit 1
 fi
 
-# Check if the JDK version is 9 or greater...
-java9_check="$(jrunscript -e "print(java.lang.Double.parseDouble(java.lang.System.getProperty('java.specification.version')) >= java.lang.Double.parseDouble('1.9'))" 2>/dev/null)"
-if [ "x$java9_check" = "xtrue" ]; then
+# Check if the JDK version is 9 or greater via feature detection on
+# sun.security.pkcs11.wrapper...
+java9_check() {
+    tmp_dir="$(mktemp -d)"
+    if [ "x$tmp_dir" = "x" ]; then
+        echo "Cannot create temp directory!" 1>&2
+        exit 1
+    fi
+
+    tmp_file="$tmp_dir/Main.java"
+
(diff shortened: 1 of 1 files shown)


cipherboy:
I think this is part of the reason I'd like to rely on a better build system here. Something like CMake that is easier to use and is a little better tested and doesn't require us to rely on a collection of shell and Perl scripts. 

I think the first steps towards replacing the existing build and test system is:

1. Understanding how it is built.
2. Understanding why it was built.
3. Understanding what unique functionality it provides.
4. Understanding what other build systems (ant, gradle, cmake, make, etc.) provide similar functionality.

And then we can write a replacement.

emaldona:
Yes, see https://github.com/emaldona/jss/commit/f2345a0ca031b850e78c355152edc5ae86abb000
and that worked for me on JDK 8 as well.

cipherboy:
@emaldona Yeah, so that's the extra dependency I was talking about: `glassfish-jaxb-api` (on Fedora) provides the `/usr/share/java/jaxb-api.jar` jar file. If you install that package (+JDK 11), it should build fine on F29 with the above patch. On Debian, `libjaxb-api-java` provides that jar file. This is how the Debian 11 build is succeeding in #47: installing that package requires no code changes on our part.

If we introduce `glassfish-jaxb-api` as a dependency in 4.5.1, this will give us JDK8+ support at the cost of a useless under Jar in JDK8. AFAICT, including this jar in the classpath doesn't break anything on JDK8.

emaldona:
I found out why I had problems. My local builds are manual ones
I had to manually do 'export JDK9_BUILD=1' before building just like I do with JAVA_HOME and USE_64, as I am not using Docker stuff. Just got this
<cipherboy> emaldonado: You can use the tools/autoenv.sh without using the Docker stuff :)
<cipherboy> emaldonado: We just need to merge the fix in the PR though for JDK9+ :/

cipherboy:
Per initial discussion on [jss issue#27 -- RFC](https://pagure.io/jss/issue/27), I'll be closing this PR and opening a new one which adds Glassfish's JAXB as a dependency.

PR #54, opened 2018-10-19
Add dependency on JAXB for JDK9+

cipherboy:
Per discussion in #49 and [jss issue#27 / RFC](https://pagure.io/jss/issue/27), add a dependency on JAXB for JDK9+ builds. This is the last major issue preventing a build for JDK8 - JDK11 besides #41.


cipherboy:
@edewata Is there anything else I need to do to add this dependency upstream?

PR #59, opened 2018-10-22
Fix problem where firefox could not display the INDEX tab as of jdk 11

emaldona:
- address: https://pagure.io/jss/issue/28
- add "-overview ./overview.html" option to the javadoc command invocation
- add overview.html file at the top of the source tree
- overview.html provides a title and a brief description of jss
- no longer need to replace index.html with overview-summary.html

### DIFF
--- a/build_java.pl
+++ b/build_java.pl
@@ -455,9 +455,7 @@ sub javadoc {
     ensure_dir_exists("$dist_dir/jssdoc");
     my $targets = join(" ", @packages);
     print "$targets\n";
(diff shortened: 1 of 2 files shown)


cipherboy:
(I rebased this on top of the latest master).

cipherboy:
Okay, this looks good to me. I can confirm that the original problem does exist, that these changes build the javadoc correctly, and that after this, the issue no longer occurs. Thanks @emaldona! :)

PR #61, opened 2018-10-23
Update javadocs from master at 2018-10-23 16:18

cipherboy:
This is generated because #59 was merged.

@emaldona I notice the title has changed on the `index.html` file... Do you know if we can get the old title back or otherwise change the title?


emaldona:
@cipherboy which previous title do you prefer?

cipherboy:
@emaldona Well, "Overview" is a little better than "Generated Documentation (Untitled)"... But I'd prefer to customize it. Looking at https://docs.oracle.com/javase/6/docs/technotes/tools/windows/javadoc.html#overviewcomment it looks like I just need to add a title to the overview.html file. Let me try that out. :)

emaldona:
Yes, once you try it out you could post your suggested changes, perhaps in patch form,  and I, along with others, can give you feedback.

cipherboy:
I'm not the most fond of putting the window title in parenthesis, but that's what `javadoc` does with the `-windowtitle` flag, so I guess I'll have to live with it.

What do you think @emaldona? This'll show up as the title on the hosted version when merged: https://dogtagpki.github.io/jss/javadoc/

emaldona:
The boldface on title is gone when looking at it in the site. The Oracle one at
https://docs.oracle.com/en/java/javase/11/docs/api/index.html looks the way we want.
From their example
C:> javadoc -sourcepath \java\jdk\src\share\classes            ^
    -overview \java\jdk\src\share\classes\overview.html      ^
    -d \java\jdk\build\api                                   ^
    -use                                                     ^
    -splitIndex                                              ^
    -windowtitle 'Java 2 Platform 5.0 API Specification'    ^
    -doctitle 'Java<sup><font size="-2">TM</font></sup> 2 Platform 5.0 API Specification' ^
    -header '<b>Java 2 Platform </b><br><font size="-1">5.0</font>' ^
    -bottom '<font size="-1"><a href="http://java.sun.com/cgi-bin/bugreport.cgi">Submit
I don't see the parentheses you mention. They may have something we lack. 
On build_java.pl 
 print_do("$javadoc -overview ./tools/javadoc/overview.html -windowtitle 'JSS: Java Security Services' -notimestamp -breakiterator $classpath -sourcepath . -d $dist_dir/jssdoc $html_header_opt $targets");
where $html_header_opt is is set a few lines eralier
    my $html_header_opt = "";
    if( defined $ENV{HTML_HEADER} && $ENV{HTML_HEADER} ) {
        $html_header_opt = "-header '$ENV{HTML_HEADER}'";
    }
A command line option that I don't think we are not taking advantage off.  I'll look into this along the way.

emaldona:
I should mention that t as I tried to rebuild with the new changes it failed. I  got several errors like this:
...... : error: tag not supported in the generated HTML version: tt
that "tt" refers to the <tt> </tt> tags that are no longer good, I had mentioned it elsewhere and though I had fixed a few and missed others. I do have a fix and will file an issue and send a pull request.

emaldona:
Oops, I was in the wrong directory - no errors. Will fix the warnings anyway later on.

cipherboy:
Okay, verbal ack from @emaldona and I'm fine with these changes. Pushing.

```
                     │21:26:15       emaldonado | anyway, the page seems okay to me, what do you think?   │
                     │21:26:32       emaldonado | any modifications you may wish for?                     │
                     │21:27:12        cipherboy | Yeah, looks good I think.                               │
                     │21:27:44       emaldonado | yes, make a comment to the effect that we are happy     │
                     │                          | with it                                                 │

```

PR #62, opened 2018-10-23
Documentation update

cipherboy:
After #59, this tries to make the javadocs prettier by adding a `-windowtitle` option and moves the `overview.html` to a different location. 

@emaldona Thoughts?


emaldona:
I like the change, similar to the prior version I had but better as the overview.html looks cleaner and tools/javadoc/ is a better location for it.  Feel free to merge them.

cipherboy:
Thanks @emaldona!

PR #63, opened 2018-10-24
[RFC] Increment version number to 4.5.1 post-release

cipherboy:
Since 4.5.0 was released by @edewata, it'd be nice to bump the version to the next release version, 4.5.1 such that users who build JSS can install it on a system 4.5.0 without needing to do this themselves. In general I'm in favor of "release+increment post release" versus "increment during release" but that's my 2c.

This was based on the commit by @edewata: 3ee3c9d1637a88c5dde5a93521680e42ba0aeefe
Built an unofficial COPR image to begin testing other Dogtag projects with: https://copr.fedorainfracloud.org/coprs/cipherboy/jss/build/814786/ 

Thoughts? /cc @edewata, @SilleBille, @emaldona, @jmagne

### DIFF
--- a/jss.spec
+++ b/jss.spec
@@ -6,7 +6,7 @@ Summary:        Java Security Services (JSS)
 URL:            http://www.dogtagpki.org/wiki/JSS
 License:        MPLv1.1 or GPLv2+ or LGPLv2+
 
-Version:        4.5.0
+Version:        4.5.1
 Release:        1%{?_timestamp}%{?_commit_id}%{?dist}
 # global         _phase -a1
 
--- a/org/mozilla/jss/CryptoManager.java
+++ b/org/mozilla/jss/CryptoManager.java
@@ -984,7 +984,7 @@ public final class CryptoManager implements TokenSupplier
 
 
     public static final String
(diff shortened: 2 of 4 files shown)


SilleBille:
Can you explain a bit more about "without needing to do this themselves" ?

Also, by "post-release" are you trying to imply that increment the release upstream but don't build packages?

cipherboy:
@edewata Ah thanks :) I missed the `--with-timestamp` and `--with-commit-id` options; they appear to do what I want. I'll close this PR then.

cipherboy:
@SilleBille Yes, post-release is the process I'm most familiar with:

- [OpenSCAP](https://github.com/OpenSCAP/openscap/commit/388ee7d56e224d5d1b971135c7a3f847459113c4)
- [MIT/krb5](https://github.com/krb5/krb5/commit/fc4a601f6c238c5b2ee8a078f573f7645ee6edfa) -- note that 1.17 isn't _yet_ released but is about to be released, so they've already bumped the version to 1.18.
- Some GNOME projects like [Evolution](https://github.com/GNOME/evolution/commit/8e68c3cb3b41e41b84ce8c42511d3009ead7a617)

I think this is mostly personal preference of the maintainers though.

PR #64, opened 2018-10-24
Reproducible JAR builds

cipherboy:
This PR adds support for reproducible builds in two parts:

- In `build_java.pl` and `tests/all.pl`, inconsistent use was made for using `xpclass.jar` directly as a literal or as a variable. I've updated this such that the variables set in `rules.mk` (which are passed to the `build_java.pl` script) control the jar name throughout building and testing. This allows us to change the `xpclass.jar` name at a later point if we desire.
- Implement reproducible JAR builds via a shell script, `tools/reproducible_jar.sh`. This is added as a make target, `reproducible`, a test target of `reproducibleCheck`, and then added to the `jss.spec` for use in Fedora.

To validate that this build really is reproducible:

```bash
cd sandbox/jss
rm ../dist -rf
make clean all reproducible
md5sum ../dist/*.jar
sha256sum ../dist/*.jar
```

and note the values / save them to a file. Then repeat all steps (including removing `dist`) and compare the new values.

### DIFF
--- a/build_java.pl
+++ b/build_java.pl
@@ -346,6 +346,8 @@ MyLabel
     our $javac_deprecation_flag;
     our $classpath;
     our $jar;
+    our $class_jar;
+
     if( scalar(@source_list) > 0 ) {
         ensure_dir_exists($class_dir);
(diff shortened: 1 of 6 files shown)


cipherboy:
Thanks folks! Merging. :)

Issue #65, opened 2018-10-25
javadoc generation will fail when time comes to switch to building with jdk11


emaldona:
The use of `<tt>/</tt>` causes failures with jdk11 tools
see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt
This is not a problem with either jdk8 or jdk10 only with jdk11

emaldona:
Replace that '/' with less-than-symbol + tt + greater-than-symbol

cipherboy:
I've updated the comments :) GitHub preserved your original text so I was able to wrap it in backticks and have it just work.

PR #66, opened 2018-10-25
Fix javadoc reported errors when switching to jdk11

emaldona:
- Address: https://github.com/dogtagpki/jss/issues/65
- use of `<tt>/</tt>` tags causes failures with jdk11 tools
- see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tt
- not a problem with either jdk8 or jdk10

### DIFF
--- a/org/mozilla/jss/ssl/SSLSocket.java
+++ b/org/mozilla/jss/ssl/SSLSocket.java
@@ -1596,7 +1596,7 @@ public class SSLSocket extends java.net.Socket {
      * Returns a list of cipher suites that are implemented by NSS.
      * Each element in the array will be one of the cipher suite constants
      * defined in this class (for example,
-     * <tt>TLS_RSA_WITH_AES_128_CBC_SHA</tt>).
+     * <code>TLS_RSA_WITH_AES_128_CBC_SHA</code>).
      */
     public static native int[] getImplementedCipherSuites();
 }
--- a/org/mozilla/jss/CryptoManager.java
+++ b/org/mozilla/jss/CryptoManager.java
@@ -1003,11 +1003,11 @@ public final class CryptoManager implements TokenSupplier
      * no means of specifying which token to use.
      *
      * <p>If no token is set, the InternalKeyStorageToken will be used. Setting
-     * this thread's token to <tt>null</tt> will also cause the
+     * this thread's token to <code>null</code> will also cause the
      * InternalKeyStorageToken to be used.
      *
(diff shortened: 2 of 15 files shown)


cipherboy:
So reading the attached Mozilla article, they do use `<code>` in a few places. Could we perhaps trade our `<tt>` elements for `<code>` blocks?

emaldona:
Sure, I'll use that. Thanks for looking that up.

emaldona:
I should add that when one is dealing with javadoc changes that need to be work for jdk11 and prior jdk's, such as 8, one uses "sudo alternatives ... " to switch, besides java and javac, also do it switch javadocdir. It makes a difference as jdk11 generates additional files such as several package-frame.html and some javascript ones. Perhaps we should add some note regarding this to the instructions given for package maintainers and contributors who likely need to support multiple versions of openjdk.

emaldona:
Noticed that the PR makes a reference to #65 but should be to #66 instead, and error in the commit comments. It can be confusing. @cipherboy do you know an easy way to fix that?

cipherboy:
Sure that sounds like a good idea. I'll draft some docs and have you review
them after merging this one.

emaldona:
Okay, trying git commit --amend

org/mozilla/jss/ssl/SSLSocket.java
```diff
@@ -1596,7 +1596,7 @@ private static native boolean isFipsCipherSuiteNative(int ciphersuite)
      * Returns a list of cipher suites that are implemented by NSS.
      * Each element in the array will be one of the cipher suite constants
      * defined in this class (for example,
-     * <tt>TLS_RSA_WITH_AES_128_CBC_SHA</tt>).
+     * <code>TLS_RSA_WITH_AES_128_CBC_SHA</code>).
```
cipherboy: I force pushed a change reverting the parenthesis to the original location.

emaldona: The changes you made are fine, ACK's feel free to push. Thanks.

emaldona:
In my comment at the top it shows Address: #65 but it should be Address: #66, I can't edit that part.   @cipherboy could you please fix it? Thanks.

cipherboy:
I think #65 is correct. That's the issue you created that's associated with this change, correct? #66 is the Pull Request, #65 is an issue.

emaldona:
You are right. It's the PR that has the 66 which confused me.

PR #68, opened 2018-10-30
Fedora 29 was released!

cipherboy:
As promised... [Fedora 29 is here](https://fedoramagazine.org/announcing-fedora-29) :)

Celebrate by kicking Fedora 27 out of the Travis images and adding Fedora 29 in. Keeping the Docker file around for now in case we decide we want to keep it later or want it for testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,8 @@ stages:
   - extra
 
 env:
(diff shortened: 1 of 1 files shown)


cipherboy:
Merging as it is a CI-only change.

PR #69, opened 2018-10-31
DER encode error for zero-valued enumerated types

cipherboy:
When an enumerated type has value zero, it is incorrectly given a length
of zero. Per X609, Section 8.4:

    The encoding of an enumerated value shall be that of the integer value
    with which it is associated. NOTE - It is primitive.

An integer is always encoded as a pair <length>+<value>, with value
always being specified, and thus <length> is always at least 1. However,
we were encoding enumerated as:

    0A 00

instead of the correct value:

    0A 01 00

This fixes the above error.

This error occurs because the range check on `length` allows it to reach zero, which occurs precisely when the value (`i`) is zero. This logic is correct in the negative case above, though negative zero does not occur in Java. This includes the original test case from the BZ along with a new test case that can be extended as necessary to test encoding bugs in the future.

RH BZ: 1582323
RH BZ URL: https://bugzilla.redhat.com/show_bug.cgi?id=1582323

### DIFF
--- a/org/mozilla/jss/netscape/security/util/DerOutputStream.java
+++ b/org/mozilla/jss/netscape/security/util/DerOutputStream.java
@@ -206,7 +206,7 @@ public class DerOutputStream
             }
         } else {
             // positive case
-            for (length = 4; length > 0; --length) {
+            for (length = 4; length > 1; --length) {
                 if ((i & bytemask) != 0)
                     break;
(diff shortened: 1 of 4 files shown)


cipherboy:
Seeing as this introduces new test cases and fixes the old test cases, I'm going ahead and merging this.

PR #72, opened 2018-11-09
Remove old, unsupported, or defunct platforms

cipherboy:
First of a series of patches which eventually prune our build tree down to the modern *nix variants. We're largely only targeting OpenJDK 8+ and there's no indication that these platforms have been tested or used recently. If someone wishes to support these, feel free to grab the files from this PR or the tree prior to their removal and re-contribute them. I'd also personally like to see CI support if a new system will be introduced.

This will eventually be followed up by build system changes that prune checks and shims for these platforms.

Note that a portion of these platforms don't exist any more and are defunct. Those are definitely safe to remove. Some of the other platforms still exist, but (as mentioned above) I'm inclined not to support them without community involvement who have the time and resources for testing new releases and/or CI, so I'm also removing them.

This does remove the out of date Windows support. I lack a Windows system to test it on, but I'm guessing it doesn't build currently. Again, community involvement here would be nice if anyone wishes to maintain Windows support and set up CI.


cipherboy:
CI failure fixed by #74.

cipherboy:
I'm closing this in favor of replacing the build system with CMake.

PR #73, opened 2018-11-09
Replace instances of loser label with finish

cipherboy:
I'm not a fan of the `loser` label, so I'm opting to replace it with `finish` instead as it is more clear.


cipherboy:
Updated with out-of-band feedback by @frozencemetery.

cipherboy:
Merging as this is a small change and passes CI.

PR #75, opened 2018-11-14
Remove unused variables from ssl/callbacks.c

cipherboy:
Note that `rc` is unused in both of these functions. This removes two compile warnings discovered while porting to CMake that were otherwise being lost in the build output.

### DIFF
--- a/org/mozilla/jss/ssl/callbacks.c
+++ b/org/mozilla/jss/ssl/callbacks.c
@@ -276,7 +276,6 @@ JSSL_AlertReceivedCallback(const PRFileDesc *fd, void *arg, const SSLAlert *aler
 {
     JSSL_SocketData *socket = (JSSL_SocketData*) arg;
 
-    jint rc;
     JNIEnv *env;
     jclass socketClass, eventClass;
     jmethodID eventConstructor, eventSetLevel, eventSetDescription;
@@ -336,7 +335,6 @@ JSSL_AlertSentCallback(const PRFileDesc *fd, void *arg, const SSLAlert *alert)
 {
     JSSL_SocketData *socket = (JSSL_SocketData*) arg;
 
-    jint rc;
     JNIEnv *env;
     jclass socketClass, eventClass;
     jmethodID eventConstructor, eventSetLevel, eventSetDescription;


cipherboy:
Ahh so this is a side effect of "smarter" compilers...

```c
/home/ascheel/GitHub/sandbox/jss/org/mozilla/jss/ssl/callbacks.c: In function ‘JSSL_AlertReceivedCallback’:
/home/ascheel/GitHub/sandbox/jss/org/mozilla/jss/ssl/callbacks.c:279:10: warning: variable ‘rc’ set but not used [-Wunused-but-set-variable]
     jint rc;
          ^~
/home/ascheel/GitHub/sandbox/jss/org/mozilla/jss/ssl/callbacks.c: In function ‘JSSL_AlertSentCallback’:
/home/ascheel/GitHub/sandbox/jss/org/mozilla/jss/ssl/callbacks.c:339:10: warning: variable ‘rc’ set but not used [-Wunused-but-set-variable]
     jint rc;
          ^~
```

The code in question looks like this:

```c
jint rc;
rc = some_call();
PR_ASSERT(rc == PR_OK);
```

With `PR_ASSERT` defined:

```c
#if defined(DEBUG) || defined(FORCE_PR_ASSERT)

#define PR_ASSERT(_expr) \
    ((_expr)?((void)0):PR_Assert(# _expr,__FILE__,__LINE__))

#define PR_NOT_REACHED(_reasonStr) \
    PR_Assert(_reasonStr,__FILE__,__LINE__)

#else

#define PR_ASSERT(expr) ((void) 0)
#define PR_NOT_REACHED(reasonStr)

#endif /* defined(DEBUG) || defined(FORCE_PR_ASSERT) */
```

So when `DEBUG` or `FORCE_PR_ASSERT` is defined, we clearly see a `PR_ASSERT(...)` call, but when `DEBUG` is not defined, we lack an implemented `PR_ASSERT(...)` call.

Note that since [`assert(...)`](http://www.cplusplus.com/reference/cassert/assert/) and [`PR_ASSERT(...)`](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSPR/Reference/PR_ASSERT) have similar semantics (up to the difference in macro definitions...), replacing `PR_ASSERT(...)` with `assert(...)` won't help us -- we'd still get a set-but-not-used error...

cipherboy:
I'm closing this as a lack of understanding on my part. At some point, if we want to compile with stricter GCC flags (e.g., `-Wall -Wextra -Werror`), we might consider adding a `EVAL_ASSERT` wrapper that ensure we at least evaluate the expression when not in `DEBUG` mode, but we can leave that for later.

PR #76, opened 2018-11-14
Remove unused JNI for NativeProxy

cipherboy:
While examining the current build system, it was noticed that this file both doesn't compile and isn't compiled. It refers to an `nativeUtil.h` which doesn't exist any more, and `NativeProxy.java` ironically, lacks any native/JNI calls.

Lets remove these files.


cipherboy:
After additional examination, no additional C files were included in the source tree but not built, outside of the files in `coreconf/` which will disappear with the introduction of the CMake build system.

cipherboy:
Merging as this passes CI and has been verified to not be needed and contains all changes necessary.

PR #77, opened 2018-11-14
Remove unused class KeyFactorySpi1_4

cipherboy:
Note that this file has been relatively untouched since the
reorganization and is excluded from the build.

Additionally, I verified that `KeyFactorySpi1_2` is used in a few places and that `KeyFactorySpi1_4` (the file being removed) is not included by either TomcatJSS, PKI.


edewata:
I'm not familiar with this code. @jmagne, could you take a look at this? Thanks.

cipherboy:
Yep. For future reference, a permanent link is available here: https://github.com/dogtagpki/jss-archive/blob/master/org/mozilla/jss/provider/java/security/KeyFactorySpi1_4.java

cipherboy:
Thanks @jmagne! Merging :)

PR #78, opened 2018-11-15
[RFC] New CMake build system

cipherboy:
This PR introduces a new CMake build system.

#### Description

There are three commits:
1. Add CMake files to the .gitignore.
2. Add the new CMake build system.
3. Update `jss.spec` to build JSS with the new build system.

Commit 2 is by far the largest. It contains the following files:

- `CMakeLists.txt` -- The core CMake file; this is a high-level file which describes how JSS gets built. Required for every CMake project.
- `cmake/FindNSPR.cmake` -- Copied from [dogtagpki/pki:master](https://github.com/dogtagpki/pki/blob/master/cmake/Modules/FindNSPR.cmake) and extended to support Debian testing.
- `cmake/FindNSS.cmake` -- Copied from [dogtagpki/pki:master](https://github.com/dogtagpki/pki/blob/master/cmake/Modules/FindNSS.cmake) and extended to support Debian testing.
- `cmake/JSSCommon.cmake` -- Core of the new build system; this builds all Java files in one pass (including generating JNI output -- `jss_build_java`), generates an includes directory for in-source header files (`jss_build_includes`), generates object files from the C-source files and links it into a library (`jss_build_c`), generates a jar (`jss_build_jar`), and generates a javadoc target (`jss_build_javadocs`).
- `cmake/JSSConfig.cmake` -- this file parses the environment and sets useful CMake variables; this includes configuring the JSS version, creating output directories, setting `CFLAGS` and `LDFLAGS` contents, and finding Java JARs.
- `cmake/JSSTests.cmake` -- this file is the test suite, implemented in CTest for parallel execution.
    - Note that some tests (namely JDK8-only and client/server tests aren't yet implemented in CTest.
- `lib/MANIFEST.MF.in` -- template for JSS JAR's manifest; previously created by `build_java.pl`.
- `lib/jss.map` -- copied from `lib/<LIBARY_PATH>/jssmap.linux` -- (`jss/lib/Linux4.18_x86_64_glibc_PTH_64_DBG.OBJ/jssmap.linux` on my system). For use as a linker script. 

#### Reviewing

The files under `cmake/JSS*.cmake` and the changes to `jss.spec` deserve the most review.

#### Testing

The suggested testing route is:

1. Clone and Rebase JSS:
    ```bash
    git clone https://github.com/dogtagpki/jss && cd jss
    git remote add cipherboy https://github.com/cipherboy/jss
    git checkout cipherboy/cmake-buildsystem
    git rebase -i master
    ```
2. Build via CMake:
    ```bash
    cd jss
    rm -rf build && mkdir build && cd build
    cmake .. && make all test
    ```
3. Build via old build system:
    ```bash
    cd jss
    source tools/autoenv.sh
    make clean all check
    ```
4. Build the RPM:
    ```bash
    cd jss
    ./build.sh --with-commit-id --with-timestamp
    ```
Note that swapping steps 2 and 3 won't work as CMake generates files which the old JSS build system depends on. 

#### Future Work

Later PRs will:

1. Extend CI to use this new build system.
2. Remove the old build system.
3. Document this build system better via source comments and update documentation in `docs/`.
4. Add client/server tests into the CTest environment.

#### ACKs

I'd like 2-3 acks from @edewata, @jmagne, @SilleBille, @frasertweedale and/or @ladycfu.


cipherboy:
Note that the Optional CI images failing is expected as they use the old build system and aren't running `cmake` prior to build -- thus files are missing, per the `Testing` section note above. This is fixed in #79.

cipherboy:
@edewata had some feedback about version numbers, so I'll take care of some of that via JNI in a separate PR before revisiting this one.

cmake/JSSTests.cmake
```diff
@@ -0,0 +1,200 @@
...
+        "Mozilla_JSS_JCA_Signature_FIPSMODE" # DEPENDS
+    )
+    jss_test_java(
+        "Disable_FipsMODE"
+        "org.mozilla.jss.tests.FipsTest;${RESULTS_FIPS_OUTPUT_DIR};disable"
+        "JSS_Signature_test_FipsMODE" # DEPENDS
+    )
+endmacro()
+
+function(jss_test_java TEST_NAME TEST_COMMAND)
```
edewata: This is fine but for future improvement it would be nice if we can use keywords to separate the arguments like this:
```
jss_test_java(
    NAME
        Disable_FipsMODE"
    COMMAND
        "org.mozilla.jss.tests.FipsTest;${RESULTS_FIPS_OUTPUT_DIR};disable"
    DEPENDS
        JSS_Signature_test_FipsMODE
)
```

edewata: Here's an example:
https://github.com/dogtagpki/pki/blob/master/cmake/Modules/Java.cmake#L21
But probably it can be done more easily with CMakeParseArguments().

edewata: Looks good!

cipherboy: Thanks for the suggestion, didn't know that existed and makes it more readable. :)

cipherboy:
This has also passed a re-run of PKI-nightly:

- https://travis-ci.org/dogtagpki/pki-nightly-test/builds/456801237
- https://travis-ci.org/dogtagpki/pki-nightly-test/jobs/456801238
- https://copr.fedorainfracloud.org/coprs/g/pki/10.6-nightly/build/826836/
- https://copr-be.cloud.fedoraproject.org/results/@pki/10.6-nightly/fedora-29-x86_64/00826836-jss/build.log.gz

PR #84, opened 2018-11-19
cmake in-source documentation

cipherboy:
Adds more extensive source-level documentation about CMake to the CMake build itself. This also fixes a few minor issues, such as better parallelizing the test suite and missing dependencies during build.

Acceptance Criteria:
- [x] spell check
- [x] accuracy of documentation

PR #86, opened 2018-11-20
Add ClientAuth tests to ctest

cipherboy:
Note that, per [jss issue#30](https://pagure.io/jss/issue/30), these fail under `FUTURE` crypto policy. 

Adds SSLClientAuth tests to the ctest test suite. These were present in `all.pl` at time of moving to CMake, but weren't yet added. This adds them back in.

Note that this PR is dependent upon #84, and includes commits in that PR.


cipherboy:
Merging as this adds back in a previously existing test, and it passes CI.

PR #87, opened 2018-11-20
DER Encoding to Ctest

cipherboy:
Adds DER Encoding tests to the ctest test suite. These were present in all.pl at time of moving to CMake, but weren't yet added. This adds them back in.

Note that this PR is dependent upon #84, and includes commits in that PR.


cipherboy:
Merged as this passes CI and introduces tests previously in `all.pl`.

PR #88, opened 2018-11-20
Refactor CMake to support building other C targets

cipherboy:
In preparation for a C language test, this refactors CMake so that it
can support building other C targets besides the `libjss4.so` library.
This will allow us to build stand-alone executables which don't depend
on `libjss4.so` but could depend on the NSS/NSPR libraries.

### DIFF
--- a/cmake/JSSCommon.cmake
+++ b/cmake/JSSCommon.cmake
@@ -79,7 +79,7 @@ macro(jss_build_c_file C_FILE C_OUTPUT C_TARGET C_DIR)
     # finished, else many headers wouldn't exist.
     add_custom_command(
         OUTPUT "${C_OUTPUT}"
-        COMMAND ${CMAKE_C_COMPILER} ${JSS_C_FLAGS} -o ${C_OUTPUT} ${C_FILE}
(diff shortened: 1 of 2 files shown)


cmake/JSSCommon.cmake
```diff
@@ -118,7 +118,7 @@ macro(jss_build_c)
 
     add_custom_command(
         OUTPUT "${JSS_SO_PATH}"
-        COMMAND ${CMAKE_C_COMPILER} -o ${JSS_SO_PATH} ${LIB_OUTPUT_DIR}/*.o ${JSS_LD_FLAGS}
+        COMMAND ${CMAKE_C_COMPILER} -o ${JSS_SO_PATH} -shared ${LIB_OUTPUT_DIR}/*.o ${JSS_LD_FLAGS} ${JSS_LIBRARY_FLAGS}
```
cipherboy: `-shared` is duplicated here and in `JSS_LIBRARY_FLAGS`

cipherboy:
This looks good to me and passes tests, so I'm going ahead and merging it.

PR #90, opened 2018-11-26
Add build directory

cipherboy:
As reported by @emaldona, the `build` directory doesn't exist in the default `git` source.

Two options here:

1. Update the README to create the build directory
2. Add the build directory to the `git` source.

My thought is go with the second, to mirror what other projects such as [`SSG`](https://github.com/ComplianceAsCode/content/pull/1377/files#diff-b1bbf4eded3f398323c5fe00fb1013ce) have done in the past. 


cipherboy:
The only problem with `rm -rf build/*` is that `build/.targets` is a directory that should be removed as well. That contains pseudo-targets to plaster over a quirk of the CMake build system (`add_custom_command` needs to be used in conjunction with `add_custom_target` if you want to do parallel builds). If the build directory (or more precisely `build/.gitkeep`) is removed, this doesn't meaningfully impact the build, and only affects the ability to do `git add -A` without having to reset the changes to `build/.gitkeep` / `touch build/.gitkeep` afterwards.

PR #91, opened 2018-11-26
Update docs/pkcs11constants.md according to new build system

emaldona:
Address: https://pagure.io/jss/issue/31


cipherboy:
This is fine and needs to be done, but we also need to add this test back in to JSS, so I'll take that as a AI after merging.

PR #92, opened 2018-11-27
Bump keysizes to 4096

cipherboy:
This is required if we wish to run the JSS test suite on a system
with the `FUTURE` cryptopolicy currently enforcing. In particular, key
sizes of 1024 and 2048 are too small for later use by NSS, causing test
failures. Note that `FUTURE` currently sets a minimum key length for RSA
of 3072 bits.

Part of this (changing `GenerateTestCert`) is required for the later work on SSLEngine, and for reintroducing the Client/Server SSL tests to JSS. Currently a 1024-bit RSA certificate is created, which is too small for any (except LEGACY?) policy.

This is related to [jss issue#30](https://pagure.io/jss/issue/30). 


cipherboy:
I don't believe there are any other places where key sizes are used in JSS in association with crypto policy at the moment, so I'm going ahead and merging this. 

Note that, when we reintroduce the client/server tests, many of them use their own keys so we'll have to up the sizes there as well.

PR #93, opened 2018-11-27
Implement a NSPR PRFileDesc using buffers - 0:1

cipherboy:
~This PR is blocked on #103 and depends on commits in it.~

In order to support the `javax.net.ssl.SSLEngine` interface, we need to implement SSL-over-buffers instead of SSL over Sockets. This is the first step towards that: implementing a `PRFileDesc` which NSS can wrap with SSL to provide read/write methods. This can then be used in Java and then used to implement SSLEngine.

There are two commits here:

- Add the `PRFileDesc` implementation as a new header
- Use this new header in a test to ensure that it provides the requisite features.

Much of this code was previewed in [`cipherboy/testbed/c-prfd`](https://github.com/cipherboy/testbed/tree/master/c-prfd) before being cleaned up and introduced to JSS.

### DIFF
--- a/cmake/JSSTests.cmake
+++ b/cmake/JSSTests.cmake
@@ -162,6 +162,11 @@ macro(jss_tests)
         COMMAND "org.mozilla.jss.tests.SSLClientAuth" "${RESULTS_NSSDB_OUTPUT_DIR}" "${PASSWORD_FILE}" "${JSS_TEST_PORT_CLIENTAUTH}" "50"
         DEPENDS "List_CA_certs"
     )
+    jss_test_exec(
+        NAME "TestBufferPRFD"
+        COMMAND "${BIN_OUTPUT_DIR}/TestBufferPRFD" "${RESULTS_NSSDB_OUTPUT_DIR}" "${DB_PWD}"
+        DEPENDS "List_CA_certs" "generate_c_TestBufferPRFD"
+    )
     jss_test_java(
         NAME "Key_Generation"
         COMMAND "org.mozilla.jss.tests.TestKeyGen" "${RESULTS_NSSDB_OUTPUT_DIR}" "${PASSWORD_FILE}"
@@ -306,6 +311,7 @@ endmacro()
 macro(jss_tests_compile)
     jss_tests_compile_c("${PROJECT_SOURCE_DIR}/org/mozilla/jss/tests/buffer_size_1.c" "${BIN_OUTPUT_DIR}/buffer_size_1" "buffer_size_1")
     jss_tests_compile_c("${PROJECT_SOURCE_DIR}/org/mozilla/jss/tests/buffer_size_4.c" "${BIN_OUTPUT_DIR}/buffer_size_4" "buffer_size_4")
+    jss_tests_compile_c("${PROJECT_SOURCE_DIR}/org/mozilla/jss/tests/TestBufferPRFD.c" "${BIN_OUTPUT_DIR}/TestBufferPRFD" "TestBufferPRFD")
 endmacro()
 
 macro(jss_tests_compile_c C_FILE C_OUTPUT C_TARGET)
--- a/org/mozilla/jss/ssl/javax/BufferPRFD.h
+++ b/org/mozilla/jss/ssl/javax/BufferPRFD.h
@@ -1,0 +1,32 @@
+#include <nspr.h>
+
+/* Necessary for correctly propagating E_WOULDBLOCK. */
+#include <errno.h>
+
+#include <stdlib.h>
+#include <string.h>
+
+/* Ring buffer implementation to handle read/write calls. */
+#include "buffer.h"
+
+/* Use a modern pragma guard... */
+#pragma once
+
+/* Free a Buffer-backed PRFileDesc. Note that it is usually sufficient to call
+ * PR_Close(...) on the buffer instead. This is provided for completeness and
+ * should not be called in a SSL context as the buffer PRFileDesc is wrapped
+ * by the SSL PRFileDesc. Note that this only removes references to the
+ * underlying j_buffers and does not free them; it is up to the caller to
+ * do so. */
+void freeBufferPRFileDesc(PRFileDesc *fd);
+
+/* Construct a new PRFileDesc backed by a pair of buffers. Note that these
+ * should be separate buffers, but need not be unique to this PRFileDesc;
+ * that is, a client and server could share (but be swapped) j_buffers.
+ * The caller is expected to provide a peer_info to be used for optional
+ * session resumption; this will be truncated at 16 bytes of data, and
+ * extended with nulls if it is shorter. It is suggested that this be an IPv4
+ * or IPv6 address. Note that this value is not used to validate the hostname
+ * in any way (see SSL_SetURL to validate the peer). */
+PRFileDesc *newBufferPRFileDesc(j_buffer *read_buf, j_buffer *write_buf,
+    uint8_t *peer_info, size_t peer_info_len);
--- a/org/mozilla/jss/tests/TestBufferPRFD.c
+++ b/org/mozilla/jss/tests/TestBufferPRFD.c
@@ -1,0 +1,418 @@
+/*
+ * Test case for Buffer PRFileDesc implementation located under the
+ * org.mozilla.jss.ssl.javax package. This ensures that we can do a
+ * basic SSL handshake and verify that it works alright.
+ */
+
+/* Optional, for enabling asserts */
+#define DEBUG 1
+
+/* Header file under test */
+#include "BufferPRFD.h"
+
+/* NSPR required includes */
+#include <prio.h>
+#include <prlog.h>
+#include <prmem.h>
+#include <prnetdb.h>
+
+/* NSS includes */
+#include <nss.h>
+#include <ssl.h>
+#include <pk11pub.h>
+#include <cert.h>
+#include <certdb.h>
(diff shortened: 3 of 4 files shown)


org/mozilla/jss/ssl/javax/BufferPRFD.h
```diff
@@ -0,0 +1,344 @@
...
+        // So we don't dereference a NULL pointer in case we've been closed
+        // multiple times, exit with success.
+        return PR_SUCCESS;
+    }
+
+    if (fd->secret != NULL) {
+        /* In case the data on the wire is sensitive (it shouldn't be -- all
+         * incoming and outgoing data should be encrypted except for the
+         * client hello -- memset the private structure first. */
+        memset(fd->secret, 0, sizeof(PRFilePrivate));
```
cipherboy: This comment and `memset` call are actually out of date -- the structure itself only contains pointers to data and not the data itself. I think the `memset` call is extraneous and only ensures that our internal pointers are null in case anyone maintains a pointer to it.

We _can't_ actually free or zero the internal pointers as we don't own them and there may still be data (e.g., a `CLOSE_NOTIFY` event) that needs to be sent.

org/mozilla/jss/ssl/javax/BufferPRFD.h
```diff
@@ -0,0 +1,344 @@
...
+     * Because we're implementing this as a (large) buffer, copy from the
+     * buffer if possible, else ignore it -- never error. */
+    PRFilePrivate* internal = fd->secret;
+
+    if (*internal->read_ptr > 0) {
+        size_t read_len = amount;
+        uint8_t* offset = internal->read_bytes;
+        size_t memmove_len = *internal->read_ptr - amount;
+
+        if (read_len > (*internal->read_ptr)) {
```
cipherboy: We could document the internals of this a bit more.

cipherboy:
\o/ Fedora stalled.

cmake/JSSTests.cmake
```diff
@@ -202,6 +209,30 @@ macro(jss_tests)
...
+macro(jss_tests_compile)
+    jss_tests_compile_c("${PROJECT_SOURCE_DIR}/org/mozilla/jss/tests/TestBufferPRFD.c" "${BIN_OUTPUT_DIR}/TestBufferPRFD" "TestBufferPRFD")
+endmacro()
+
+macro(jss_tests_compile_c C_FILE C_OUTPUT C_TARGET)
+    # Generate the target executable from C_FILE
+
+    add_custom_command(
+        OUTPUT "${C_OUTPUT}"
+        COMMAND ${CMAKE_C_COMPILER} ${JSS_C_FLAGS} -o ${C_OUTPUT} ${C_FILE} ${JSS_LD_FLAGS}
```
cipherboy: Ubuntu was using an older GCC compiler, which strictly requires linker flags to be at the end.

org/mozilla/jss/ssl/javax/BufferPRFD.h
```diff
@@ -0,0 +1,305 @@
...
+}
+
+// Fake responses to getSocketOption requests
+static PRStatus PRBufferGetSocketOption(PRFileDesc *fd, PRSocketOptionData *data)
+{
+    /* getSocketOption takes a PRFileDesc and modifies the PRSocketOptionData
+     * with the options on this. We set a couple of sane defaults here:
+     *
+     *   non_blocking = true
+     *   reuse_addr = false
```
cipherboy: Perhaps suggest this should be true.

org/mozilla/jss/ssl/javax/BufferPRFD.h
```diff
@@ -0,0 +1,305 @@
...
+    /* If fd->secret is none, close() was called first so we can just exit. */
+    if (fd->secret == NULL) {
+        return;
+    }
+
+    fd->secret->read_buffer = NULL;
+    fd->secret->write_buffer = NULL;
+}
+
+PRFileDesc *newBufferPRFileDesc(j_buffer *read_buf, j_buffer *write_buf,
```
cipherboy: Add documentation here.

org/mozilla/jss/ssl/javax/BufferPRFD.h
```diff
@@ -0,0 +1,305 @@
...
+    (PRSetsocketoptionFN)PRBufferSetSocketOption,
+    (PRSendfileFN)invalidInternalCall,
+    (PRConnectcontinueFN)invalidInternalCall,
+    (PRReservedFN)invalidInternalCall,
+    (PRReservedFN)invalidInternalCall,
+    (PRReservedFN)invalidInternalCall,
+    (PRReservedFN)invalidInternalCall
+};
+
+void freeBufferPRFileDesc(PRFileDesc *fd)
```
cipherboy: Add documentation here.

org/mozilla/jss/ssl/javax/BufferPRFD.h
```diff
@@ -0,0 +1,305 @@
...
+     * reference to it as well. */
+
+    /* If fd->secret is none, close() was called first so we can just exit. */
+    if (fd->secret == NULL) {
+        return;
+    }
+
+    fd->secret->read_buffer = NULL;
+    fd->secret->write_buffer = NULL;
+}
```
cipherboy: `fd->secret = NULL` missing.

org/mozilla/jss/tests/TestBufferPRFD.c
```diff
@@ -0,0 +1,454 @@
...
+    if (newfd == NULL) {
+        const PRErrorCode err = PR_GetError();
+        fprintf(stderr, "error: NSPR error code %d: %s\n",
+            err, PR_ErrorToName(err));
+        exit(1);
+    }
+
+    model = newfd;
+    newfd = NULL;
+    if (SSL_OptionSet(model, SSL_ENABLE_SSL2, PR_FALSE) != SECSuccess) {
```
cipherboy: Figure out which set of these are default and remove them.

org/mozilla/jss/tests/TestBufferPRFD.c
```diff
@@ -0,0 +1,454 @@
...
```
cipherboy: As above; check which can be removed because they are default.

org/mozilla/jss/tests/TestBufferPRFD.c
```diff
@@ -0,0 +1,454 @@
...
+        const PRErrorCode err = PR_GetError();
+        fprintf(stderr, "error: SSL_SetURL error %d: %s\n",
+            err, PR_ErrorToName(err));
+        exit(1);
+    }
+
+    return s_nspr;
+}
+
+bool is_finished(PRFileDesc *c_nspr, PRFileDesc *s_nspr)
```
cipherboy: Add doc text.

org/mozilla/jss/tests/TestBufferPRFD.c
```diff
@@ -0,0 +1,454 @@
...
+
+    ret = PR_Read(s_nspr, buf2, buf_size - 1);
+    if (ret < 0) {
+        const PRErrorCode err = PR_GetError();
+        fprintf(stderr, "error: PR_Read error %d: %s\n",
+            err, PR_ErrorToName(err));
+        exit(1);
+    }
+
+    memset(buf, 0, buf_size);
```
cipherboy: Could drop this if we wanted to; print `buf2` instead.

org/mozilla/jss/tests/TestBufferPRFD.c
```diff
@@ -0,0 +1,454 @@
...
+
+    ret = PR_Read(c_nspr, buf2, buf_size - 1);
+    if (ret < 0) {
+        const PRErrorCode err = PR_GetError();
+        fprintf(stderr, "error: PR_Read error %d: %s\n",
+            err, PR_ErrorToName(err));
+        exit(1);
+    }
+
+    memset(buf, 0, buf_size);
```
cipherboy: Same as above; print `buf2` instead.

org/mozilla/jss/tests/TestBufferPRFD.c
```diff
@@ -0,0 +1,413 @@
...
+        const PRErrorCode err = PR_GetError();
+        fprintf(stderr, "error: PR_Shutdown server error %d: %s\n",
+            err, PR_ErrorToName(err));
+        exit(1);
+    }
+
+    /* Closes the underlying POSIX file descriptors */
+    PR_Close(c_nspr);
+    PR_Close(s_nspr);
+
```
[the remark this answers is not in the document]
cipherboy: Failure cases are handled by `exit(1)` -- intentional choice to throw away all memory management.

org/mozilla/jss/tests/TestBufferPRFD.c
```diff
@@ -0,0 +1,413 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <strings.h>
+
+static char *return_password(PK11SlotInfo *slot, PRBool retry, void *arg) {
+    /* Return the password passed in via arg as the password for the PKCS11
+     * slot. From NSS semantics it appears that this function should allocate
+     * a new copy with strdup as the caller expects to free it. */
+    if (retry == PR_FALSE) {
+        return strdup((char*) arg);
```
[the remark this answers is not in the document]
cipherboy: We'll leave this for later if static analyzers are picky about it.

PR #94, opened 2018-11-27
Reintroduce PKCS11 Constants test

cipherboy:
When switching to CMake in #78, this was one of the tests dropped from the new test system. This PR adds it back in and addresses the AI from @emaldona's #91. 


cipherboy:
As this passes CI and otherwise reintroduces code that was already present, I'll go ahead and merge it.

PR #67, opened 2018-10-30
Sync JSS with NSS's TLS cipher suite lists

cipherboy:
This includes a few new TLS 1.3 ciphers, TLS names for old SSL ciphers,
and RFC5487 support.

I'd like review on a few points:

1. I'm not sure if the omission of `TLS_NULL_WITH_NULL_NULL` was intentional or not, but I've gone ahead and added it.
2. There are several new TLS-for-SSL names--should I deprecate the SSL names? E.g., JSS already has `SSL3_RSA_EXPORT_WITH_RC4_40_MD5` but I've added `TLS_RSA_EXPORT_WITH_RC4_40_MD5`.
3. The introduction of `TLS_EMPTY_RENEGOTIATION_INFO_SCSV` and `TLS_FALLBACK_SCSV` -- I'm not sure if they were intentionally missing.
4. Adding them to the test suite.

`Signed-off-by: Alexander Scheel <ascheel@redhat.com>`

PR #96, opened 2018-11-27
Add TLS_DHE_DSS_WITH_AES_256_CBC_SHA256 to tests

cipherboy:
This was omitted from the earlier TLS sync; add it back in. It was detected while trying to re-enable the client/server NSS tests. 

This PR is kinda icky because we have to renumber a bunch of items and thus it isn't immediately obvious what actually changed. I also removed the trailing space on the ones I actually touched.

I'm not sure why a bunch of these are duplicated (e.g., `TLS_DHE_RSA_WITH_AES_128_CBC_SHA256`), but I kept the pattern with the new addition, `TLS_DHE_DSS_WITH_AES_256_CBC_SHA256` which was added in #67. 


cipherboy:
This looks fine, so I'm going ahead and merging. The diff could be minified by one line as element 53 actually remained the same but the spacing was modified anyways.

In the future we can replace this by using the contents of `SSLCipher` instead of `SSLSocket.<suite>`.

PR #97, opened 2018-11-28
CMake reproducible JARs

cipherboy:
When CMake was introduced, reproducible JAR builds were not part of the pull request. 

This reintroduces reproducible JAR builds in three commits:

 - Document the usage of `tools/reproducible_jar.sh`
 - Add reproducible JAR build back to CMake
 - Make `tools/reproducible_jar.sh` silent except on errors.


cipherboy:
Debian and Ubuntu lack `zip`/`unzip` functionality in the stock install, so I'll have to add this to the list of dependencies.

cipherboy:
That fixed the issue so I'm going ahead and merging as this was functionality previously introduced in #64.

PR #98, opened 2018-11-28
Separate test artifacts

cipherboy:
Under the old build system, it was not easily possible to split all build and test artifacts into separate locations. With CMake, it is possible and has been done.

This allows us to NOT ship our test cases as part of the JSS JAR packaged by most distros. Instead, a separate JAR could be shipped when necessary, say, as part of the main package or as part of the debug package. This helps also to ensure that no end users rely on our tests.

Would appreciate feedback on this from @edewata or @jmagne if they have time.

This PR is dependent upon #97, and includes commits from there; it will be rebased before merging.


cipherboy:
I have confirmed that no packages in dogtagpki or FreeIPA rely on `org.mozilla.jss.tests` packages, so it should be safe to remove from the standard JAR.

jmagne:
Question, does this need a mod to the spec file or does the spec file pickup the same entities, that just happen to now have the tests stuff in it?

cipherboy:
The following is in regard to the "combined" result of #98, #99, and #100. 

@jmagne said:
> Question, does this need a mod to the spec file or does the spec file pickup the same entities, that just happen to now have the tests stuff in it?

Assuming what you typed is what you meant, neither actually. The output of the non-test artifacts has not changed in path, so the `.spec` file needs no change. In concise terms, across these PRs:

1. We create a new JAR with the contents of `org.mozilla.jss.tests` (#98).
    New file, new location, so `.spec` need not change as the "old" file is in the same place.
    ("old" because it is "old contents" - `org.mozilla.jss.tests`).
    - From the perspective of the `.jar` inside the `.rpm`, we've lost the `org.mozilla.jss.tests` package.
2. We create a new shared library with all the contents visible (#99).
    New file, new location, so `.spec` need not change as the old library is in the same place.
    ("old" is exactly the same as the previous contents).
    - From the perspective of the `.rpm`, nothing has changed.
2. We can now build with debug or release mode; all artifacts are the same, just different compile/configure flags (#100).
    - From the perspective of the `.rpm`, nothing has changed.

In this PR, **if** we want to ship the tests, then yes, we'd need to modify the `.spec` file to include the new file (and I'd say, this PR slightly to create a better JAR). If we don't care about shipping the tests, then we can go ahead and merge this as-is.

Hopefully that answers your question?

cipherboy:
Per comments with @edewata and @jmagne, I'm going ahead and merging this. We don't think that anyone needs the testing classes currently and we can easily ship another RPM if someone wishes to have them easily available.

PR #99, opened 2018-11-28
Library testing

cipherboy:
This PR is dependent upon #97 and #98 and includes commits from both. Only the last commit is new.

In a future PR to test #93 and the SSLEngine work, we'd like to also be able to test C code which lives in `libjss4.so` but _isn't_ a JNI call. Thus, we require an unversioned `libjss4.so` or directly link to all of the objects we wish to test. If the objects we wish to test are complex and in turn depend on other objects, an unversioned `libjss4.so` is best; if however, these objects are always simple, directly compiling in the objects (from `build/lib/*.o`) would suffice.

In the test suite, we continue using the versioned library for Java calls, but use the unversioned library for directly executed programs: this ensures that any new tests we introduce for new methods will fail if these new methods aren't added to the version script as well. 

Thoughts?

### DIFF
--- a/cmake/JSSCommon.cmake
+++ b/cmake/JSSCommon.cmake
@@ -153,16 +153,23 @@ macro(jss_build_c)
         DEPENDS ${C_OUTPUTS}
     )
 
+    # We generate two libraries: build/lib/libjss.so and build/libjss.so:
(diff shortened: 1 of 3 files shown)


cipherboy:
This looks good and works for my use case so I'm going ahead and merging it.

PR #100, opened 2018-11-28
Enable release builds

cipherboy:
Respect `CMAKE_BUILD_TYPE` for Release vs Debug

This is the final step in enabling CMake builds to replace the old
functionality: specifying `-DCMAKE_BUILD_TYPE=Release` on the CMake
command line is equivalent to setting `BUILD_OPT=1` in the old build
system.

This has been verified by comparing the RPM builds from branch HEAD to the commit before the introduction of CMake and checking the contents of the JAR after extraction. 

Depends on commits from #98 and #99; will be rebased after they merge. 


cipherboy:
As this blocks #102, I'm going ahead and merging.

PR #101, opened 2018-11-29
Sync PKI and JSS common classes

cipherboy:
This PR depends on #104 and includes commits from that series; it will be rebased when that PR is merged.

Each commit in question includes two paths: the path under the JSS source repo and the path under the PKI source repo. This allows easy review: clone PKI at master, clone JSS at this PR's `HEAD`, and diff each of the changed files. The changes should be minimal, only changing associated imports and other fully qualified package names.

This will allow us to remove the code from PKI, release a new JSS version with the changes, make PKI depend on this new JSS version, and remove from PKI these classes. See: 

 - [PKI PR#121](https://github.com/dogtagpki/pki/pull/121) - which updates package paths (`netscape.security` -> `org.mozilla.jss.netscape.security`)
 - [PKI PR#122](https://github.com/dogtagpki/pki/pull/122) - which removes `netscape.security` from the PKI tree. 
 - [PKI PR#123](https://github.com/dogtagpki/pki/pull/123) - which removes the `getKeyWrapAlgorithmFromOID(...)` function from the PKI tree. This is implemented as part of #104. 

These combined commits (this PR and the three against PKI) have been tested with [`ansible-vms`](https://github.com/cipherboy/ansible-vms/blob/master/pki-integration.yml) and a CA set up. This ensures that PKI builds after these changes and is a basic sanity check to make sure PKI installs and runs.

A helpful script for minimizing differences is available [on my gist](https://gist.githubusercontent.com/cipherboy/07995403dd194b7e76bbbed62180cecf/raw/7516984cfaded4bc8028118cbe1559925c657c1a/filter_packages.sh). Usage is `filter_packages.sh > /tmp/output < pki/some_input_class.java`, and then you can run `diff /tmp/output jss/some_input_class.java` and have minimal differences.

### DIFF
--- a/org/mozilla/jss/tests/BigObjectIdentifier.java
+++ b/org/mozilla/jss/tests/BigObjectIdentifier.java
@@ -1,0 +1,54 @@
+package org.mozilla.jss.tests;
+
+import java.io.IOException;
+import java.math.BigInteger;
+
+import org.mozilla.jss.netscape.security.util.*;
+
+public class BigObjectIdentifier {
+    public static void main(String[] args) throws Exception {
+
+        long[] oid_components_long = { 1L, 3L,6L,1L,4L,1L,5000L,9L,1L,1L,1526913300628L, 1L};
+        int[] oid_components_int =  { 1, 3,6,1,4,1,2312,9,1,1,15269, 1, 1};
+        BigInteger[] oid_components_big_int = { new BigInteger("1"), new BigInteger("3"), new BigInteger("6"), new BigInteger("1"),
+            new BigInteger("4"), new BigInteger("1"), new BigInteger("2312"),
+            new BigInteger("9"), new BigInteger("1"),
+            new BigInteger("152691330062899999999999997777788888888888888889999999999999999"), new BigInteger("1")
+        };
+
+        String oidIn = "1.3.6.1.4.1.2312.9.1.152691330062899999999999997777788888888888888889999999999999999.1";
+        ObjectIdentifier oid = new ObjectIdentifier(oidIn);
+
+        ObjectIdentifier fromDer = null;
+        ObjectIdentifier fromStaticMethod = null;
+        ObjectIdentifier fromComponentList = null;
+        ObjectIdentifier  fromComponentListInt = null;
+        ObjectIdentifier fromComponentListBigInt = null;
+
+        System.out.println("oid: " + oid.toString());
+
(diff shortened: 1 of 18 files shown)


cipherboy:
This includes commits from #104 and is thus blocked on it. It will be rebased once #104 is merged.

org/mozilla/jss/tests/BigObjectIdentifier.java
```diff
@@ -0,0 +1,59 @@
+package org.mozilla.jss.tests;
+
+import java.io.IOException;
+import java.math.BigInteger;
+
+import org.mozilla.jss.netscape.security.util.*;
+
+public class BigObjectIdentifier {
+    public static void main(String[] args) {
```
edewata: Let's make main() throw an Exception.

org/mozilla/jss/tests/BigObjectIdentifier.java
```diff
@@ -0,0 +1,59 @@
...
+            fromComponentListInt = new ObjectIdentifier(oid_components_int);
+
+            System.out.println("fromComponentListInt: " + fromComponentListInt);
+
+            fromComponentListBigInt = new ObjectIdentifier(oid_components_big_int);
+
+            System.out.println("fromComponentListBigInt: " + fromComponentListBigInt);
+
+        } catch (IOException e) {
+            e.printStackTrace();
```
edewata: This try-catch block catches and swallows the exception, so the method will exit without error. I think it should rethrow the exception, or just remove the try-catch block completely. This way the caller will know that there's an error.

cipherboy:
OK I have tested this with the PKI related changes ([`cipherboy/pki:new-merged`](https://github.com/cipherboy/pki/tree/new-merged)) and this looks good to go. Note that this also requires [pki pr#128](https://github.com/dogtagpki/pki/pull/128) or something like it.

cipherboy:
I have tested this with the latest upstream [`candlepin`](https://github.com/candlepin/candlepin) code and it appears to build fine.

@edewata would you like to do another review, or should I go ahead and merge and tag 4.5.2?

cipherboy:
I've also gone ahead and validated that the [upstream source](https://github.com/dogtagpki/pki/commits/master/base/util/src/netscape) hasn't changed since this PR was originally authored, so I'm going ahead and merging it. 

Thanks for the review @edewata! :)

PR #106, opened 2018-12-10
Fix NSS deprecation warnings

cipherboy:
NSS has begun deprecating `key.h` and `keyt.h`, replacing them with `keyhi.h`
and `keythi.h` instead. Update our includes as appropriate.

I'm interested on if this passes CI (in particular, on tested Debian and Ubuntu versions). If it does, I'm inclined to push it in the 4.5.2 time frame, else I'll look into it more under the 4.5.3 time frame. 

### DIFF
--- a/org/mozilla/jss/CryptoManager.c
+++ b/org/mozilla/jss/CryptoManager.c
@@ -7,7 +7,7 @@
 #include <secmod.h>
 #include <cert.h>
 #include <certt.h>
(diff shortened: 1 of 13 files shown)


cipherboy:
This looks like it passes CI, so I'm leaning towards including it with 4.5.2. I'll dig up a minimum NSS version for this change and update the `.spec` file appropriately.

PR #113, opened 2019-01-17
Backport - Bump keysizes to 4096

cipherboy:
Backport of #92.

This is required if we wish to run the JSS test suite on a system
with the FUTURE cryptopolicy currently enforcing. In particular, key
sizes of 1024 and 2048 are too small for later use by NSS, causing test
failures. Note that FUTURE currently sets a minimum key length for RSA
of 3072 bits.


cipherboy:
Safe backport; only affects testing code.

PR #114, opened 2019-01-18
Backport CI Improvements

cipherboy:
This backports most of the CI improvements from `master` to `v4.4.x` and fixes the build on older Fedoras (pre F27) and CentOS 7. 

In particular:

 - Backport Containerized (`Dockerfile`) based CI
 - Fix `pki.spec.in` to not use pre-F27 macros.
 - Fix `build.sh` to not use `git -C` as CentOS doesn't support `git -C`.
 - Backport Travis configuration.

Note that `build.sh` doesn't pass `ShellCheck` and while we could include the `test_perl_style.sh` as the `v4.4.x` still uses the Perl build and test system, I've decided to omit the `stylecheck` CI pass.

@SilleBille Thoughts from the CI czar? :)

### DIFF
--- a/tools/run_container.sh
+++ b/tools/run_container.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+
+# This tool attempts to detect the presence of various tools to run the CI
+# images. If present, it'll use them to run the specified container image.
+
+function rc_buildah() {
+    buildah_path="$(command -v buildah)"
+    podman_path="$(command -v podman)"
+
+    [ "x$buildah_path" != "x" ] && [ "x$podman_path" != "x" ]
+}
+
+function rc_docker() {
+    docker_path="$(command -v docker)"
+
+    [ "x$docker_path" != "x" ]
+}
+
+function rc_run() {
+    local image="$1"
+    local ret=0
+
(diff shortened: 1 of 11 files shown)


tools/run_container.sh
```diff
@@ -0,0 +1,73 @@
...
+    local ret=0
+
+    if [ ! -f "tools/Dockerfiles/$image" ]; then
+        echo "Error: tools/Dockerfiles/$image is not a file; must be an" 1>&2
+        echo "existing location to launch container." 1>&2
+
+        exit 1
+    fi
+
+    if rc_buildah; then
```
SilleBille: Do we support both docker and buildah? I see only Dockerfiles along with this PR. Is buildah better than Docker?

edewata:
I agree with @SilleBille about using existing vanilla OS containers for testing if available. We should only use Dockerfiles if we want to create images that contain our product for other people to use.

cipherboy:
I want to keep the comments and the discussion, but there's no way for me to move them to `master` branch. 

However, I do want to use this to test another PR, so I'm merging this on the basis of it being a backport and CI passing now (thanks to the LDAPSDK fix).

We should discuss this on master. I'll open another PR there.

PR #115, opened 2019-01-23
Fix LDAPJDK circular dependency

cipherboy:
These changes were backported from master; jss no longer requires
ldapjdk so we can remove it from the spec file and the build scripts.


cipherboy:
As this is a minor spec file change, I'm going ahead and merging. I've validated that all ldapjdk APIs are removed from this version.

PR #119, opened 2019-01-29
Bump NSS version to 3.30 for SSLAlert

cipherboy:
Reported by @emaldona via email.

Minimum NSS version is 3.28 in the spec file, but SSLAlert is introduced in 3.30:

https://github.com/nss-dev/nss/blob/NSS_3_28_BRANCH/lib/ssl/ssl.h
https://github.com/nss-dev/nss/blob/NSS_3_29_BRANCH/lib/ssl/ssl.h
https://github.com/nss-dev/nss/blob/NSS_3_30_BRANCH/lib/ssl/ssl.h#L832


cipherboy:
We'll need to backport this to `v4.4.x` branch as well. It appears RHEL 6/7 ships with NSS 3.36 ([source](https://pkgs.org/download/nss)), so that should be fine.

PR #125, opened 2019-02-11
JNI improvements: ToByteArray

cipherboy:
~Includes a commit from #124. Will be rebased after that PR is merged.~

A small helper method with consistent implementation for creating a new `jByteArray` from an existing array-like item in JSS. This proved useful in #122, so I split it off here. 

Ultimately I'd like more small wrappers like this around JNI functionality to have more consistent error handling.

I'm not sure what we should do instead of `assert` in C when `env` is `NULL`, but I'm guessing we don't have many options. Wrapping error handling would be another series of good JNI improvements.

### DIFF
--- a/org/mozilla/jss/util/jssutil.c
+++ b/org/mozilla/jss/util/jssutil.c
@@ -317,20 +317,7 @@ JSS_getPtrFromProxyOwner(JNIEnv *env, jobject proxyOwner, char* proxyFieldName,
 jbyteArray
 JSS_ptrToByteArray(JNIEnv *env, void *ptr)
 {
-    jbyteArray byteArray;
-
-    /* Construct byte array from the pointer */
-    byteArray = (*env)->NewByteArray(env, sizeof(ptr));
-    if(byteArray==NULL) {
-        PR_ASSERT( (*env)->ExceptionOccurred(env) != NULL);
-        return NULL;
-    }
-    (*env)->SetByteArrayRegion(env, byteArray, 0, sizeof(ptr), (jbyte*)&ptr);
-    if((*env)->ExceptionOccurred(env) != NULL) {
-        PR_ASSERT(PR_FALSE);
-        return NULL;
-    }
-    return byteArray;
+    return JSS_ToByteArray(env, (void *)&ptr, sizeof(ptr));
 }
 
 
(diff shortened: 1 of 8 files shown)


org/mozilla/jss/util/jssutil.c
```diff
@@ -593,6 +580,32 @@ JSS_ByteArrayToSECItem(JNIEnv *env, jbyteArray byteArray)
...
 
+/************************************************************************
+** JSS_ToByteArray.
+**
+** Converts the given chararacter array to a Java byte array.
+**
+** Returns
+**  The new jbyteArray object or NULL on failure.
+*/
+jbyteArray JSS_ToByteArray(JNIEnv *env, const void *data, int length)
```
edewata: Since the data param is supposed to be a Java byte array, it might make more sense to declare it as const jbyte * data, but it's just nitpicking, the code is fine as is.

cipherboy: I think the argument I make is that we use `JSS_ToByteArray` to store a lot of non-`jbyte` contents from time to time. Making the input `void *` ensures we don't need to cast the input parameter. 

Some examples are given in the second commit (where we use `JSS_ToByteArray`), but this includes storing pointers for `*Proxy` classes.

(Java/JNI lacks a true way of storing pointers managed by C code on a Java object. The best you can do is store the pointer as a `byte[]` and coerce to-from pointers).

cipherboy: Trying out `jbyte *` instead of `void *` shows that you get a bunch of warnings:

```c
/home/cipherboy/GitHub/c2/jss/org/mozilla/jss/pkcs11/PK11Cipher.c: In function ‘Java_org_mozilla_jss_pkcs11_PK11Cipher_updateContext’:
/home/cipherboy/GitHub/c2/jss/org/mozilla/jss/pkcs11/PK11Cipher.c:166:37: warning: pointer targets in passing argument 2 of ‘JSS_ToByteArray’ differ in signedness [-Wpointer-sign]
     outArray = JSS_ToByteArray(env, outbuf, outlen);
                                     ^~~~~~
```

cipherboy:
Recognizing feedback, I've decided to push for reasons in the comments. Thanks!

PR #130, opened 2019-02-28
[WIP] JNI for NSS_Init - 1:1

cipherboy:
Currently based on top of #122, though that could be changed.

This adds JNI for the one NSS method that is called in #93: `NSS_Init`, to initialize a NSS DB.

I'm inclined to drop this PR.


cipherboy:
I'm inclined to drop this PR, as we should be using the CryptoManager class, which calls `NSS_Init` for us.

PR #133, opened 2019-03-01
Added Cert.sortCertificateChain()

edewata:
The sortCertificateChain() has been moved from
com.netscape.cmsutil.crypto.CryptoUtil in PKI to
org.mozilla.jss.netscape.security.util.Cert in
JSS for reusability.

The version number has been updated to 4.5.3 such
that the JSS dependency can be set properly in
other packages.

### DIFF
--- a/org/mozilla/jss/netscape/security/util/Cert.java
+++ b/org/mozilla/jss/netscape/security/util/Cert.java
@@ -21,14 +21,24 @@ import java.io.IOException;
 import java.security.cert.CertificateException;
 import java.security.cert.X509CRL;
 import java.security.cert.X509Certificate;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
 
+import org.apache.commons.lang.ArrayUtils;
 import org.mozilla.jss.crypto.SignatureAlgorithm;
-
 import org.mozilla.jss.netscape.security.pkcs.PKCS7;
 import org.mozilla.jss.netscape.security.x509.X509CRLImpl;
 import org.mozilla.jss.netscape.security.x509.X509CertImpl;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class Cert {
+
+    private static Logger logger = LoggerFactory.getLogger(Cert.class);
 
     public static final String HEADER = "-----BEGIN CERTIFICATE-----";
     public static final String FOOTER = "-----END CERTIFICATE-----";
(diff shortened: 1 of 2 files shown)


org/mozilla/jss/netscape/security/util/Cert.java
```diff
@@ -204,4 +214,130 @@ public static String normalizeCertStrAndReq(String s) {
...
+                String parent = parentMap.get(subjectDN);
+                if (parent != null) logger.debug("   parent: " + parent);
+
+                String child = childMap.get(subjectDN);
+                if (child != null) logger.debug("   child: " + child);
+            }
+        }
+
+        // find leaf cert
+        List<String> leafCerts = new ArrayList<>();
```
cipherboy: Isn't another way of expressing this (at the cost of having to change the type, which we don't particularly care about / can use a .toArray();` call:

```java
<set>  leafCertsKeys = certMap.keySet();
       leafCertsKeys.removeAll(childMap.keySet());
<list> leafCerts = leafCertsKeys.toArray(...);
```

This saves us the loop. Not sure if you want to make that change or not.

org/mozilla/jss/netscape/security/util/Cert.java
```diff
@@ -204,4 +214,130 @@ public static String normalizeCertStrAndReq(String s) {
...
+
+            // found leaf cert
+            leafCerts.add(subjectDN);
+        }
+
+        if (leafCerts.isEmpty()) {
+            throw new Exception("Unable to find leaf certificate");
+        }
+
+        if (leafCerts.size() > 1) {
```
cipherboy: @edewata One additional thought while I'm here, regarding future use of JSS for a `PKICertImport` replacement.

OpenSSL only creates `.p12` files with one chain and one key in them. Java and a few other providers lets you stick multiple chains/keys in one `.p12` file.

Do we want to extend this to support multiple chains, and leave chain quantity validation as a separate function? Or leave that to a later time?

cipherboy: I guess the behavior is non-trivial:

Do we want a list of complete chains? Or just a topological sort of the graph? Which we want probably depends on usage... But guessing topological sort of the graph would be best.

org/mozilla/jss/netscape/security/util/Cert.java
edewata: OK, I think that'll work.

org/mozilla/jss/netscape/security/util/Cert.java
edewata: Our tools can export the whole NSS database (which may contain multiple chains/keys) into a PKCS #12 file, which can be used to set up a clone or for migration. If we need to create a topological graph we can certainly refactor the code, but we still need to keep this method to sort a single chain from PKCS #7 data. It should be easy to do that separately later. This patch is mainly for moving the code, so I'd like to minimize code changes in this patch.

cipherboy: Fine by me. Thanks!

org/mozilla/jss/netscape/security/util/Cert.java
edewata: Please take a look at the new patch. It's slightly different since we should not modify the 
certMap.keySet() result directly.

edewata:
I'll add the unit test in a separate patch (need to generate some test certs).

org/mozilla/jss/netscape/security/util/Cert.java
cipherboy: Cool, :+1: from me.

PR #134, opened 2019-03-01
JNI JString improvements

cipherboy:
This simplifies the interaction between C and Java with regards to JNI's `jstring` representation of Java's `String` objects. 

In particular, we only ever access JString's via their UTF8 representations, so this wrapper assumes we're doing that. It also handles `NULL` strings, simplifying the logic to call these wrappers: always call them and condition around a `NULL` return value instead of a `NULL` input value. 

This also caught two types of bugs:
 - Where we forget to dereference at all.
 - Where we (incorrectly) conditionally dereference around whether or not we're acting on a copy. Common consensus seems to be to always dereference, even on copies.

Questions: 
 - Should we add error handling logic when the string fails to parse/reference? E.g., `jstring != NULL` but the return value is `NULL`? This influences whether or not the changes to `configureOCSP` are correct.
 - Should we make `Ref` return a `const char *` instead of a `char *`? This means that some references will have to cast to `(char *)` themselves. Perhaps we can update those places to use `const char *`...

### DIFF
--- a/org/mozilla/jss/CryptoManager.c
+++ b/org/mozilla/jss/CryptoManager.c
@@ -109,8 +109,8 @@ int ConfigureOCSP(
         jstring ocspResponderURL,
         jstring ocspResponderCertNickname )
 {
-    char *ocspResponderURL_string=NULL;
-    char *ocspResponderCertNickname_string=NULL;
+    const char *ocspResponderURL_string = NULL;
+    const char *ocspResponderCertNickname_string = NULL;
     SECStatus status;
     int result = SECSuccess;
     CERTCertDBHandle *certdb = CERT_GetDefaultCertDB();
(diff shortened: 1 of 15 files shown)


org/mozilla/jss/CryptoManager.c
```diff
@@ -120,27 +120,8 @@ int ConfigureOCSP(
...
-        ocspResponderCertNickname_string =
-            (char*) (*env)->GetStringUTFChars(env, ocspResponderCertNickname, NULL);
-        if (ocspResponderCertNickname_string == NULL) {
-            JSS_throwMsg(env, GENERAL_SECURITY_EXCEPTION,
-                    "OCSP invalid nickname");
-            result = SECFailure;
-            goto finish;
-        }
-    }
+    ocspResponderURL_string = JSS_RefJString(env, ocspResponderURL);
```
cipherboy: @edewata -- The behavior changes here because we no longer throw `"OCSP invalid URL"` in the event of decode failure. (Notice the loop change on new line 133 as well). 

This ties into question 1.

edewata: The change in line 133 might actually hide the memory issue. If ocspResponderURL is not NULL but if ocspResponderURL_string is NULL, the code will be skipped as if the ocspResponderURL was NULL, so the behaviour might be wrong. This is probably a corner case and the code might fail in other places as well due to memory issue, but I'd suggest reverting to the original code for line 133.

cipherboy: I've added error checking as part of referencing and updated the conditional on new line 172 to preserve original behavior. (OCSP responder is neither enabled nor disabled in the event of an error).

cipherboy:
I'm going to run an integration check in my test environments to smoke test these changes some more. Mostly want to make sure I didn't introduce an error into any obvious places.

PR #135, opened 2019-03-01
Remove needless dependency on hamcrest

cipherboy:
jUnit pulls in `hamcrest-core` for us, which is all we need. This should fix Rawhide builds, since `hamcrest` fails to install.


cipherboy:
It works! :D

PR #138, opened 2019-03-11
JNI Enhancements: JSS_FromByteArray

cipherboy:
Similar to #134, this changes how byte arrays are handled and using them in several places.

We already have a `JSS_ToByteArray` that we're using extensively, but `JSS_FromByteArray` does the opposite (`jbyteArray -> uint8_t *`). It also adds underlying `JSS_RefByteArray` and `JSS_DerefByteArray` helpers to access the raw byte array elements. 

Unlike `JSS_RefByteArray`, `JSS_FromByteArray` always makes a copy, so rather than returning it with `JSS_DerefByteArray`, free it with `free(...)`.

~Currently WIP because I need to:~
 - [x] Look at `JSS_RefByteArray` usages
 - [x] Decide whether or not we should care about returning the copy value.

### DIFF
--- a/org/mozilla/jss/util/jssutil.c
+++ b/org/mozilla/jss/util/jssutil.c
@@ -617,8 +617,8 @@ jbyteArray JSS_ToByteArray(JNIEnv *env, const void *data, int length)
 ** JSS_DerefByteArray.
 **
 ** Returns
-**  bool - whether or not the operation succeeded. The operation succeeds if
-**  array is NULL, it has zero length, or if *data is successfully referenced.
+**  bool - whether or not the operation succeeded. The operation succeeds
+**  if *data is successfully referenced (i.e., is non-null).
 */
 bool JSS_RefByteArray(JNIEnv *env, jbyteArray array, jbyte **data,
     jsize *length)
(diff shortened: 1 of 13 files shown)


cipherboy:
@edewata I tested this via my PKI integration tests and it doesn't appear to break PKI install, so I think we're good.

org/mozilla/jss/util/jssutil.c
```diff
@@ -609,6 +607,107 @@ jbyteArray JSS_ToByteArray(JNIEnv *env, const void *data, int length)
...
+    *data = NULL;
+
+    if (!JSS_RefByteArray(env, array, &array_data, &array_length)) {
+        return false;
+    }
+
+    /* Defensive coding: Java's byte arrays are not null terminated, allocate
+     * a structure one larger to guarantee C functions work as expected. */
+    *data = calloc(array_length + 1, sizeof(uint8_t));
+    memcpy(data, array_data, array_length);
```
cipherboy: This is a bug: Should be `memcpy(*data, ...)`.

cipherboy:
I think this should be fine. It is getting tested and exercised as part of #147.

PR #141, opened 2019-03-15
Fix release procedure in jss.spec

cipherboy:
As discovered with the v4.4.6 release, the tarball creation process as
described in the spec file was wrong and/or generated a different
tarball than linked to by the Source URL. Update the procedure to match
the URL.


cipherboy:
(As a side note, since `master` ships with the new CMake-based build system, we have no use for `perl-interpreter`, so we can drop it as a dependency)

PR #142, opened 2019-03-18
Use memset for wipeCharArray, add length variant

cipherboy:
When the passed data is not null-terminated, it is useful to have a
length-terminated variant.

### DIFF
--- a/org/mozilla/jss/util/jssutil.c
+++ b/org/mozilla/jss/util/jssutil.c
@@ -7,6 +7,8 @@
 #include <plstr.h>
 #include <seccomon.h>
 #include <secitem.h>
+#include <string.h>
+
 #include "jssutil.h"
 #include "jss_bigint.h"
 #include "jss_exceptions.h"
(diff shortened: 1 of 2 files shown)


cipherboy:
I've not yet decided if I want to merge this. If I take #138 to its logical conclusion, we'd have length-terminated data for all of the passwords we'd need to wipe. While they should already be null-terminated, they're passed in from Java (which doesn't guarantee null-termination), so this way would be a little safer.

cipherboy:
While we currently guarantee (through much care) that the values passed over the JNI layer into C are null-terminated UTF8, the null-termination is done through an extra UTF8 conversion. This is important since the `GetStringUTFChars()` doesn't _actually_ return UTF-8 -- it returns [Modified UTF-8](https://docs.oracle.com/javase/8/docs/api/java/io/DataInput.html#modified-utf-8).

cipherboy:
I'm leaving this as-is. Neither this code nor the existing code is [guaranteed](http://www.daemonology.net/blog/2014-09-04-how-to-zero-a-buffer.html) to [work](http://www.daemonology.net/blog/2014-09-05-erratum.html).

PR #103, opened 2018-12-04
Common Buffer Implementation - 0:0

cipherboy:
This is an implementation of a ring buffer for use by internal JSS code. In particular, it will be used for the Buffer PRFD implementation (#93) for NIO work. 

Test cases were generated by [`test_gen.py`](https://github.com/cipherboy/testbed/blob/master/ringbuf/test_gen.py) from [`cipherboy/testbed`](https://github.com/cipherboy/testbed/tree/master/ringbuf). They exhaustively test all pairwise combinations of put/write and get/read, with all counts/sizes of reads from 0 to `<size>+1` (e.g., exceeding the size of the buffer).

Marking WIP as 4.5.2 isn't released and this is slated for 4.5.3.

PR #145, opened 2019-03-20
JNI for j_buffer - 2:0

cipherboy:
Dependencies:

 - ~#138 - for JNI enhancements.~
 - ~#103 - for the buffer implementation.~
 - ~#93 - for buffer-prfd and associated `j_buffer` improvements.~
 - ~#122 - for accessing NSPR's `PRFileDesc` objects in Java.~

(Note that #93 and #122 aren't strictly necessary for this PR, but are for the next PR in the series).

This PR adds support for calling into `j_buffer` objects from Java. The next PR in this series (`2:1`) will let us construct `BufferPRFD` objects in Java.

Then `3:0` will combine `2:1` with `1:1` and let us reproduce the [test case](https://github.com/cipherboy/jss/blob/buffer-prfd/org/mozilla/jss/tests/TestBufferPRFD.c) for a Buffer PRFileDesc (i.e., do a single-threaded async SSL handshake) in Java, and `4:0` will begin implementing a `SSLEngine`.

### DIFF
--- a/org/mozilla/jss/nss/Buffer.java
+++ b/org/mozilla/jss/nss/Buffer.java
@@ -1,17 +1,66 @@
 package org.mozilla.jss.nss;
 
 public class Buffer {
+    /**
+     * Create a new j_buffer object with the specified number of bytes.
+     *
+     * See also: jb_alloc in org/mozilla/jss/ssl/javax/buffer.h
+     */
     public static native BufferProxy Create(long length);
 
+    /**
+     * Check the total capacity of a buffer object.
+     *
+     * See also: jb_capacity in org/mozilla/jss/ssl/javax/buffer.h
+     */
     public static native long Capacity(BufferProxy buf);
+
+    /**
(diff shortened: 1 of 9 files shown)


org/mozilla/jss/nss/Buffer.java
```diff
@@ -0,0 +1,17 @@
+package org.mozilla.jss.nss;
+
+public class Buffer {
```
cipherboy: Add documentation here.

org/mozilla/jss/nss/BufferProxy.h
```diff
@@ -0,0 +1,7 @@
+#include <jni.h>
+#include "buffer.h"
+
+#pragma once
+
+jobject JSS_PR_wrapJBuffer(JNIEnv *env, j_buffer **buffer);
```
cipherboy: Documentation here as well.

PR #148, opened 2019-03-27
Remove 3s sleep from SetupDBs

cipherboy:
As far as I can tell, this 3s sleep is extraneous. This test creates the
NSS DB (instead of calling certutil -N) and provides an anchor point in
the test suite. Our alternatives are to use `certutil -N` instead of
SetupDBs, but we might as well keep this, just without the sleep.

It is possible the sleep was used, perhaps to ensure that the DB is
flushed to disk before the test suite continues, but this should be
handled by the JVM and shouldn't be considered as a side-effect of a
Sleep.

### DIFF
--- a/org/mozilla/jss/tests/SetupDBs.java
+++ b/org/mozilla/jss/tests/SetupDBs.java
@@ -36,8 +36,6 @@ public class SetupDBs {
             new FilePasswordCallback( args[1] )
         );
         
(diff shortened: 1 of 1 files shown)


cipherboy:
@edewata Do you happen to remember why you updated the call vs. removing the sleep? The tests seem to pass fine without the sleep...

PR #149, opened 2019-03-28

cipherboy:
This lets us see which commit a user is at when using our various auto-built COPRs. This'd let us triage bugs better to know if something recent broke, or if a bug was fixed recently, etc. Thoughts?

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -1,5 +1,5 @@
 srpm:
-	./build.sh srpm
+	./build.sh srpm --with-timestamp --with-commit-id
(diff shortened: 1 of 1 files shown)


cipherboy:
@SilleBille I was for sure going to do it for all JSS builds, I think it'd make sense to apply it across all automated branches and dependencies, yes. @edewata thoughts?

edewata:
Yeah, as discussed on IRC, we should include the timestamp/commit ID anywhere we implement autobuild to avoid duplicate NVRs.

PR #151, opened 2019-03-29

cipherboy:
Backport of #149. 

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -1,4 +1,5 @@
 srpm:
+	dnf install -y git
 	./build.sh srpm --with-timestamp --with-commit-id
 	if [[ "${outdir}" != "" ]]; then \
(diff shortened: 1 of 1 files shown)


.copr/Makefile
```diff
@@ -1,5 +1,5 @@
 srpm:
-	./build.sh srpm
+	./build.sh srpm --with-timestamp --with-commit-id
```
SilleBille: please change it as:

`./build.sh --with-timestamp --with-commit-id srpm`

Just noticed the issue. :man_facepalming:

cipherboy: I wonder if it'd be possible to remove the dependency on `git`... I already had to patch `v4.4.x`'s `build.sh` to actually support building on RHEL 7 (since `git` there doesn't understand `git -C`...).

cipherboy:
@SilleBille's comments have been addressed, merging.

PR #152, opened 2019-03-29

cipherboy:
Backport of #149. 


.copr/Makefile
```diff
@@ -1,5 +1,5 @@
 srpm:
-	./build.sh srpm
+	./build.sh srpm --with-timestamp --with-commit-id
```
SilleBille: please change it as:

`./build.sh --with-timestamp --with-commit-id srpm`

cipherboy:
@SilleBille's comments have been addressed, merging.

PR #153, opened 2019-03-29
Support in-tree, gitless builds

cipherboy:
When the builder lacks the git executable but has a full git tree
checked out, --with-commit-id (and the resulting build) will fail.
However, there's no reason to as we have all the information necessary
to propagate the commit id correctly.


SilleBille:
I have few concerns with this patch. I'll try to clarify over irc before this patch gets merged.

PR #155, opened 2019-03-29
Return positive BigIntegers

cipherboy:
Several of the usages of `BigInteger`s lead to the potential of negative
numbers, though only positive numbers should be returned.

`PK11RSAPublicKey`'s `getModulus()` and `getPublicExponent()` methods can
potentially return negative values, when both the modulus and exponent
should be strictly positive.

`PK11PrivKey`'s `getDSAParams()` could return negative values depending on
the PQG parameters.

The Key Identifiers printed by `tests.CloseDBs` should be positive (as
they're UIDs) but could be displayed as negative values; the same
happens in `crmf.CertReqMsg`.

This negation property is discussed [here](https://docs.oracle.com/javase/7/docs/api/java/math/BigInteger.html#BigInteger(int,%20byte[])).

### DIFF
--- a/org/mozilla/jss/pkcs11/PK11PrivKey.java
+++ b/org/mozilla/jss/pkcs11/PK11PrivKey.java
@@ -111,9 +111,9 @@ public class PK11PrivKey extends org.mozilla.jss.pkcs11.PK11Key
         byte[][] pqgArray = getDSAParamsNative();
 
         return new DSAParameterSpec(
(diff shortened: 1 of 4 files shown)


cipherboy:
@edewata / @ladycfu / @jmagne: I've tested this as part of my integration tests and it doesn't seem to break install. Should I go ahead and merge?

PR #157, opened 2019-04-05
Extend SSLVersion to support JDK names

cipherboy:
`SSLVersion` currently only supports a single alias for naming TLS
protocol versions. Keep providing (and defaulting to) the NSS style
names, but also provide the JDK style names. This allows us to map
between the NSS constant and the JDK names.

@edewata Are there other changes you'd make to `SSLVersion` while we're here? 

### DIFF
--- a/org/mozilla/jss/ssl/SSLVersion.java
+++ b/org/mozilla/jss/ssl/SSLVersion.java
@@ -6,22 +6,39 @@ package org.mozilla.jss.ssl;
 
 public enum SSLVersion {
 
-    SSL_3_0 ("SSL3",   SocketBase.SSL_LIBRARY_VERSION_3_0),
-    TLS_1_0 ("TLS1_0", SocketBase.SSL_LIBRARY_VERSION_TLS_1_0),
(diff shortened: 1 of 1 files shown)


org/mozilla/jss/ssl/SSLVersion.java
```diff
@@ -36,14 +53,24 @@ public static SSLVersion valueOf(int value) {
         throw new IllegalArgumentException("Invalid SSLVersion value: " + value);
     }
 
+    public boolean matchesAlias(String value) {
+        for (String alias : aliases) {
+            if (value.toUpperCase() == alias.toUpperCase()) {
```
edewata: I think this should be:
```
if (alias.equalsIgnoreCase(value)) {
```
Notes:
* String comparison in Java should be done with equals...() instead of ==.
* The equals...() is called against alias since it's guaranteed not to be null.

cipherboy: Ah duh, sorry. I've been not-Java-ing for too long.

PR #160, opened 2019-04-09
Update provider to expose PK11Certs - 5:2

cipherboy:
~Depends on #158, will be rebased once that merges.~

This updates our [`KeyStoreSpi`](https://docs.oracle.com/javase/8/docs/api/java/security/KeyStoreSpi.html) implementation, [`JSSKeyStoreSpi`](https://dogtagpki.github.io/jss/javadoc/org/mozilla/jss/provider/java/security/JSSKeyStoreSpi.html), to return [`PK11Cert`](https://dogtagpki.github.io/jss/javadoc/org/mozilla/jss/pkcs11/PK11Cert.html)s instead of encoding and decoding into a [`java.security.cert.X509Certificate`](https://docs.oracle.com/javase/8/docs/api/java/security/cert/X509Certificate.html). 

### DIFF
--- a/org/mozilla/jss/provider/java/security/JSSKeyStoreSpi.java
+++ b/org/mozilla/jss/provider/java/security/JSSKeyStoreSpi.java
@@ -41,6 +41,7 @@ import org.mozilla.jss.crypto.TokenSupplierManager;
 import org.mozilla.jss.crypto.X509Certificate;
 import org.mozilla.jss.netscape.security.util.Utils;
 import org.mozilla.jss.pkcs11.PK11Token;
+import org.mozilla.jss.pkcs11.PK11Cert;
 import org.mozilla.jss.pkcs11.TokenProxy;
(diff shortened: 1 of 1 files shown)


org/mozilla/jss/provider/java/security/JSSKeyStoreSpi.java
```diff
@@ -369,6 +375,10 @@ public String engineGetCertificateAlias(Certificate cert) {
                 X509Certificate cert = certs[i];
                 logger.debug("JSSKeyStoreSpi: - " + cert.getSubjectDN());
 
+                if (cert instanceof PK11Cert) {
+                    chain[i] = (PK11Cert) cert;
+                }
```
edewata: Do we need a `continue` here?

cipherboy: Ah, great catch.

PR #162, opened 2019-04-09
Add GitHub Pages generation script

cipherboy:
Adds new script to automatically generate GitHub Pages content from the
latest master.

I'd take feedback about integrating this into CI automation. Thoughts @SilleBille? 

### DIFF
--- a/tools/regen_gh_pages.sh
+++ b/tools/regen_gh_pages.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+
+function _e() {
+    if [ ! -z "$VERBOSE" ]; then
+        echo cd "$(pwd)" '&&' "$@" 1>&2
+    fi
+
+    local command="$1"
+    shift
+
+    "$command" "$@"
+}
+
+function main() {(
+    set -e
+
+    local user="$1"
+    local remote="https://github.com/$user/jss"
+    local upstream="https://github.com/dogtagpki/jss"
+    local sandbox="/tmp/jss-docs-sandbox"
+
+    if [ -z "$user" ]; then
+        echo "Usage: update-gh-pages.sh <username>"
+        echo "Update the user's gh-pages branch with the latest javadocs content."
+        return 1
+    fi
+
+    _e rm -rf "$sandbox"
+    _e mkdir -p "$sandbox"
+    pushd "$sandbox"
+        _e git clone "$remote" jss
+        cd jss/build
+        _e git remote add upstream "$upstream"
+        _e git fetch --all
+        _e git checkout "upstream/master"
+
+        # Build Javadoc only
+        _e cmake ..
+        _e make javadoc
+
+        # Preserve javadoc
+        _e cp -rv docs "$sandbox/docs"
+
+        # Update javadocs
+        cd ../
+        _e git clean -xdf
+        _e git checkout "upstream/gh-pages"
+        _e git branch -D "gh-pages" || true
+        _e git checkout -b "gh-pages"
+        _e rm -rf javadoc
+        _e cp -rv "$sandbox/docs" javadoc
+        _e git add --all
+        _e git commit -m "Update javadocs from master at $(date '+%Y-%m-%d %H:%M')"
+        _e git push --set-upstream origin gh-pages --force
+    popd # "$sandbox"
+
+    _e rm -rf "$sandbox"
+
+    echo ""
+    echo ""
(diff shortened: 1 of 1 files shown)


SilleBille:
@cipherboy That's a good initiative! :+1:  Can you add the params that the script expects or can process? :)

When you say you want to integrate this into CI, how are you planning to handle different versions of JSS? I see that you use a `gh-pages` branch to publish contents to GH pages. You need to handle directories yourself. Take a look at this [sample repo](https://github.com/koumoul-dev/gh-pages-multi) :)

cipherboy:
@SilleBille For the time being, we're only supporting master (and optionally `v4.5.x`) branch builds as they're the easiest due to the rewrite to cmake. That lets us prep and then build only the `javadoc` target without a full package build. The older `v4.4.x` branch uses the Perl build system; we're relegated to building on systems we know well (to only build the javadoc), or building NSS/NSPR each time (prior to building the javadoc) if they're not installed on the system. 

I'd propose the following layout if we wanted to support multiple branches. 

```
gh-pages/master/docs
gh-pages/master/javadocs
gh-pages/v4.5.x/docs
gh-pages/v4.5.x/javadocs
gh-pages/v4.4.x/javadocs
```

We could rebuild all branches content on commit to any branch. The one parameter the current script expects is the name of the user to push as. Ideally this would be @pki-jenkins-bot.

cipherboy:
@edewata / @SilleBille I've updated this PR to generate the javadocs from all branches and created a PR (#165) to show the change set (large, since this is the first one with the new structure). Thoughts?

tools/regen_gh_pages.sh
```diff
@@ -0,0 +1,132 @@
...
+        clean_checkout "upstream/master"
+        build_javadoc "$artifacts/javadocs-master"
+        build_doc "$artifacts/docs-master"
+
+        # Build docs for the v4.5.x branch
+        clean_checkout "upstream/v4.5.x"
+        build_javadoc "$artifacts/javadocs-v4.5.x"
+        build_doc "$artifacts/docs-v4.5.x"
+
+        # Build docs for the v4.4.x branch
```
SilleBille: This patch is against `master` branch. Should we build for all branches?

Or my other concern is if I push a change to v4.4.x branch, will the CI run this script from master?

tools/regen_gh_pages.sh
```diff
@@ -0,0 +1,132 @@
...
+        _e git add --all
+        _e git commit -m "Update javadocs from master at $(date '+%Y-%m-%d %H:%M')"
+        _e git push --set-upstream origin gh-pages --force
+    popd # "$sandbox"
+
+    _e rm -rf "$sandbox"
+
+    echo ""
+    echo ""
+    echo "All done! To open a PR, click the following link:"
```
SilleBille: Once the script is executed successfully on the CI, i see that it'll force push it to **it's** forked version. :+1: 

Does it create a PR automatically or should we click on the link? If you think you can automate this step too, you can take a look at [github hub](https://hub.github.com/)

tools/regen_gh_pages.sh
cipherboy: I'm mostly placing this here as a placeholder for now. Currently:

- This script clones a git repo in a completely separate location
- This script builds all branches and updates accordingly

Because of that, master is the best place to put it, and it'd only be run from master. Logic is maintained in one place and that's it.

Ideally, this script would:

- Live in the otherwise-discussed automation pipeline and not in JSS.
- And yes, it'd support building just a single branch instead of all branches (actually, a fairly easy fix). 
- Use an existing git tree and only fetch the latest changes instead of everything.
- Support in-tree builds of JSS v4.4.x branch instead of pulling them from the javadocs RPM.

tools/regen_gh_pages.sh
cipherboy: Currently the PR is manual, especially since I like to review it and as you can see in #165, add additional commits. The side effect is that anything which triggers a build triggers an automatic nuke of the branch; perhaps this is fine, but we could fix this by building into a branch-specific GH pages branch.

And yes, if we place this (or a version of it) in the automation pipeline, I'd definitely go ahead and open the PR automatically. I think we can do it with `curl` via the [GitHub API](https://developer.github.com/v3/pulls/#create-a-pull-request), instead of requiring that `hub` be installed on all CI systems.

cipherboy:
This has been added internally and is currently running.

PR #164, opened 2019-04-16
Add tests for JSSProvider - 5:3

cipherboy:
Depends on #159 and #160. Will be rebased when they merge. Only the [last commit](https://github.com/dogtagpki/jss/pull/164/commits/7b9c64397b9041ee21bea8aeb1a854b53230bde2) in this series is useful.

Adds tests for the [`JSSProvider`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/JSSProvider.java) to ensure that it exposes stuff useful for our consumption internally (subclasses in `org.mozilla.jss.provider`). Also ensures that the provider we insert (via the `CryptoManager`) gets utilized by default.

### DIFF
--- a/cmake/JSSTests.cmake
+++ b/cmake/JSSTests.cmake
@@ -255,6 +255,11 @@ macro(jss_tests)
         COMMAND "org.mozilla.jss.tests.JCASymKeyGen" "${RESULTS_NSSDB_OUTPUT_DIR}"
         DEPENDS "Setup_DBs"
     )
+    jss_test_java(
+        NAME "JSSProvider"
+        COMMAND "org.mozilla.jss.tests.JSSProvider" "${RESULTS_NSSDB_OUTPUT_DIR}" "${PASSWORD_FILE}"
+        DEPENDS "List_CA_certs"
+    )
 
     # FIPS-related tests
     jss_test_java(
--- a/org/mozilla/jss/tests/JSSProvider.java
+++ b/org/mozilla/jss/tests/JSSProvider.java
@@ -1,0 +1,89 @@
+package org.mozilla.jss.tests;
+
+import java.util.*;
+import java.security.*;
+import javax.crypto.*;
+import javax.net.ssl.*;
+
+import org.mozilla.jss.*;
+import org.mozilla.jss.pkcs11.*;
+
+public class JSSProvider {
+    public static String BASE_PACKAGE = "org.mozilla.jss.provider.";
+
+    public static void shouldProvide(Provider p, String algo, String _type) {
+        String item = (String) p.get(algo);
+        assert(item != null);
+        assert(item.startsWith(BASE_PACKAGE));
+        assert(item.endsWith("." + _type));
+    }
+
+    public static void shouldNotProvide(Provider p, String algo) {
+        assert(p.get(algo) == null);
+    }
+
+    public static void main(String[] args) throws Exception {
+        // Before we initialize the CryptoManager, the JSS Provider shouldn't
+        // exist.
+        assert(Security.getProvider("Mozilla-JSS") == null);
+
+        CryptoManager.initialize(args[0]);
+        CryptoManager cm = CryptoManager.getInstance();
+        cm.setPasswordCallback(new FilePasswordCallback(args[1]));
+
+        // Validate that the CryptoManager registers us as the
+        // default/first provider.
+        Provider p = Security.getProviders()[0];
+        assert(p.getName().equals("Mozilla-JSS"));
+        assert(p instanceof org.mozilla.jss.JSSProvider);
+
+        // Validate assumptions about how provider.get() works:
+        //
+        // Something which by no means should ever be implemented returns
+        // null.
+        shouldNotProvide(p, "Charlie's Chocolate Factory");
+
+        // Validate that our provider implements certain interfaces we care
+        // about. In particular, ensure they're overriden as part of our
+        // org.mozilla.jss.provider package, and that the class name matches
+        // what we expect our implementation to be.
+        shouldProvide(p, "KeyManagerFactory.NssX509", "JSSKeyManagerFactory");
+
+        // Validate that our provider is "default"; that is, when we get an
+        // algorithm instance without explicitly providing the provider, we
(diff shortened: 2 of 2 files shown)


cmake/JSSTests.cmake
```diff
@@ -110,6 +110,7 @@ macro(jss_tests)
         NAME "JUnit_ChainSortingTest"
         COMMAND "org.junit.runner.JUnitCore" "org.mozilla.jss.tests.ChainSortingTest"
     )
+
```
cipherboy: Cleaner commit, drop this.

org/mozilla/jss/tests/JSSProvider.java
```diff
@@ -0,0 +1,85 @@
...
+
+        // Validate that our provider implements certain interfaces we care
+        // about. In particular, ensure they're overriden as part of our
+        // org.mozilla.jss.provider package, and that the class name matches
+        // what we expect our implementation to be.
+        shouldProvide(p, "KeyManagerFactory.SunX509", "JSSKeyManagerFactory");
+
+        // Validate that our provider is "default"; that is, when we get an
+        // algorithm instance without explicitly providing the provider, we
+        // end up with an instance from the provider.
```
cipherboy: `the` -> `our`.

org/mozilla/jss/tests/JSSProvider.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        TrustManagerFactory tmf = TrustManagerFactory.getInstance("SunX509");
+        assert(tmf.getTrustManagers().length > 0);
+        for (TrustManager tm : tmf.getTrustManagers()) {
+            System.out.println("got TrustManager: " + tm);
+            assert(tm instanceof org.mozilla.jss.provider.javax.crypto.JSSTrustManager);
+        }
+
+        // Our KeyStore should return instances of PK11Cert.
```
cipherboy: Instances, or extensions of `PK11Cert`.

org/mozilla/jss/tests/JSSProvider.java
```diff
@@ -0,0 +1,85 @@
...
+        KeyStore ks = KeyStore.getInstance("PKCS11");
+        ks.load(null, null);
+        Enumeration<String> aliases = ks.aliases();
+        while (aliases.hasMoreElements()) {
+            String alias = aliases.nextElement();
+            java.security.cert.Certificate cert = ks.getCertificate(alias);
+
+            if (cert == null || cert.getClass() == null) {
+                break;
```
cipherboy: `break` -> `continue`.

org/mozilla/jss/tests/JSSProvider.java
```diff
@@ -0,0 +1,85 @@
...
+
+        // Our KeyStore should return instances or extensions of PK11Cert.
+        KeyStore ks = KeyStore.getInstance("PKCS11");
+        ks.load(null, null);
+        Enumeration<String> aliases = ks.aliases();
+        while (aliases.hasMoreElements()) {
+            String alias = aliases.nextElement();
+            java.security.cert.Certificate cert = ks.getCertificate(alias);
+
+            if (cert == null || cert.getClass() == null) {
```
edewata: Not sure if we need to check `cert.getClass()`. Will it ever be null?

org/mozilla/jss/tests/JSSProvider.java
```diff
@@ -0,0 +1,85 @@
...
+                continue;
+            }
+
+            assert(org.mozilla.jss.pkcs11.PK11Cert.class.isAssignableFrom(cert.getClass()));
```
edewata: Would this work?
```
assert(cert instanceof org.mozilla.jss.pkcs11.PKCS11Cert);
```

org/mozilla/jss/tests/JSSProvider.java
```diff
@@ -0,0 +1,85 @@
...
+        // about. In particular, ensure they're overriden as part of our
+        // org.mozilla.jss.provider package, and that the class name matches
+        // what we expect our implementation to be.
+        shouldProvide(p, "KeyManagerFactory.NssX509", "JSSKeyManagerFactory");
+
+        // Validate that our provider is "default"; that is, when we get an
+        // algorithm instance without explicitly providing the provider, we
+        // end up with an instance from our provider.
+        Mac m = Mac.getInstance("HmacSHA512");
+        assert(m.getProvider().getName() == p.getName());
```
edewata: This only works if they are pointing to the same String object, right? Is that the assumption that we want to test here, or should we do something like:
```
assert(m.getProvider().getName().equals(p.getName()));
```

org/mozilla/jss/tests/JSSProvider.java
cipherboy: No, because there's derivative classes/interfaces and such that make this messy and include things that aren't an instance of `PKCS11Cert`. (If I recall correctly, this has to do with the JSS X509Certificate interface). 

What this is validating is that we can make a statement such as `PK11Cert something = cert`.

org/mozilla/jss/tests/JSSProvider.java
cipherboy: D'uh, sorry. This still gets me.

org/mozilla/jss/tests/JSSProvider.java
```diff
@@ -0,0 +1,85 @@
...
+        assert(item.endsWith("." + _type));
+    }
+
+    public static void shouldNotProvide(Provider p, String algo) {
+        assert(p.get(algo) == null);
+    }
+
+    public static void main(String[] args) throws Exception {
+        Provider p = Security.getProvider("Mozilla-JSS");
+        System.out.println(p);
```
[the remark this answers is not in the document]
cipherboy: I think this is safe to add; the syso was a debug statement.

org/mozilla/jss/tests/JSSProvider.java
cipherboy: On second thought, I think either is fine, I'll rewrite this to use `instanceof`.

PR #165, opened 2019-04-24
Update javadocs from master at 2019-04-23 15:12

cipherboy:
This is in concert with #162. We add the following to our GitHub pages:

- `master/docs`
- `master/javadocs`
- `v4.5.x/docs`
- `v4.5.x/javadocs`
- `v4.4.x/javadocs`

Some points of discussion: 

 - Should `v4.6.x` be a symbolic link to master right now, such that the version link will always work?
 - Should we try and make the `v4.5.x`/`master` and `v4.4.x` look the same?
 - Where should we maintain an `index.md` for the `docs/` folders? In each branch or in the `gh-pages` location?

This PR can be previewed here: https://cipherboy.github.io/jss


edewata:
> Should v4.6.x be a symbolic link to master right now, such that the version link will always work?

Either way should be fine, but "master" implies that it's still under development whereas "4.6" might imply a released product (which is not the case).

> Should we try and make the v4.5.x/master and v4.4.x look the same?

You mean the missing "docs" for v4.4.x?

> Where should we maintain an index.md for the docs/ folders? In each branch or in the gh-pages location?

I think the /docs in each branch should have an index.md so people browsing the repo directly can locate the docs more easily (instead of reading the filenames).

Overall this looks good. The above concerns can be addressed later.

cipherboy:
@edewata said:
> You mean the missing "docs" for v4.4.x?

No, there are a few javadoc generation differences between `v4.4.x` and `v4.5.x` (mostly to do with frames). I think we could actually pull out the javadoc generation logic and use the same logic on both branches. It requires us to maintain a separate javadoc generation target on `v4.4.x` though.

edewata:
I don't think there's any specific requirement about javadoc, so if it takes too much work, don't worry about that.

cipherboy:
OK. I'm going ahead and merging this. `index.md` can be added at a later date and then we can update this with recent API changes.

PR #166, opened 2019-04-24
CC: Enable AIA OCSP cert checking for entire certificate chain

cipherboy:
This is a forward port of #121 and #126. For more information, see those corresponding PRs.

@jmagne, what's the testing plan for this look like?

### DIFF
--- a/org/mozilla/jss/CryptoManager.java
+++ b/org/mozilla/jss/CryptoManager.java
@@ -1201,6 +1201,44 @@ public final class CryptoManager implements TokenSupplier
     // OCSP management
     ///////////////////////////////////////////////////////////////////////
 
+    /* OCSP Policy related */
+
+    public enum OCSPPolicy {
+        NONE,
+        NORMAL,
+        LEAF_AND_CHAIN;
+    }
+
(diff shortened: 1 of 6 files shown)


cipherboy:
@jmagne I'm wondering if we can add an upstream test case here somehow. Thoughts?

How hard would it be to generate a few certificates that use a custom URL for OCSP checks and just validate that the right certificates are checked?

jmagne:
The code looks largely the same if not completely. I will have to think today about how to test this thing. I assume we want a jss test program. I think we need of course a simple client and server , with the client making an ssl request to the server. As for the certs maybe we can just pre-create a cert chain and import it into the test's nss db. Then have the client make an ssl request to the server. But then we would need some sort of dummy ocsp responder process that just takes the requests and logs them maybe and return a successful result. Something like this. but that is a lot of work ... but if we did it, we could drop in a bunch of different chains and server certs to cover various positive and negative test cases.

cipherboy:
OK, in conversation with @SilleBille, I think I'll go ahead and merge this. It'll take more time to implement the tests as discussed with @jmagne. This is technically a regression from v4.4.x, so we'll want this in a v4.5.3 and a v4.6.0 release...

PR #168, opened 2019-04-24
Add hamcrest-core for Debian

cipherboy:
While Debian packages a useful junit4 (that already includes
hamcrest-core in its MANIFEST's Class-Path), we might as well include
the correct path in the auto detection script to silence the incorrect
warning.


cipherboy:
This removed the warning from the [Debian CI image](https://travis-ci.org/dogtagpki/jss/jobs/524119415#L2740) so I'm going ahead and merging.

PR #170, opened 2019-05-06
Define DEBUG and FORCE_PR_ASSERT in debug build

cipherboy:
This ensures that all `PR_ASSERT(...)` calls are processed and executed as
part of the test suite in a debug build. Otherwise, these calls are
ignored and their asserts are not checked by definition of `PR_ASSERT` in
the NSPR library.


cipherboy:
Low-risk patch, merging. This doesn't affect package builds or testing right now, as we only build in release mode (this should change...) :-)

PR #172, opened 2019-05-06
Update CI: Fedora 30

cipherboy:
[Fedora 30](https://fedoramagazine.org/announcing-fedora-30/) was released! Let's use it in the CI system. :)


cipherboy:
F30 worker has finished, I'm going ahead and merging this.

PR #178, opened 2019-05-13
Allow Travis Extras section to succeed quickly

cipherboy:
By giving a simple `true` script, the Extras section will have a job
which succeeds quickly. This should make the entire Travis run take
less time, as while the Extras section is fast-finish, it won't finish
until the first job succeeds.


cipherboy:
No response, though perhaps we should contact the actual Travis support instead of the community. But that's for another time. Merging this..

PR #187, opened 2019-05-20
Add nss.SSL.{CipherPrefSet,CipherPrefGet}

cipherboy:
These allow us to configure the set of supported cipher suites on one particular SSLSocket. Modifying the defaults seems out of scope as it doesn't play nicely across instances. We could add those in the future if we find them necessary. 


cipherboy:
Had to kick the F30 image a couple of times; modular repos apparently failed to sync. \o/

PR #190, opened 2019-05-30
Fix CHECK_DEPRECATION compatibility

cipherboy:
The old build system respected the `CHECK_DEPRECATION` environment
variable to enforce deprecation checks during (Java) compilation
time. In the new system, this was broken by not using the `ENV{}`
construct and not adding the flag to the correct variable.

This was reported by @emaldona.

Should be backported to `v4.5.x` branch as well. 


cipherboy:
This behavior can be seen here (pre CMake):
https://github.com/dogtagpki/jss/blob/v4.5.0/build_java.pl#L170-L174

In the future, we should think about adding this as an optional test if we ever get deprecation-free. Until then, let's hold off though.

PR #33, opened 2018-10-04
shellcheck build.sh

cipherboy:
This PR does two things:
 - Improves the quality of the `build.sh` script
 - Adds a style checker (`shellcheck`) to the repository that validates shell scripts.

Most of the issues found were minor style nits; a later PR might refactor the build script.

PR #195, opened 2019-06-05
Fix some trivial deprecation warnings #33

emaldona:
The scope is this patch is very narrow.

### DIFF
--- a/org/mozilla/jss/ssl/SSLTest.java
+++ b/org/mozilla/jss/ssl/SSLTest.java
@@ -56,7 +56,7 @@ public class SSLTest {
         dumpParams();
         CryptoManager.initialize(".");
 
-        int port = (new Integer( params.get("port") )).intValue();
+        int port = Integer.valueOf(params.get("port"));
 
         Socket s = new Socket(params.get("host"), port);
 
--- a/org/mozilla/jss/tests/JSSE_SSLClient.java
+++ b/org/mozilla/jss/tests/JSSE_SSLClient.java
@@ -18,6 +18,7 @@ import java.security.KeyStoreException;
 import java.security.NoSuchProviderException;
 import java.util.ArrayList;
 import java.util.Iterator;
+import java.lang.reflect.InvocationTargetException;
 
 import javax.net.ssl.HandshakeCompletedEvent;
 import javax.net.ssl.HandshakeCompletedListener;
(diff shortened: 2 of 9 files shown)


emaldona:
Notice a section that  has 
+    @Deprecated
which is not a fix but adding a deprecation warning. 

This diff with 8 lines of context should make it clear why it's needed.
diff --git a/org/mozilla/jss/ssl/SSLVersionRange.java b/org/mozilla/jss/ssl/SSLVersionRange.java
index 188887c5..8823f451 100644
--- a/org/mozilla/jss/ssl/SSLVersionRange.java
+++ b/org/mozilla/jss/ssl/SSLVersionRange.java
@@ -51,16 +51,17 @@ public class SSLVersionRange {
 
     /**
      * Used by the C code, do not use it directly
      * @deprecated Replaced with SSLVersionRange(SSLVersion minVersion, SSLVersion maxVersion).
      * @param min_enum
      * @param max_enum
      * @throws IllegalArgumentException
      */
+    @Deprecated
     public SSLVersionRange(int min_enum, int max_enum) throws IllegalArgumentException {
         this(SSLVersion.valueOf(min_enum), SSLVersion.valueOf(max_enum));
     }
 
     public SSLVersion getMinVersion() {
         return minVersion;
     }

org/mozilla/jss/tests/JSS_SelfServServer.java
```diff
@@ -112,13 +113,14 @@ public void doIt(String[] args) throws Exception {
...
             if (args.length >= 5)
-                port = new Integer(args[4]).intValue();
+                port = new Integer(Integer.valueOf(args[4]));
             if (args.length >=6 && args[5].equalsIgnoreCase("verbose")) {
                 bVerbose = true;
             }
             if (args.length >=7 && !args[6].equalsIgnoreCase("default")) {
                 fServerCertNick = args[6];
             }
+        } catch (NumberFormatException e) {
```
cipherboy: Out of curiousity, why adding this line? Seems like it should be caught with the next line. :)

org/mozilla/jss/ssl/SSLTest.java
```diff
@@ -56,7 +56,7 @@ public SSLTest(String[] args) {
         dumpParams();
         CryptoManager.initialize(".");
 
-        int port = (new Integer( params.get("port") )).intValue();
+        int port = (Integer.valueOf(params.get("port")));
```
cipherboy: You could probably remove the outer parenthesis while you're here.

org/mozilla/jss/tests/JSSE_SSLClient.java
```diff
@@ -274,13 +275,13 @@ private void initSocketFactory() {
                 System.setProperty("java.protocol.handler.pkgs",
                         "com.ibm.net.ssl.www.protocol.Handler");
                 java.security.Security.addProvider((java.security.Provider)
-                Class.forName("com.ibm.jsse2.IBMJSSEProvider2").newInstance());
+                Class.forName("com.ibm.jsse2.IBMJSSEProvider2").getConstructor().newInstance());
```
cipherboy: Looking at this, I'm not sure why we keep it around. I'm guessing that this test case probably doesn't work on the OpenJDK any more. It definitely isn't being run by our test harness but that's alright.

I'd probably update it to match the newer style (`main(...) throws Exception` and not catch the error) if I were picky, but like I said, not being used and likely won't be used, so I wouldn't bother.

emaldona: Agree, not worth the effort.

emaldona:
Will submit a revised patch.

emaldona:
Need some help, I am out of practice and forgot how to do revision to patches. I'll ask you online.

PR #196, opened 2019-06-06
[WIP] Test SSLEngine

cipherboy:
Initial test case for the [`SSLEngine`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html) implementation in #150. Will be rebased once that is merged.

Checklist of tests:
 - [x] We create a JSSEngine by default after registering the provider. 
 - [x] We can do a basic handshake.
 - [x] We can send data.
 - [ ] We fail to handshake with an invalid certificate.
 - [ ] We test client authentication.
 - [ ] We test HSM-backed keys.
 - [ ] We interoperate with and can handshake with JSSE's SSLEngine.
     - [ ] server: JSSE -- client: JSS
     - [ ] server: JSS -- client: JSSE
 - [ ] We follow CryptoPolicies by default. 
 - [ ] We interoperate with `openssl` or similar.

For reference, the Oracle standard test is in [my testbed](https://github.com/cipherboy/testbed/tree/master/java-sslengine).

---

Only the commits after "Begin `TestSSLEngine.java`" are unique to this PR. The rest are rebased from #150.


cipherboy:
Closing this in favor of placing tests in #150.

PR #198, opened 2019-06-06
Add ssl.javax.JSSParameters

cipherboy:
This class extends [`SSLParameters`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLParameters.html) to provide convenience wrappers for
interacting with NSS classes in a standard Java environment. In
particular, they allow for the parsing of cipher suites into and out of
`ssl.SSLCipher` enum values, and for handling the differences between the
Java standard form of TLS versions as a String, `ssl.SSLVersion` enum
values, and `ssl.SSLVersionRange` instances.

Blocks #150 as this is included in that range of commits.

### DIFF
--- a/org/mozilla/jss/ssl/javax/JSSParameters.java
+++ b/org/mozilla/jss/ssl/javax/JSSParameters.java
@@ -1,0 +1,164 @@
+package org.mozilla.jss.ssl.javax;
+
+import javax.net.ssl.*;
+import java.util.*;
+
+import org.mozilla.jss.ssl.*;
+
+/**
+ * JSSParameters is an implementation of SSLParameters to interoperate
+ * with NSS.
+ *
+ * In particular, we extend SSLParameters to provide cipher suites and
+ * protocol versions as NSS constants. This aids with the JNI translation
+ * layer. We also create a default set of SSLParameters which use a null
+ * cipher suite list and null protocol version list to fall back on the
+ * NSS default. On RHEL-like systems, this will default to CryptoPolicies.
+ */
+public class JSSParameters extends SSLParameters {
+    private SSLCipher[] suites;
+    private SSLVersionRange range;
+
+    public JSSParameters() {
+        // Choose our default set of SSLParameters here; default to null
+        // everywhere to rely on the default NSS policies.
+        super(null, null);
+    }
+
+    public JSSParameters(SSLParameters downcast) {
+        this();
+        if (downcast == null) {
+            return;
+        }
+
+        // Copy data from downcast
+        setCipherSuites(downcast.getCipherSuites());
+        setProtocols(downcast.getProtocols());
+        setAlgorithmConstraints(downcast.getAlgorithmConstraints());
+        setEndpointIdentificationAlgorithm(downcast.getEndpointIdentificationAlgorithm());
+        setNeedClientAuth(downcast.getNeedClientAuth());
+        setServerNames(downcast.getServerNames());
+        setSNIMatchers(downcast.getSNIMatchers());
+        setUseCipherSuitesOrder(downcast.getUseCipherSuitesOrder());
+        setWantClientAuth(downcast.getWantClientAuth());
+    }
+
+    public JSSParameters(String[] cipherSuites) {
+        super(cipherSuites);
+    }
+
+    public JSSParameters(String[] cipherSuites, String[] protocols) {
+        super(cipherSuites, protocols);
+    }
+
+    public void setCipherSuites(String[] cipherSuites) throws IllegalArgumentException {
+        if (cipherSuites == null || cipherSuites.length == 0) {
+            suites = null;
+            return;
+        }
+
+        ArrayList<SSLCipher> converted = new ArrayList<SSLCipher>();
+        for (String cipherSuite : cipherSuites) {
+            try {
+                converted.add(SSLCipher.valueOf(cipherSuite));
+            } catch (Exception e) {
+                throw new IllegalArgumentException("JSSParameters.setCipherSuites() - Unknown cipher suite (" + cipherSuite + "): " + e.getMessage(), e);
+            }
+        }
+
+        suites = converted.toArray(new SSLCipher[0]);
+    }
+
+    public void setCipherSuites(SSLCipher[] cipherSuites) {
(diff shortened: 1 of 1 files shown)


org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -0,0 +1,174 @@
...
+                converted.add(SSLCipher.valueOf(cipherSuite));
+            } catch (Exception e) {
+                throw new IllegalArgumentException("Unknown cipher suite (" + cipherSuite + "): " + e.getMessage(), e);
+            }
+        }
+
+        suites = converted.toArray(new SSLCipher[0]);
+    }
+
+    public void setCipherSuites(AbstractList<SSLCipher> cipherSuites) {
```
cipherboy: This actually isn't used any more, so we could get away with removing it.

org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        return ciphers.toArray(new String[0]);
+    }
+
+    public SSLCipher[] getSSLCiphers() {
+        return suites;
+    }
+
+    public void setProtocols(String[] protocols) {
```
cipherboy: Should probably add a `throws IllegalArgumentException` here.

org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * JSSParameters is an implementation of SSLParameters to interoperate
+ * with NSS.
+ *
+ * In particular, we extend SSLParameters to provide cipher suites and
+ * protocol versions as NSS constants. This aids with the JNI translation
+ * layer. We also create a default set of SSLParameters which use a null
+ * cipher suite list and null protocol version list to fall back on the
+ * NSS default. On RHEL-like systems, this will default to CryptoPolicies.
```
cipherboy: Crypto Policies needs to be implemented; I think I'll make that a separate PR as it touches multiple parts of the new system.

(We can use @abbra's 1c1dad189073a9aa2fa5b5503c659967a25dc119 as a reference on the protocol side but there doesn't seem to be a corresponding commit for the SSLCipher side).

cipherboy:
@edewata What do you think about adding `String[] aliases` here as a parameter? We have a few problems:

 - `SSLParameters` only has a [`getServerNames()`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLParameters.html#getServerNames--) function.
 - [`SNIServerName`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SNIServerName.html) maps most closely to the list of subjects on a certificate, not a NSS DB alias.
 - It only references the server's preferred name, so if we're on the client end of the connection, `getServerName()` won't inform us of the certificates we should use for the client certificate.

The rest of this class does simplify the handling inside `SSLEngine` by parsing all of this for us though. Eventually we'll include `crypto-policies` support here as well. This would let us re-use it for [`jss.ssl.SSLSocket`](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/ssl/SSLSocket.html) as well.

org/mozilla/jss/ssl/javax/JSSParameters.java
edewata: OK

org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -0,0 +1,174 @@
...
+
+    public void setProtocols(String[] protocols) {
+        if (protocols == null || protocols.length == 0) {
+            range = null;
+            return;
+        }
+
+        try {
+            SSLVersion minProtocol = SSLVersion.findByAlias(protocols[0]);
+            SSLVersion maxProtocol = SSLVersion.findByAlias(protocols[0]);
```
edewata: It's possible to do this too:
```
SSLVersion maxProtocol = minProtocol;
```

org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -0,0 +1,174 @@
...
+
+            for (String protocol : protocols) {
```
edewata: Just a slight redundancy, the protocols[0] will be compared by itself.

org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * In particular, we extend SSLParameters to provide cipher suites and
+ * protocol versions as NSS constants. This aids with the JNI translation
+ * layer. We also create a default set of SSLParameters which use a null
+ * cipher suite list and null protocol version list to fall back on the
+ * NSS default. On RHEL-like systems, this will default to CryptoPolicies.
+ */
+public class JSSParameters extends SSLParameters {
+    private SSLCipher[] suites = null;
+    private SSLVersionRange range = null;
```
edewata: The null assignments are redundant. Attributes are null by default.

org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -0,0 +1,174 @@
...
+
+    public JSSParameters() {
+        // Choose our default set of SSLParameters here; default to null
+        // everywhere to rely on the default NSS policies.
+        super(null, null);
+
+        suites = null;
+        range = null;
```
edewata: Redundant too.

org/mozilla/jss/ssl/javax/JSSParameters.java
cipherboy: Yeah; it is easier than handling `null` in `{min,max}Protocol` or starting `protocols` from index 1. There's also what, 6 protocols?

`SSLv2`, `SSLv3`, `TLSv1.0`, `TLSv1.1`, `TLSv1.2`, and `TLSv1.3`

So anything else is really overkill and this is provided for compatibility with the official standards; in TomcatJSS, we'll likely construct from either a `SSLVersion` pair or `SSLVersionRange` directly.

PR #209, opened 2019-06-17
Support LD_FLAGS from environment

cipherboy:
This is necessary for downstream builds. 


cipherboy:
Need to backport to `v4.5.x` branch

PR #211, opened 2019-06-18
Wrap SSL_ConfigServerCert

cipherboy:
`SSL_ConfigServerCert` (exposed as `jss.nss.SSL.ConfigServerCert`) is the
more modern form of `SSL_ConfigSecureServer`. Implement the wrapper with
no additional data (i.e., default configuration) and test it. Also test
ECDSA signed certificates in both C and Java BufferPRFD test suites.

Deprecate `SSL_ConfigSecureServer` as well.

At some point, we could include an additional form with additional data. 


org/mozilla/jss/tests/TestBufferPRFD.java
```diff
@@ -140,15 +146,15 @@ public synchronized static void TestSSLHandshake(String database, String passwor
                 int error = PR.GetError();
 
                 if (error != PRErrors.WOULD_BLOCK_ERROR) {
-                    System.out.println("Unexpected error: " + error);
+                    System.out.println("Unexpected error: " + new String(PR.ErrorToName(error)) + " (" + error + ")");
```
edewata: Just wondering, why not make `PR.ErrorToName()` return a `String` object directly? The orinal `PR_ErrorToName()` returns a `char*` which presumably is meant to be used as a string.

Same question about `PR.GetErrorText()`.

org/mozilla/jss/tests/TestBufferPRFD.java
```diff
@@ -93,19 +95,23 @@ public synchronized static void TestSSLVersionGetSet(PRFDProxy s_nspr) throws Ex
...
-        String host = "localhost";
-        byte[] peer_info = host.getBytes();
-
-        /* Find SSL Certificate */
+    public static void InitializeCM(String database, String password) throws Exception {
         CryptoManager manager;
         CryptoManager.initialize(database);
         manager = CryptoManager.getInstance();
         manager.setPasswordCallback(new Password(password.toCharArray()));
         CryptoToken token = manager.getInternalKeyStorageToken();
```
edewata: Probably an existing issue. The `token` object doesn't seem to be used.

cipherboy: ACK, I can drop that.

org/mozilla/jss/tests/TestBufferPRFD.java
cipherboy: Ah good point, I could do that. The issue is that `jss/util/jssutil.h` doesn't currently have methods for converting `char *` to `jstring` so I'd want to add that wrapper first.

cipherboy: Addressed in #212.

PR #213, opened 2019-06-19
Switch to SSL_ConfigServerCert

cipherboy:
In `SSLServerSocket.c`, we use the deprecated form, `SSL_ConfigSecureServer`.
Switch to using the newer form, `SSL_ConfigServerCert`. This also saves us
a call to check the KEA usage.

`SSL_ConfigSecureServer` [deprecation notice](https://github.com/nss-dev/nss/blob/master/lib/ssl/ssl.h#L986-L998).

`SSL_ConfigServerCert` [description](https://github.com/nss-dev/nss/blob/master/lib/ssl/ssl.h#L933-L984).

NSS likely won't remove `SSL_ConfigSecureServer`, but we might as well do our part to help with its deprecation. Since we're just querying KEA directly, passing the default `NULL` struct will do the same. 


cipherboy:
I think this will be fine. We're not planning on backporting it and we don't otherwise modify KEA.

PR #217, opened 2019-07-19
 Fixing uncleared password in MacData

edewata:
The MacData has been modified to call PBEKeyGenParams.clear()
in a finally block to ensure that the password is cleared.


cipherboy:
Looks good, thanks @edewata!

PR #219, opened 2019-07-19
Add alternative HMAC names for JSSProvider

cipherboy:
As mentioned in [`dogtagpki/pki pr#230`](https://github.com/dogtagpki/pki/pull/230), we should add these to the JSS provider to avoid the need for `CryptoUtil.getHMACAlgName(...)`. 


cipherboy:
These are, I think, easy to merge. There's nothing really changing here other than adding a few more aliases. We can debate whether or not we want these aliases and what the scope of them is, but I think in general they're fine as convenience helpers. 

The other interesting thing is that SHA 224 isn't exposed here at all. According to the [Oracle Java Cryptographic Provider docs](https://docs.oracle.com/javase/7/docs/technotes/guides/security/SunProviders.html), SHA 224 just isn't defined in their standard. I think we could expose this at a later time if we wanted to.

PR #220, opened 2019-07-19
Remove tests/all.pl - old test suite runner

cipherboy:
Since 4.5.1, we've replaced the test suite runner with a CMake/ctest
implementation. At this point, we're preferring `SSLEngine` over
`SSLSocket`, so keeping the last few remaining test cases in `all.pl` (that
we're not running anyways) doesn't do much.

`tests/all.pl` is tracked in the jss archive for historical reference:

https://github.com/dogtagpki/jss-archive/blob/master/org/mozilla/jss/tests/all.pl


cipherboy:
Got internal feedback from @ladycfu; merging. Thanks!

PR #222, opened 2019-07-22
Fix SHA1-HMAC in SymmetricKey

cipherboy:
In `SymmetricKey.java`, SHA-1 HMAC was defined as [`PBA_SHA1_HMAC`](http://docs.oasis-open.org/pkcs11/pkcs11-curr/v2.40/os/pkcs11-curr-v2.40-os.html#_Toc228894782), a version
of HMAC for extending a password (using a salt) and turning it into a
key. Usually when one requests HMAC, one expects vanilla HMAC; use this
instead. We expose the old `SHA1_HMAC` (with PBA) as `PBA_SHA1_HMAC` now.
Note that later SHA-2 and SHA-3 algorithms lack PBA-based HMACs (in PKCS
v2.40 and v3.0 standards).

### DIFF
--- a/org/mozilla/jss/crypto/KeyGenAlgorithm.java
+++ b/org/mozilla/jss/crypto/KeyGenAlgorithm.java
@@ -108,6 +108,12 @@ public class KeyGenAlgorithm extends Algorithm {
             "PBA/SHA1/HMAC", new FixedKeyStrengthValidator(160),
             null, PBEKeyGenParams.class );
 
+    public static final KeyGenAlgorithm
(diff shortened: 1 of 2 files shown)


jmagne:
So this looks simple enough. I just wonder if we have any simple tests that exercise things?

cipherboy:
This algorithm in particular (HMAC-SHA1) gets exercised as part of the existing HMAC tests.

In a later PR, (once #224 is merged), I'll add more extensive tests for other HMAC algorithms by directly constructing the key versus doing the wrap + unwrap construction.

PR #224, opened 2019-07-22
Add enum for Algorithm's PKCS #11 Constants

cipherboy:
We need to map PKCS11 Constants between their names and their values and
also between their values and Algorithm identifiers. To do so, we add
PKCS11Algorithm which enumerates identifiers common to both Algorithm
and PKCS11Constants. This will let us call into the NSS API with the
correct constant identifier from PKCS11, when needed.

(More concretely, we can now call into the PKCS11 NSS API without resolving the values from Algorithm if we desire. This will let us upcast generic Keys (which are, oddly, instances of Algorithms) into `PK11SymKey` objects (which contain a NSS pointer)). 

### DIFF
--- a/org/mozilla/jss/crypto/Algorithm.java
+++ b/org/mozilla/jss/crypto/Algorithm.java
@@ -127,6 +127,10 @@ public class Algorithm {
         return false;
     }
 
+    public PKCS11Algorithm getEnum() {
+        return PKCS11Algorithm.valueOfIndex(this.oidIndex);
+    }
+
     /**
      * Index into the SECOidTag array in Algorithm.c.
      */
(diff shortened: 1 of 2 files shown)


org/mozilla/jss/tests/TestPKCS11Constants.java
```diff
@@ -70,4 +72,24 @@ public static void main(String[] args) throws Exception {
...
         }
     }
+
+    public static void testMap() throws Exception {
+        String name = "CKM_AES_KEY_GEN";
+        long value = org.mozilla.jss.pkcs11.PKCS11Constants.CKM_AES_KEY_GEN;
+        short id = 32;
+
+        assert(PK11ConstMap.byName(name) == value);
+        assert(PK11ConstMap.byValue(value).equals(name));
```
edewata: I'm not sure this is a valid test since the values of the constants in PKCS11Constants are not unique. Perhaps the mapping should only be limited to a subset of the constants that are unique?

edewata: I'm not sure if mapping a value to multiple names would be useful since there might be no way to determine the ultimate result (i.e. a single name).

Could we group these constants based on their prefixes (i.e. CKN, CKF, etc.) and assume that they are unique within their corresponding group? If that's the case, we might be able to create separate mapping for each group, e.g.:
```
    private static Map<Long, String> map_ckn_value_name;
    private static Map<Long, String> map_ckf_value_name;
```
or use nested mapping, e.g. map(prefix, value) -> name:
```
    private static Map<String, Map<Long, String>> map_prefix_value_name;
```
Another option is to define a new enumeration that links the constants in PKCS11Constants and Algorithm, e.g.:
```
public enum PKCS11Algorithm {
    CKM_RSA_PKCS_KEY_PAIR_GEN(
        Algorithm.CKM_RSA_PKCS_KEY_PAIR_GEN,
        PKCS11Constants.CKM_RSA_PKCS_KEY_PAIR_GEN),
    CKM_DSA_KEY_PAIR_GEN(
        Algorithm.CKM_DSA_KEY_PAIR_GEN,
        PKCS11Constants.CKM_DSA_KEY_PAIR_GEN),
    ...

    private int index;
    private int value;
}

cipherboy: @edewata I decided to go the `enum` path, and limit it in scope to only those algorithms common to PKCS11Constants and Algorithm.

cipherboy:
To clarify -- at a later time, if we're still interested in whole mappings of PKCS11Constants value <-> names, we can add it. Otherwise, we'll avoid adding it for now.

PR #225, opened 2019-08-07
fix typo in JUnit CMake variable

Dessa:
might need to get backported to 4.5 if so desired

also for jaxb, the errormessage refers to javaee-jaxb-api.jar while only checking for jaxb-api.jar but i don't know how common that name actually is so i didn't touch that part

### DIFF
--- a/cmake/JSSConfig.cmake
+++ b/cmake/JSSConfig.cmake
@@ -243,7 +243,7 @@ macro(jss_config_java)
         message(WARNING "Test dependency sfl4j-jdk14.jar not found by find_jar! Tests might not run properly.")
     endif()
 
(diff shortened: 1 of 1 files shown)


cipherboy:
ACK, LGTM. Thanks @Dessa!

cipherboy:
To give a little more context, when I wrote the initial CMake build system, I was having issues with supporting the Debian build system as well. I think we've since gotten that taken care of, but I wanted to make sure whoever got that message knew that jaxb came from (what was then) JavaEE (and is now Jakarta or similar under the Eclipse foundation).

The only major thing I think I'm still looking for in the CMake build system is support for installing JSS and then a common FindJSS implementation. Currently that is all handled downstream on Debian and Fedora; it might be nice to have upstream as well.

Issue #226, opened 2019-08-07
Support ECGenParameterSpec in PK11KeyPairGenerator


ZuluForce:
> JSS, jss4.jar, is still built with JDK 1.4.2. While JDK 1.4.2 is EOL'd and all new product development should be using the latest JavaSE, legacy business products that must use JDK 1.4 or 1.5 can continue to add NSS/JSS security fixes/enhancements.

That quote is from https://www.dogtagpki.org/wiki/JSS and the warning on that page says it is sorely out of date. I am here questioning if Java 1.4 support is really still necessary.

Here's the problem I am running into right now. I am attempting to use JSS as the JCE provider for JSSE TLS 1.2 connections. I would much rather use an `SSLEngine` implementation backed by JSS which I know you're working on but I don't expect to be ready in a timeframe I need.

I have TLS connections functioning when the key agreement is DH[E] but not ECDH[E]. In ECDH the key agreement attempts to create an EC key using the `ECGenParameterSpec` class. I have no control to change this but it does throw an exception in JSS' `PK11KeyPairGenerator` because this parameter spec only exists in Java 1.5+.

```
Caused by: java.security.InvalidAlgorithmParameterException
	at org.mozilla.jss.pkcs11.PK11KeyPairGenerator.initialize(PK11KeyPairGenerator.java:439)
	at org.mozilla.jss.crypto.KeyPairGenerator.initialize(KeyPairGenerator.java:75)
	at org.mozilla.jss.provider.java.security.JSSKeyPairGeneratorSpi.initialize(JSSKeyPairGeneratorSpi.java:46)
	at org.mozilla.jss.provider.java.security.JSSKeyPairGeneratorSpi$EC.initialize(JSSKeyPairGeneratorSpi.java:71)
	at java.base/java.security.KeyPairGenerator$Delegate.initialize(KeyPairGenerator.java:699)
	at java.base/sun.security.ssl.ECDHKeyExchange$ECDHEPossession.<init>(ECDHKeyExchange.java:112)
	... 39 more
```

The creation of JSS specific configuration classes may work for anyone using JSS directly but it's a terrible pattern if you want adoption in projects using standard frameworks or 3rd party code that expects a typical JCE contract. Is supporting Java 1.4 more important than supporting the framework users?

cipherboy:
@ZuluForce \o welcome from a former Minnesota native.

[...]

PR #230, opened 2019-08-08
WIP Add ECPublicKey / ECPrivateKey interfaces

cipherboy:
WIP; more details to follow. 

Related: #226.


lib/jss.map
```diff
@@ -394,3 +394,9 @@ Java_org_mozilla_jss_nss_Buffer_Free;
     local:
         *;
 };
+JSS_4.6.2 {
```
cipherboy: I've done a terrible job maintaining this map. @edewata, @jmagne, or @ladycfu -- do you know if it is possible to retroactively update the version numbers that functions were added in? :-)

org/mozilla/jss/pkcs11/PK11ECPrivateKey.java
```diff
@@ -22,21 +27,25 @@ protected PK11ECPrivateKey(byte[] pointer) {
...
-// requires JAVA 1.5
-//    public ECParams getParams() {
-//      try {
-//        return getECParams();
-//      } catch(TokenException te) {
-//            return null;
-//      }
-//    }
+    public ECParameterSpec getParams() {
+        try {
```
cipherboy: I think we can actually drop this try/catch block.

org/mozilla/jss/util/ECCurve.java
```diff
@@ -0,0 +1,199 @@
...
+                if (curve_oid.equals(oid)) {
+                    return curve;
+                }
+            }
+        }
+
+        return null;
+    }
+
+    // OID Base Arcs
```
cipherboy: @edewata how should I store these? If I introduce them in the `jss.util.EC` class, they'll be publicly accessible, which I guess is fine. Perhaps I should make a new `jss.util.ECOIDs` class for storing them? I can't use them in this enum because Java complains about forward references. I think it'd make the definition up top a lot easier and a lot less error prone. We can also deduplicate them out of `PKC11KeyPairGenerator` once we get a final resting place for them.

Thoughts?

cipherboy: I've introduced an ECOIDs class which I'll use for this.

cipherboy:
@edewata I think this is ready for review. I'm of the opinion to add more curves when we need them and not grow the Curve database too quickly. A lot of those curves simply don't get used.

org/mozilla/jss/pkcs11/PK11ECPublicKey.java
```diff
@@ -4,49 +4,37 @@
...
-        return new BigInteger( getWByteArray() );
-      } catch(NumberFormatException e) {
-          throw new RuntimeException("Unable to decode EC public value: " + e.getMessage(), e);
-      }
+    public ECPoint getW() {
+        byte[] pointData = getWByteArray();
+        return EC.decodeNSSPoint(pointData);
+    }
+
+    public byte[] getWBA() {
```
edewata: Instead of adding getWBA() we could simply make getWByteArray() public.

org/mozilla/jss/util/ECCurve.java
```diff
@@ -0,0 +1,161 @@
...
+        this.a = a;
+        this.b = b;
+        this.generatorX = generatorX;
+        this.generatorY = generatorY;
+        this.order = order;
+        this.cofactor = cofactor;
+    }
+
+    public String[] getNames() {
+        return this.names;
```
edewata: It's not necessary to use `this`.

org/mozilla/jss/util/ECCurve.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    public OBJECT_IDENTIFIER[] getOIDs() {
+        return this.oids;
```
edewata: It's not necessary to use `this`.

org/mozilla/jss/pkcs11/PK11ECPrivateKey.java
```diff
@@ -19,24 +24,21 @@ protected PK11ECPrivateKey(byte[] pointer) {
...
-// requires JAVA 1.5
-//    public ECParams getParams() {
-//      try {
-//        return getECParams();
-//      } catch(TokenException te) {
-//            return null;
-//      }
-//    }
+    public ECParameterSpec getParams() {
+        PK11PubKey publicKey = this.getPublicKey();
```
edewata: It's not necessary to use `this`.

org/mozilla/jss/pkcs11/PK11ECPublicKey.java
cipherboy: Ah, sure, good catch.

cipherboy: I'm confused why this is actually needed. I might remove it in the future, but for now I've done this.

cipherboy:
@edewata Ok, this is fine. I've tested this against PKI and [opened a PR](https://github.com/dogtagpki/pki/pull/243) addressing issues I found with `nssdb.py`.

After that, I was able to self-host a PKI instance with EC certs and have it work with Tomcat just fine: 

```
$./testssl.sh https://localhost.localdomain:8443

~snip~

 Testing server defaults (Server Hello) 

 TLS extensions (standard)    "renegotiation info/#65281" "extended master secret/#23"
 Session Ticket RFC 5077 hint no -- no lifetime advertised
 SSL Session ID support       yes
 Session Resumption           Tickets no, ID: no
 TLS clock skew               0 sec from localtime
 Signature Algorithm          ECDSA with SHA512
 Server key size              EC 521 bits
 Server key usage             Digital Signature, Data Encipherment, Key Agreement
 Server extended key usage    TLS Web Server Authentication
 Serial / Fingerprints        03 / SHA1 AE46DF88ED2D93E539BB90665B4DE32EB1ABAB1B
                              SHA256 7F7BF4E7B196CC8A7B98189CF3710FCD97298732364B151578EED931F83A60EC
 Common Name (CN)             localhost.localdomain
 subjectAltName (SAN)         localhost.localdomain 
 Issuer                       CA Signing Certificate (EXAMPLE)
 Trust (hostname)             Ok via SAN and CN (same w/o SNI)
 Chain of trust               NOT ok (self signed CA in chain)
 EV cert (experimental)       no 
 ETS/"eTLS", visibility info  not present
 Certificate Validity (UTC)   719 >= 60 days (2019-08-22 20:32 --> 2021-08-11 20:32)
 # of certificates provided   2
 Certificate Revocation List  --
 OCSP URI                     http://localhost.localdomain:8080/ca/ocsp
 OCSP stapling                not offered
 OCSP must staple extension   --
 DNS CAA RR (experimental)    not offered
 Certificate Transparency     --
```

Is there anything else I need to test or can I go ahead and merge this?

PR #232, opened 2019-08-09
Increase size of async io buffer in TestBufferPRFD

cipherboy:
NSS on s390x seems to ignore the semantics of async io/`EWOULDBLOCK`:
rather than retrying the failing write, it drops the data and refuses to
continue. This manifests itself as RSA-based TLS handshakes failing,
because the certificate and associated packets exceeded the size of the
internal buffer (2048).

Bumping the buffer for the tests to 4096 plasters over the issue (so
long as we only include one certificate and not the entire chain).

redhat-bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1730109

### DIFF
--- a/org/mozilla/jss/tests/TestBufferPRFD.c
+++ b/org/mozilla/jss/tests/TestBufferPRFD.c
@@ -292,8 +292,8 @@ int main(int argc, char** argv)
      * took access (or created access itself), we'd need to get access to
      * them befor giving it to NSS, as NSS wraps our PRFileDesc in one of
      * their PRFileDescs, removing our access to fd->secret. */
-    j_buffer *c_read_buf = jb_alloc(2048);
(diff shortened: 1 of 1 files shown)


cipherboy:
Right, I initially introduced this code as a test case for the buffer interface I wrote (and to ensure NSS worked the way I expected it to). 

The open BZ (bug against NSS) hasn't really made much progress. Since this is a test case, we can really set the buffer to whatever we want. The JDK sets a buffer of ~16kb, I set a smaller limit of 4k (since I'm assuming that NSS has some internal buffer of its own (which appears to be the case based upon the above tests, especially on non-s390x architectures).

We might need to bump this to a value closer to 16k if/when I send the entire certificate chain (instead of just the server cert). -- with a CA, intermediate, and leaf cert each of RSA-4096, you're going to want the extra space I guess.

PR #233, opened 2019-08-09
Support ECGenParameterSpec in PK11KeyPairGenerator

cipherboy:
~Based on #230. Will be rebased once that gets reviewed and merged.~

The standard Java interface for EC key pair generation only includes a
name. Use the ECCurve enum to resolve this to an OID for use by NSS.
    
A similar patch was initially proposed by Andrew Helgeson (@ZuluForce)
in upstream issue https://github.com/dogtagpki/jss/issues/226.
    
### DIFF
--- a/org/mozilla/jss/pkcs11/PK11KeyPairGenerator.java
+++ b/org/mozilla/jss/pkcs11/PK11KeyPairGenerator.java
@@ -5,6 +5,7 @@
 package org.mozilla.jss.pkcs11;
 
 import java.math.BigInteger;
+import java.security.AlgorithmParameters;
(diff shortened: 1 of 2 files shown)


cipherboy:
Ok. This PR fails on the Ubuntu 18.04 image because NSS is broken (actually, #230 was the broken PR). In particular, `SECKEY_ConvertToPublicKey` returns `NULL`, setting `SEC_ERROR_BAD_DATA` when converting an EC private key to an EC public key. I'm not sure what's up, but this seems to be an issue with NSS @ 3.35. Debian Testing is fine so I'm merging.

PR #238, opened 2019-08-14
Move buffer.{c,h} -> j_buffer.{c,h}

cipherboy:
Naming a file "buffer.h" on MacOS has the unfortunate consequence of
conflicting with other includes. Rename our copy to j_buffer.h to
avoid this problem.

Reported by @c42-arthur in issue #226.


cipherboy:
Thanks!

PR #239, opened 2019-08-15
Fix BufferPRFD's PRBufferGetSocketOption

cipherboy:
`PRBufferGetSocketOption` takes two parameters: the `PRFileDesc` that we're
operating on and a `PRSocketOptionData` where we place the result. I
incorrectly treated this as a struct holding all options for a socket.
In reality, it contains two fields:
    
  - the option requested
  - the value of that option (via a union)
    
We thus need to condition on the option requested and return *only* its
value.
    
Under the previous implementation, we clobbered all options we set,
except the last one, `data->value.send_buffer_size`. In TestBufferPRFD.c,
we set the capacity of the buffer as 2048. Since `sizeof(PRSize) >= sizeof(PRBool)`, we did not perform an out of bounds write. On big endian
systems such as s390x, we stored the value 0x00000000 00000800: this
meant accessing `data->value.non_blocking` returned `PR_FALSE` (0x00).
    
Since the condition in `ssl_FdIsBlocking` is `!opt.value.non_blocking`,
this resulted in NSS assuming our buffer was blocking.
    
Many thanks to Bob Relyea for finding this.
    
Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1730109
    

cipherboy:
Verified on s390x, merging.

PR #240, opened 2019-08-21
Support favoring other Cryptographic Providers

cipherboy:
We're previously given the option to either install the JSS provider or
not, and optionally to remove the Sun Provider. Users can re-order the
providers later if they want, but the most common other option will be
to only use JSS for specified operations. To achieve this, we put JSS as
the very last provider. Hence, add:

    installJSSProviderFirst = true

as a new value in InitializationValues.

### DIFF
--- a/org/mozilla/jss/CryptoManager.java
+++ b/org/mozilla/jss/CryptoManager.java
@@ -522,8 +522,12 @@ public final class CryptoManager implements TokenSupplier
         logger.debug("Loaded " + kt);
 
         if( values.installJSSProvider ) {
-            int position = java.security.Security.insertProviderAt(
-                            new JSSProvider(), 1);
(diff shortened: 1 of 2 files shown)


edewata:
As discussed on IRC, the patch works, but maybe the ideal solution is to modify everything that requires JSS to explicitly use JSS provider so that the order of system providers does not matter.

cipherboy:
Well, I think this is a good PR to have from an upstream provider, I agree that we can explicitly call into JSS in TomcatJSS / PKI if we want to.

But the real PR that we want to have in JSS is to use the system trust store. *That* is what will really fix your issue I think.

cipherboy:
Pending a system trust store sensitive TrustManager, I'm going to go ahead and merge this as well.

PR #241, opened 2019-08-22
Fix debug builds

cipherboy:
Debug builds (via setting CMake release types) used to work when it was first introduced. This fixes support for them:

 - Fix signed vs unsigned comparisons,
 - Fix uninitialized use of a variable,
 - Fix an assertion error by defining PR_Read and PR_Write


cipherboy:
Ok with the new commit it works with clang, but when `-Werror` is specified, clang needs the `-Wno-unused-command-line-argument` parameter to build. I think we could fix this eventually but since we don't really build with clang anywhere outside of MacOS, we don't strictly need it.

PR #248, opened 2019-08-26
WIP: OCSP-sensitive TrustManager

cipherboy:
Opening this as WIP PR -- some portions might get split into other, smaller PRs, else we might drop this in favor of another approach.

This approach assumes we wish to fully integrate into the Java ecosystem. We introduce a TrustManager that understands OCSP policy and can downcast java.X509Certificates into jss.X509Certificates. This lets us run full OCSP-checking against them.

The upside of this PR is that it integrates nicely into other SSLEngines if they understand TrustManager and need to support OCSP checking. 

The downside of this PR is that we need to call from Java, through JNI, to NSS to configure a callback to check the TrustManager. Going backwards, when the callback fires, we need to call from NSS's callback into Java, get the TrustManager instance somehow, and report back the status. This all seems slow and complicated. Alternatively, we could add a null handler that always reports success and do the work in the SSLEngine directly on the extracted certificates. However this means we might not be able to send a badcert alert. 

Some work should be done to make this more configurable (chain, leaf, chain and leaf, neither -- for both OCSP and CRL). Perhaps we want to make this the default TrustManager and make the other one a basic version.

### DIFF
--- a/lib/jss.map
+++ b/lib/jss.map
@@ -397,6 +397,7 @@ Java_org_mozilla_jss_nss_Buffer_Free;
 JSS_4.6.2 {
     global:
 Java_org_mozilla_jss_pkcs11_PK11PrivKey_getPublicKey;
+Java_org_mozilla_jss_CryptoManager_importDERCertNative;
     local:
         *;
 };
--- a/org/mozilla/jss/CertificateUsage.java
+++ b/org/mozilla/jss/CertificateUsage.java
@@ -12,6 +12,7 @@ import java.util.Iterator;
 public final class CertificateUsage {
 
     private int usage;
+    private int value;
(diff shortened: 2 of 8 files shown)


cipherboy:
PRs #267, #268, #269, and #270 have been pulled out of this PR for ease of review.

cipherboy:
I'm closing this one in favor of #579. We can us `JSSNativeTrustManager` (like we already do in TomcatJSS) for OCSP sensitive `TrustManager`s and then use #579 for other, external TrustManagers. This code might be useful if we eventually decide to do OCSP checking in `JSSTrustManager`, so I'll leave the branch around, but otherwise I don't think there's value in keeping it.

PR #249, opened 2019-08-26
Native TrustManager

cipherboy:
~Opening this as a WIP PR. This PR is fairly self-contained unlike #248.~

This approach assumes we wish to use the already-existing OCSP validation logic in [`JSSL_DefaultCertAuthCallback`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/ssl/callbacks.c#L464). To do so we add a dummy TrustManager class which doesn't do validation, but which we can use in the SSLEngine as a sentinel for using this validation (note that we can use `NULL` to use the default NSS certificate validation, which is again something different, or use `NULL` to choose a sane default and create a separate class for default NSS validation). 

The upside of this PR is that it doesn't involve invoking Java from the NSS callback, which means it is more performant than other solutions. All other solutions require several JNI calls (one from the callback handler into the TrustManager, and another from 

The downside is this TrustManager is very specific to our SSLEngine and won't work if we try and use it anywhere else (e.g., in PKI for some reason, outside of SSLEngine, to directly validate certs).

The work to make this more configurable would all live in the CryptoManager. In an ideal world, I would not really happy with defaulting to no CRL checking though. This'd align nicely with the existing OCSP Policy options.

### DIFF
--- a/lib/jss.map
+++ b/lib/jss.map
@@ -429,6 +429,7 @@ Java_org_mozilla_jss_crypto_KBKDFOptionalCounterParam_releaseNativeResources;
 Java_org_mozilla_jss_crypto_Policy_getRSAMinimumKeySize;
 Java_org_mozilla_jss_crypto_Policy_getDHMinimumKeySize;
 Java_org_mozilla_jss_crypto_Policy_getDSAMinimumKeySize;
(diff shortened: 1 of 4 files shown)


cipherboy:
@edewata As a default, I think this is sane to fallback on. What do you think? I've added support and tested it in #150 / #196.

cipherboy:
Thanks @edewata!

PR #250, opened 2019-08-26
Remove misleading indentation in ssl/callbacks.c

cipherboy:
This fixes three things while we're here:

 - Misleading indentation
 - Trailing whitespace
 - Unnecessary NULL checks prior to destroy


cipherboy:
I've simplified the diff. A full format will be done later I think.

PR #258, opened 2019-09-17
Add additional PCKS11 Constants

cipherboy:
This includes the new `CKM_AES_CMAC` and `CKM_AES_CMAC_GENERAL` constants.


cipherboy:
PKCS#11 check doesn't pass because `CKM_AES_CMAC{,_GENERAL}` isn't shipped in Fedora yet, but that's fine.

PR #259, opened 2019-09-17
Add CMAC to JSS

cipherboy:
~This is WIP as it depends on #258 being merged. This series will be rebased afterwards. The commits will be squashed and rewritten.~

This adds initial support for CMAC to the JSS Provider interface. This depends on a recent enough NSS that includes [CMAC](https://bugzilla.mozilla.org/show_bug.cgi?id=1570501) support in PKCS#11. ~Thus CI will fail until I introduce a new instance with Sandboxed NSS builds.~

I've added build time support for CMAC. What this means is that JSS only needs a rebuild in order to take advantage of CMAC support in NSS. Otherwise, it'll ignore the non-existence of CMAC and not run the applicable test from the test suite. CMAC will still be defined in the JSS provider though.

You can see this via checking the output of CI: currently there are 66 tests without CMAC, 1 CMAC test. So only `fedora_sandbox` CI image will have 67 test cases.

### DIFF
--- a/org/mozilla/jss/crypto/Algorithm.c
+++ b/org/mozilla/jss/crypto/Algorithm.c
@@ -18,6 +18,11 @@
 static PRStatus
 getAlgInfo(JNIEnv *env, jobject alg, JSS_AlgInfo *info);
 
+/* Helpers to handle differences in NSS versions. */
+#ifndef CKM_AES_CMAC
+#define CKM_AES_CMAC CKM_INVALID_MECHANISM
+#endif
+
 /***********************************************************************
 **
 **  Algorithm indices.  This must be kept in sync with the algorithm
@@ -97,7 +102,10 @@ JSS_AlgInfo JSS_AlgTable[NUM_ALGS] = {
 /* 66 */    {CKM_AES_KEY_WRAP_PAD, PK11_MECH},
 /* 67 */    {CKM_SHA256_HMAC, PK11_MECH},
 /* 68 */    {CKM_SHA384_HMAC, PK11_MECH},
-/* 69 */    {CKM_SHA512_HMAC, PK11_MECH}
+/* 69 */    {CKM_SHA512_HMAC, PK11_MECH},
+
+/* CKM_AES_CMAC is new to NSS; some implementations might not yet have it. */
+/* 70 */    {CKM_AES_CMAC, PK11_MECH}
 /* REMEMBER TO UPDATE NUM_ALGS!!! (in Algorithm.h) */
 };
 
--- a/org/mozilla/jss/crypto/CMACAlgorithm.java
+++ b/org/mozilla/jss/crypto/CMACAlgorithm.java
@@ -1,0 +1,69 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.jss.crypto;
+
+import java.security.NoSuchAlgorithmException;
+import java.util.Hashtable;
+
+import org.mozilla.jss.asn1.OBJECT_IDENTIFIER;
+
+/**
+ * Algorithms for performing CMACs. These can be used to create
+ * MessageDigests.
+ */
+public class CMACAlgorithm extends DigestAlgorithm {
+
+    protected CMACAlgorithm(int oidIndex, String name, OBJECT_IDENTIFIER oid,
+                            int outputSize) {
+        super(oidIndex, name, oid, outputSize);
+
+        if (oid != null && oidMap.get(oid) == null) {
+            oidMap.put(oid, this);
+        }
+    }
+
+    ///////////////////////////////////////////////////////////////////////
+    // OID mapping
+    ///////////////////////////////////////////////////////////////////////
+    private static Hashtable<OBJECT_IDENTIFIER, CMACAlgorithm> oidMap = new Hashtable<>();
+
+    /**
+     * Looks up the CMAC algorithm with the given OID.
+     *
+     * @param oid OID.
+     * @return CMAC algorithm.
+     * @exception NoSuchAlgorithmException If no registered CMAC algorithm
+     *  has the given OID.
+     */
+    public static CMACAlgorithm fromOID(OBJECT_IDENTIFIER oid)
+        throws NoSuchAlgorithmException
+    {
+        CMACAlgorithm alg = oidMap.get(oid);
+        if (alg == null) {
+            throw new NoSuchAlgorithmException("No such algorithm for OID: " + oid);
+        }
+
+        return alg;
+    }
+
+    /**
+     * CMAC AES-X.  This is a Message Authentication Code that uses a
+     * symmetric key together with the AES cipher to create a form of
+     * signature.
+     *
+     * Note that we pass null for the OID here: neither NIST nor any other
+     * standards body has defined an OID for use with CMAC. Since we use
+     * a PKCS#11 backend and NSS doesn't otherwise define CMAC based on a
+     * SEC OID, we don't strictly need one.
+     *
+     * We've left the fromOID code (and oid parameter in the constructor) as
+     * other projects use them for HMACAlgorith. At such time as an OID is
+     * defined, it can be added here.
+     */
+    public static final CMACAlgorithm AES = new CMACAlgorithm(CKM_AES_CMAC, "AES-CMAC", null, 16);
+    public static final CMACAlgorithm AES128 = AES;
+    public static final CMACAlgorithm AES192 = AES;
+    public static final CMACAlgorithm AES256 = AES;
+}
--- a/org/mozilla/jss/pkcs11/PK11MessageDigest.java
+++ b/org/mozilla/jss/pkcs11/PK11MessageDigest.java
@@ -36,8 +36,8 @@ public final class PK11MessageDigest extends JSSMessageDigest {
         throws DigestException, InvalidKeyException
     {
 
-        if( ! (alg instanceof HMACAlgorithm) ) {
-            throw new DigestException("Digest is not an HMAC digest");
+        if( ! (alg instanceof HMACAlgorithm || alg instanceof CMACAlgorithm) ) {
+            throw new DigestException("Digest is not an HMAC or CMAC digest");
         }
 
         reset();
@@ -90,7 +90,7 @@ public final class PK11MessageDigest extends JSSMessageDigest {
     }
 
     public void reset() throws DigestException {
-        if( ! (alg instanceof HMACAlgorithm) ) {
+        if( ! (alg instanceof HMACAlgorithm || alg instanceof CMACAlgorithm) ) {
             // This is a regular digest, so we have enough information
             // to initialize the context
             this.digestProxy = initDigest(alg);
--- a/org/mozilla/jss/provider/javax/crypto/JSSMacSpi.java
+++ b/org/mozilla/jss/provider/javax/crypto/JSSMacSpi.java
@@ -11,7 +11,9 @@ import java.security.Key;
 import java.security.NoSuchAlgorithmException;
 import java.security.spec.AlgorithmParameterSpec;
 
+import org.mozilla.jss.crypto.CMACAlgorithm;
 import org.mozilla.jss.crypto.CryptoToken;
+import org.mozilla.jss.crypto.DigestAlgorithm;
 import org.mozilla.jss.crypto.HMACAlgorithm;
 import org.mozilla.jss.crypto.JSSMessageDigest;
 import org.mozilla.jss.crypto.SecretKeyFacade;
@@ -21,9 +23,9 @@ import org.mozilla.jss.crypto.TokenSupplierManager;
 class JSSMacSpi extends javax.crypto.MacSpi {
 
     private JSSMessageDigest digest=null;
-    private HMACAlgorithm alg;
+    private DigestAlgorithm alg;
 
-    protected JSSMacSpi(HMACAlgorithm alg) {
+    protected JSSMacSpi(DigestAlgorithm alg) {
       try {
         this.alg = alg;
         CryptoToken token =
@@ -116,4 +118,9 @@ class JSSMacSpi extends javax.crypto.MacSpi {
         }
     }
 
+    public static class CmacAES extends JSSMacSpi {
+        public CmacAES() {
+            super(CMACAlgorithm.AES);
+        }
+    }
 }
--- a/org/mozilla/jss/tests/TestCmac.java
+++ b/org/mozilla/jss/tests/TestCmac.java
@@ -1,0 +1,89 @@
+package org.mozilla.jss.tests;
+
+import java.util.Arrays;
+import java.util.Base64;
+import java.security.Key;
+
+import javax.crypto.*;
+import javax.crypto.spec.*;
+
+import org.mozilla.jss.*;
+import org.mozilla.jss.crypto.*;
+import org.mozilla.jss.util.*;
+
+public class TestCmac {
+    private static final byte[] NIST_128 = Base64.getDecoder().decode("K34VFiiu0qar9xWICc9PPA==");
+    private static final byte[] NIST_192 = Base64.getDecoder().decode("jnOw99oOZFLIEPMrgJB55WL46tJSLGt7");
+    private static final byte[] NIST_256 = Base64.getDecoder().decode("YD3rEBXKcb4rc67whX13gR81LAc7YQjXLZgQowkU3/Q=");
+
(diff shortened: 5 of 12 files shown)


cipherboy:
@edewata  Do you think it is worthwhile to detect whether or not CMAC exists as a constant and not run the test at all, instead of always running the test and checking if it exists inside the test? 

The former would allow someone to check the build logs / test stdout and see if it has CMAC support. The latter would require someone to check the output of the tests (which don't get shown on non-error and are written to `build/Testing/Temporary/LastTest.log`).

cipherboy:
@edewata Never mind, I've decided to go with the build-time detection. I think it'll make life easier for all of us.

cipherboy:
Ok, this works now:

[`fedora_30`](https://travis-ci.org/dogtagpki/jss/jobs/589563046#L3573):
> 100% tests passed, 0 tests failed out of 66

[`fedora_sandbox`](https://travis-ci.org/dogtagpki/jss/jobs/589563056#L4363):
> 100% tests passed, 0 tests failed out of 67

org/mozilla/jss/pkcs11/PK11MessageDigest.java
```diff
@@ -36,7 +36,7 @@ public void initHMAC(SymmetricKey key)
         throws DigestException, InvalidKeyException
     {
 
-        if( ! (alg instanceof HMACAlgorithm) ) {
+        if( ! (alg instanceof HMACAlgorithm || alg instanceof CMACAlgorithm) ) {
             throw new DigestException("Digest is not an HMAC digest");
```
edewata: Should we say "Digest is not an HMAC/CMAC digest" instead?

org/mozilla/jss/crypto/CMACAlgorithm.java
```diff
@@ -0,0 +1,60 @@
...
+            throw new NoSuchAlgorithmException();
+        } else {
+            return (CMACAlgorithm) alg;
+        }
+    }
+
+    /**
+     * CMAC AES-X.  This is a Message Authentication Code that uses a
+     * symmetric key together with the AES cipher to create a form of
+     * signature.
```
cipherboy: Expand this comment to detail why we pass `null` as the OID here:

 1. Nobody has defined OIDs yet for CMAC.
 2. While as @edewata points out, static methods aren't inherited (and thus we don't need `fromOID`), there are usages of `fromOID` and `toOID` (the latter being inherited from Algorithm) in pki-core.
 3. We're using PKCS#11-backed implementations here (not OID-backed implementations like e.g., SHA-1 is), so we don't need an OID either.

This will minimize the amount of work in case we later get OID assignments for CMAC.

org/mozilla/jss/tests/TestCmac.java
```diff
@@ -0,0 +1,75 @@
...
+import java.security.Key;
+
+import javax.crypto.*;
+import javax.crypto.spec.*;
+
+import org.mozilla.jss.*;
+import org.mozilla.jss.crypto.*;
+import org.mozilla.jss.util.*;
+
+public class TestCmac {
```
edewata: As discussed on IRC, please include the link to NIST's examples here.

org/mozilla/jss/provider/javax/crypto/JSSMacSpi.java
```diff
@@ -21,9 +23,9 @@
 class JSSMacSpi extends javax.crypto.MacSpi {
 
     private JSSMessageDigest digest=null;
-    private HMACAlgorithm alg;
+    private DigestAlgorithm alg;
```
cipherboy: @edewata This isn't quite right I now realize (I thought it wasn't, then though it was, now realize it isn't). 

DigestAlgorithm defines things like SHA-1, SHA-2, &c. 

HMACAlgorithm is a subclass which defines things like HMAC/SHA-1, HMAC-SHA-2-256, &c. These all require keys though, unlike digests (though, some digests are keyed, but none we define are).  

CMAC (the algorithm) isn't actually based off of a digest (hash algorithm), but is instead based off a (keyed) cipher. 

So technically, we should have a class Structure like:

```
Algorithm
  DigestAlgorithm
  MACAlgorithm
    CMACAlgorithm
    HMACAlgorithm
```

But currently we have the following class structure:

```
Algorithm
  DigestAlgorithm
    CMACAlgorithm
    HMACAlgorithm
```

Lastly, these are protected so end-users won't directly be constructing them. I see the following actions we could take:

 1. Be happy how it is.
 2. Add an explicit check for HMAC or CMAC subclass. 
 3. (A breaking change for anyone using HMACAlgorithm) Fix the class structure. 
 4. Make CMAC inherit from Algorithm or a new MACAlgorithm, not DigestAlgorithm, checking either for HMACAlgorithm or MACAlgorithm.

Note that any checking that we would do here is limited to developers adding new HMAC/CMAC/*MAC implementations and making their lives easier. 


Thoughts?

org/mozilla/jss/crypto/CMACAlgorithm.java
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * @param oid OID.
+     * @return CMAC algorithm.
+     * @exception NoSuchAlgorithmException If no registered CMAC algorithm
+     *  has the given OID.
+     */
+    public static CMACAlgorithm fromOID(OBJECT_IDENTIFIER oid)
+        throws NoSuchAlgorithmException
+    {
+        Object alg = oidMap.get(oid);
```
edewata: It should be possible to get the `alg` as `CMACAlgorithm` directly since the `oidMap` is defined as `Hashtable<OBJECT_IDENTIFIER, CMACAlgorithm>`:
```
CMACAlgorithm alg = oidMap.get(oid);
```

org/mozilla/jss/crypto/CMACAlgorithm.java
```diff
@@ -0,0 +1,60 @@
...
+        if (alg == null) {
+            throw new NoSuchAlgorithmException();
```
edewata: It might be nice to include the OID in the exception message to help troubleshooting.

org/mozilla/jss/provider/javax/crypto/JSSMacSpi.java
cipherboy: (UMAC is the only other MAC implementation I could see someone wanting to add, but NSS doesn't yet support it).

org/mozilla/jss/crypto/CMACAlgorithm.java
cipherboy: ACK, this would be a nice cleanup for DigestAlgorithm / HMACAlgorithm as well.

org/mozilla/jss/crypto/CMACAlgorithm.java
cipherboy: ACK, again, nice to apply elsewhere as well.

org/mozilla/jss/provider/javax/crypto/JSSMacSpi.java
edewata: I'm OK with #3 for a new JSS minor version (4.7?). If we want to keep the current minor version we can use any other options to provide the new functionality without introducing incompatible changes, but consider that as a temporary solution until we fix the structure. As long as we document the changes clearly it should be OK to require people to change their code when upgrading JSS.

cipherboy: How about we add that as a second patchset after this merges, since it involves changing both HMAC and CMAC, and we can do additional simplifications internally?

edewata: Sure, thanks!

org/mozilla/jss/tests/TestCmac.java
```diff
@@ -0,0 +1,89 @@
...
+            return NIST_128;
+        } else if (index < 8) {
+            return NIST_192;
+        } else if (index < 12) {
+            return NIST_256;
+        }
+
+        return null;
+    }
+
```
jmagne: Question:

In TPS/TKS we start with a symmetric key such as the master key and then send that down into the routine that calculates the CMAC. The key will be on the hsm possibly as a SymmetricKey object variable. How do we use this from JSS from that starting point? The test examples start from byte array keys, which is what you expect in the HMAC stuff. 

Would it be similar to what I found in JSS here:

// perform the digesting
        JSSMessageDigest digest = token.getDigestContext(HMACAlgorithm.SHA1);
        digest.initHMAC(key);
        byte[] digestBytes = digest.digest(toBeMACed);
Or is there a way to use the nice provider support we have here starting with a sym key object?

org/mozilla/jss/crypto/Algorithm.c
```diff
@@ -97,7 +102,10 @@ JSS_AlgInfo JSS_AlgTable[NUM_ALGS] = {
 /* 66 */    {CKM_AES_KEY_WRAP_PAD, PK11_MECH},
 /* 67 */    {CKM_SHA256_HMAC, PK11_MECH},
 /* 68 */    {CKM_SHA384_HMAC, PK11_MECH},
-/* 69 */    {CKM_SHA512_HMAC, PK11_MECH}
+/* 69 */    {CKM_SHA512_HMAC, PK11_MECH},
+
+/* CKM_AES_CMAC is new to NSS; some implementations might not yet have it. */
+/* 70 */    {CKM_AES_CMAC, PK11_MECH}
```
jmagne: Another question:
I assume by adding this mechanism , which nss now implements, is enough to make existing jni code in jss do this job for us? Just curious to take a look at what's going on down there. Is PK11MessageDigest.c where the magic is happening?

org/mozilla/jss/tests/TestCmac.java
cipherboy: So there's two interfaces and a class in question here... and a pointer to the PKI code might be nice. But I'll assume the following is correct. :)

Classes/Interfaces in question:

 - [`javax.crypto.SecretKey`](https://docs.oracle.com/javase/8/docs/api/javax/crypto/SecretKey.html) an interface,
 - [`org.mozilla.jss.crypto.SymmetricKey`](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/crypto/SymmetricKey.html) an interface which extends `SecretKey`,
 - [`org.mozilla.jss.crypto.SecretKeyFacade`](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/crypto/SecretKeyFacade.html) a class which implements the above `SecretKey` interface via wrapping a `SymmetricKey`-implementing instance,

Everything that JSS uses has to be an instance of `SecretKeyFacade`. Internally, the `key` member (of interface `SymmetricKey`) is usually of type [`org.mozilla.jss.pkcs11.PK11SymKey`](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/pkcs11/PK11SymKey.html). The latter includes additional information like algorithm, and a NSS `PK11SymKey` pointer. So when you say you have a `SymmetricKey` instance, you usually actually have a `PK11SymKey` instance. But anyways...

To go from `SymmetricKey` to something usable by the JSS Provider (and, not just the underlying crypto), your easiest option is to do something like:

```java
javax.crypto.SecretKey key = new org.mozilla.jss.crypto.SecretKeyFacade(mySymmetricKey);
```

That's really what [`JSSSecretKeyFactorySpi`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/provider/javax/crypto/JSSSecretKeyFactorySpi.java) does [under the covers](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/provider/javax/crypto/JSSSecretKeyFactorySpi.java#L61).

Note that `JSSSecretKeyFactorySpi` implements the [`javax.crypto.SecretKeyFactorySpi`](https://docs.oracle.com/javase/8/docs/api/javax/crypto/SecretKeyFactorySpi.html) SPI class, so it can eventually be called from the Java Provider [`javax.crypto.SecretKeyFactory`](https://docs.oracle.com/javase/8/docs/api/javax/crypto/SecretKeyFactory.html) member (when the provider is `Mozilla-JSS`).

That brings us full-circle: if you have bytes, you can use the SecretKeyFactory with a correct `spec` class to get a `SecretKey` instance to use, otherwise, if you already have a PKCS#11-backed key from JSS, you can just use `SecretKeyFacade` to wrap it.

org/mozilla/jss/crypto/Algorithm.c
cipherboy: Yep! That's some lovely magic.

So my understanding of PKCS#11 is a bit... rough, and Bob is definitely the one to ask these questions to. :)

But there's a few basic operations in PKCS#11: Sign, Verify, Encrypt, Decrypt, etc. Each of these correspond to a top-level function call in NSS: `PK11_Sign`, `PK11_Verify`, etc. So, once you've set up the context, you essentially get away with calling one of those methods, and then you'll have the result. But anyways... let's start at the top.

You create a new [`javax.crypto.Mac`](https://docs.oracle.com/javase/8/docs/api/javax/crypto/Mac.html) instance of type [`CmacAES`](https://github.com/cipherboy/jss/blob/add-cmac/org/mozilla/jss/JSSProvider.java#L244-L257) (or one of its aliases). This gets backed by a [`JSSMacSpi`](https://github.com/cipherboy/jss/blob/add-cmac/org/mozilla/jss/provider/javax/crypto/JSSMacSpi.java) instance of subclass [`CmacAES`](https://github.com/cipherboy/jss/blob/add-cmac/org/mozilla/jss/provider/javax/crypto/JSSMacSpi.java#L121-L125). All this does is initialize the `JSSMacSpi` instance with the `CMACAlgorithm.AES` algorithm. This is an instance of [`org.mozilla.jss.crypto.CMACAlgorithm`](https://github.com/cipherboy/jss/blob/add-cmac/org/mozilla/jss/crypto/CMACAlgorithm.java) with the correct PKCS#11 constant (er, proxies). [*]

Anyhow, so when you call `Mac.init`, it calls into [`PK11MessageDigest.initHMAC(...)`](https://github.com/cipherboy/jss/blob/add-cmac/org/mozilla/jss/pkcs11/PK11MessageDigest.java#L35-L58), which creates the [PKCS#11 mechanism](https://github.com/cipherboy/jss/blob/add-cmac/org/mozilla/jss/pkcs11/PK11MessageDigest.c#L61), makes sure the key is enabled [for signing](https://github.com/cipherboy/jss/blob/add-cmac/org/mozilla/jss/pkcs11/PK11MessageDigest.c#L71), and then creates the [PKCS#11 context](https://github.com/cipherboy/jss/blob/add-cmac/org/mozilla/jss/pkcs11/PK11MessageDigest.c#L82). 

Each call to `Mac.update` then calls [`PK11_DigestOp`](https://github.com/cipherboy/jss/blob/add-cmac/org/mozilla/jss/pkcs11/PK11MessageDigest.c#L127), and in the end [`PK11_DigestFinish`](https://github.com/cipherboy/jss/blob/add-cmac/org/mozilla/jss/pkcs11/PK11MessageDigest.c#L166) is called. And to finish; `PK11_DigestOp`/`PK11_DigestFinal` is merely a [nice wrapper](https://github.com/nss-dev/nss/blob/master/lib/pk11wrap/pk11cxt.c#L989-L1008) around all of those different PKCS#11 operations.

[*]: This is where I disagree with the construction of `org.mozilla.jss.crypto.Algorithm`: rather than handling PKCS#11-backed and SecOID-backed algorithms separately, they merge it into one big table of tuples (value, type). Then, they define pretty values like `Algorithm.CKM_AES_CMAC = 70`, which is an *index* and not the actual value. They could've instead made all PKCS#11-backed algorithms use, well, the underlying [`PKCS11Constants`](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/pkcs11/PKCS11Constants.html) values and not made a redundant mapping (that class was added by me since its counterpart [in the JDK](http://www.docjar.com/docs/api/sun/security/pkcs11/wrapper/PKCS11Constants.html) was removed in JDK9, but that's well before our time)... But that's neither here nor there, and we have [`org.mozilla.jss.crypto.PKCS11Algorithm`](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/crypto/PKCS11Algorithm.html) to map between them for us.

org/mozilla/jss/tests/TestCmac.java
cipherboy: See #271. I think we can support just passing SymmetricKey directly here and be better off.

jmagne: That is the info I wanted! Just double checking I can do what we need :)

org/mozilla/jss/crypto/Algorithm.c
jmagne: Excellent info. I will approve since it looked good to me, just wanted a couple of clarifications. I just wanted to make sure I missed some other code that calls down into your new nss cmac code. Thanks!

PR #260, opened 2019-09-18
Add Fedora rawhide based sandbox builds

cipherboy:
This tests the latest upstream NSPR and NSS from Mozilla, on the latest
Fedora rawhide container, and ensures that we can build and pass our
current test suite.

We also add the `fedora_sandbox` image as a required test in Travis. 

### DIFF
--- a/tools/Dockerfiles/fedora_sandbox
+++ b/tools/Dockerfiles/fedora_sandbox
@@ -1,0 +1,41 @@
+FROM fedora:rawhide
+
+# Install generic dependencies to build jss
+RUN true \
+        && dnf update -y --refresh \
+        && dnf install -y dnf-plugins-core gcc make rpm-build cmake \
+                          glassfish-jaxb-api nss-tools apache-commons-codec \
+                          apache-commons-lang gcc-c++ jpackage-utils slf4j \
+                          zlib-devel perl slf4j-jdk14 junit ninja-build \
+                          gyp gtest mercurial \
+        && dnf build-dep nss nspr jss -y \
+        && mkdir -p /home/sandbox \
+        && dnf clean -y all \
+        && rm -rf /usr/share/doc /usr/share/doc-base \
+                  /usr/share/man /usr/share/locale /usr/share/zoneinfo \
+        && true
+
+# Link in the current version of jss from the git repository
+WORKDIR /home/sandbox
+COPY . /home/sandbox/jss
+
+# Download and build NSPR and NSS
+RUN true \
+        && cd /home/sandbox \
+        && hg clone https://hg.mozilla.org/projects/nspr \
+        && hg clone https://hg.mozilla.org/projects/nss \
+        && cd nss \
+        && ./build.sh --enable-fips --enable-libpkix \
+        && true
+
+# Perform the actual RPM build
+WORKDIR /home/sandbox/jss
+CMD true \
+        && export SANDBOX=1 \
+        && rm -rf build \
+        && mkdir build \
+        && cd build \
+        && cmake .. \
+        && make all \
+        && ctest --output-on-failure \
+        && true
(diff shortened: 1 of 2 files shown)


cipherboy:
@SilleBille Should I keep `fedora_sandbox` in the default, Test section, or should I move it to the optional Extra section? It is always going to be slower than the other two Fedora builds since it needs to build NSPR and NSS from scratch, on top of building and testing JSS. Thoughts?

tools/Dockerfiles/fedora_sandbox
```diff
@@ -0,0 +1,41 @@
...
+WORKDIR /home/sandbox
+COPY . /home/sandbox/jss
+
+# Download and build NSPR and NSS
+RUN true \
+        && cd /home/sandbox \
+        && hg clone https://hg.mozilla.org/projects/nspr \
+        && hg clone https://hg.mozilla.org/projects/nss \
+        && cd nss \
+        && ./build.sh --enable-fips --enable-libpkix \
```
SilleBille: Question: After building nss, do you need to install or let jss know which nss to use?

Also, in the description, it says that both `nspr` and `nss` are built. But, I see only nss to be built. Doesn't `nspr` require to be built?

tools/Dockerfiles/fedora_sandbox
```diff
@@ -0,0 +1,41 @@
...
+
+# Install generic dependencies to build jss
+RUN true \
+        && dnf update -y --refresh \
+        && dnf install -y dnf-plugins-core gcc make rpm-build cmake \
+                          glassfish-jaxb-api nss-tools apache-commons-codec \
+                          apache-commons-lang gcc-c++ jpackage-utils slf4j \
+                          zlib-devel perl slf4j-jdk14 junit ninja-build \
+                          gyp gtest mercurial \
+        && dnf build-dep nss nspr jss -y \
```
SilleBille: for `jss`, when you build-dep on the koji version, the CI will fail in case spec file is changed. Probably do a wget on the latest spec file? :)

tools/Dockerfiles/fedora_sandbox
cipherboy: NSS's build also builds NSPR from the same sandbox by default. You have to specify additional arguments to get it to use the system NSPR. That has actually failed for me sometimes.

By specifying `export SANDBOX=1`, JSS will pull NSS and NSPR from the sandbox, so you don't have to install it in order to use it.

tools/Dockerfiles/fedora_sandbox
cipherboy: We actually specify all of the dependencies in the previous `dnf` statement, since this is cloned from the `fedora_rawhide` container image (which doesn't build an RPM and only builds and runs the JSS test suite regularly).

Since we already maintain dependency lists outside of the spec file (for Debian and Ubuntu also), maintaining them here isn't much additional work.

And no `wget` is necessary, since the current tree's build is linked into the container; I could point to the path of the actual spec file. Perhaps I should just remove the `jss` here. 

Note that NSS and NSPR's deps (via `builddep`) are incorrect; some of the previous `dnf` statements bring in NSS / NSPR sandbox build deps that aren't in the spec file.

tools/Dockerfiles/fedora_sandbox
cipherboy: To clarify: a sandbox is a directory structure like:

```
some_dir/
    nspr/
    nss/
    jss/
```

After building nss, you'll also get a `some_dir/dist/` folder. When building JSS with `SANDBOX=1` exported, it'll look in the parent directory for a dist folder and use NSS/NSPR from the built distribution.

cipherboy:
I've dropped it from the required section.

PR #262, opened 2019-09-19
Finish creating SymmetricKeys from byte arrays

cipherboy:
In the course of #259, I wanted to finish creating a `PK11SymKey` from a byte array. In doing so, I realized that, when we specify a [`SecretKeySpec`](https://docs.oracle.com/javase/7/docs/api/javax/crypto/spec/SecretKeySpec.html), we do the import correctly already in our [`SecretKeyFactory`](https://docs.oracle.com/javase/7/docs/api/javax/crypto/SecretKeyFactory.html) implementation, [`JSSSecretKeyFactorySpi`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/provider/javax/crypto/JSSSecretKeyFactorySpi.java). However, what is missing is support for a few newer HMAC variants.

This PR does the following things:

 - Finishes adding support for SHA2 variants. This includes adding missing class stubs using the right constants.
 - Allows mapping for `SymmetricKey.Usage` to `PKCS11Constants` values. This is currently unused, but could be used in the future. 
 - Adds support for multiple names in `SymmetricKey.Types`... mostly for HMAC. Otherwise, you'd pass one name to `SecretKeyFactory` and another to `SecretKeySpec`.
 - Updates the tests to use the simplified import structure.

### DIFF
--- a/org/mozilla/jss/crypto/SymmetricKey.java
+++ b/org/mozilla/jss/crypto/SymmetricKey.java
@@ -13,6 +13,9 @@ public interface SymmetricKey extends javax.crypto.SecretKey {
     public static final Type RC4 = Type.RC4;
     public static final Type RC2 = Type.RC2;
     public static final Type SHA1_HMAC = Type.SHA1_HMAC;
+    public static final Type SHA256_HMAC = Type.SHA256_HMAC;
+    public static final Type SHA384_HMAC = Type.SHA384_HMAC;
+    public static final Type SHA512_HMAC = Type.SHA512_HMAC;
     public static final Type AES = Type.AES;
 
     public Type getType();
--- a/org/mozilla/jss/crypto/SymmetricKey.java
+++ b/org/mozilla/jss/crypto/SymmetricKey.java
@@ -6,6 +6,8 @@ package org.mozilla.jss.crypto;
 import java.security.NoSuchAlgorithmException;
 import java.util.Hashtable;
 
+import org.mozilla.jss.pkcs11.PKCS11Constants;
+
 public interface SymmetricKey extends javax.crypto.SecretKey {
 
     public static final Type DES = Type.DES;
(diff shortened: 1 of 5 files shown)


edewata:
Generally looks fine, but I'd defer to @jmagne.

org/mozilla/jss/crypto/SymmetricKey.java
```diff
@@ -48,40 +53,43 @@ public NotExtractableException(String mesg) {
...
+        public static final Type SHA384_HMAC = new Type(new String[]{ "SHA384_HMAC", "SHA384-HMAC", "SHA384HMAC", "HMAC_SHA384", "HMAC-SHA384", "HMACSHA384" },
             KeyGenAlgorithm.SHA384_HMAC);
-        public static final Type SHA512_HMAC = new Type("SHA512_HMAC",
+        public static final Type SHA512_HMAC = new Type(new String[]{ "SHA512_HMAC", "SHA512-HMAC", "SHA512HMAC", "HMAC_SHA512", "HMAC-SHA512", "HMACSHA512" },
             KeyGenAlgorithm.SHA512_HMAC);
-        public static final Type PBA_SHA1_HMAC = new Type("PBA_SHA1_HMAC",
+        public static final Type PBA_SHA1_HMAC = new Type(new String[]{ "PBA_SHA1_HMAC" },
             KeyGenAlgorithm.PBA_SHA1_HMAC);
-        public static final Type AES = new Type("AES", KeyGenAlgorithm.AES);
+        public static final Type AES = new Type(new String[]{ "AES" }, KeyGenAlgorithm.AES);
```
jmagne: Would there be any use to return all the aliases separated by a delim of some kind?

cipherboy: Ah I should add a `getAliases` (plural) to return the entire array, good point. Then the caller can (hope to) determine whether or not to show one or show all, and how they want to show all the aliases if they wish to display them all. (They can choose their own deliminator).

PR #265, opened 2019-09-25
Add SSL Storage class, SSLFDProxy

cipherboy:
Off of a SSL-enabled `PRFileDesc` created by NSS's `SSL_ImportFD`, we might
want to store various information which would be helpful (and, necessary
to free at the end):

 - Any `TrustManager`s we're using on this connection,
 - Whether or not the handshake has finished,
 - The client certificate, or
 - Additional parameters required for NSS callbacks.

Some of these will be Java-backed values, in which case they should be
added directly to `SSLFDProxy`. Others will be C-backed values, in which
case they should be freed once the SSL Socket is closed.

### DIFF
--- a/org/mozilla/jss/nss/PRFDProxy.c
+++ b/org/mozilla/jss/nss/PRFDProxy.c
@@ -5,8 +5,8 @@
 #include "jssutil.h"
 #include "PRFDProxy.h"
 
-jobject
-JSS_PR_wrapPRFDProxy(JNIEnv *env, PRFileDesc **fd)
+static jobject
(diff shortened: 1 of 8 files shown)


cipherboy:
@edewata Should I update members of the `nss.SSL` class to accept `SSLFDProxy` instances instead of `PRFDProxy`s? This requires a bit more work and would break a lot of existing typing in the tests and other open PRs. Thoughts?

cipherboy:
I've decided to update everything to use SSLFDProxy by default.

cipherboy:
Relatively sane and safe PR. I'm going ahead and merging as it is required for #248.

PR #268, opened 2019-09-26
Validate X509Certificates from CryptoManager

cipherboy:
We extend the existing VerifyCertificate logic to include support for
validating a certificate by reference (e.g., an instance of
X509Certificate) instead of purely by nickname.

We also expose the value of ocspPolicy from the CryptoManager (instead
of the ordinal, which we'd have to convert back to an enum for most
calls).

### DIFF
--- a/org/mozilla/jss/CryptoManager.java
+++ b/org/mozilla/jss/CryptoManager.java
@@ -1266,6 +1266,17 @@ public final class CryptoManager implements TokenSupplier
     }
 
     /**
+     * Gets the current OCSP Policy.
+     *
+     * @see getOCSPPolicy()
+     *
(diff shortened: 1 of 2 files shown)


org/mozilla/jss/CryptoManager.java
```diff
@@ -1224,11 +1249,21 @@ private native boolean verifyCertTempNative(byte[] certPackage,
      * have.
      * @return - The current ocsp policy in effect.
      */
-
     public static synchronized int getOCSPPolicy() {
         return ocspPolicy.ordinal();
     }
 
+    /**
```
cipherboy: These should be pulled out into a separate commit. I guess it is fine to leave them in this PR.

PR #269, opened 2019-09-26
Capture PRErrorCode <-> Exception mappings

cipherboy:
Often we'll need to take a `PRErrorCode` and raise it as a Java Exception
(via the JNI interface) of the proper type. Other times, we'll need to
take a raised Java Exception and return it as the correct
(`SECStatus`, `PRErrorCode`) pair.

We introduce `JSS_ExceptionToSECStatus` and `JSS_SECStatusToException` to
handle this. The latter has a *Message form, which takes an error
message to raise with the exception.

### DIFF
--- a/org/mozilla/jss/util/jssutil.c
+++ b/org/mozilla/jss/util/jssutil.c
@@ -16,6 +16,8 @@
 #include "certt.h"
 #include "pk11util.h"
 
+#include "secerr.h"
+
 
 /***********************************************************************
 **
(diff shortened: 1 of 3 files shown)


cipherboy:
@edewata I'd appreciate a review and if you have some better way of maintaining this mapping in C. The goal is to be able to take whatever exception a TrustManager raises and share it with NSS when possible. Some places in the SSLEngine might end up using the reverse mapping, so that's why I've added both.

org/mozilla/jss/util/jssutil.c
```diff
@@ -750,6 +752,106 @@ void JSS_DerefJString(JNIEnv *env, jstring str, const char *ref) {
...
+    }
+
+    // CERTIFICATE_REVOKED_EXCEPTION <-> SEC_ERROR_REVOKED_CERTIFICATE
+    clazz = (*env)->FindClass(env, CERTIFICATE_REVOKED_EXCEPTION);
+    if (clazz != NULL && (*env)->IsInstanceOf(env, except, clazz)) {
+        PORT_SetError(SEC_ERROR_REVOKED_CERTIFICATE);
+        return SECFailure;
+    }
+
+    return SECFailure;
```
edewata: Should we call PORT_SetError() with a generic error code in case the exception doesn't match anything?

org/mozilla/jss/util/jssutil.c
```diff
@@ -750,6 +752,106 @@ void JSS_DerefJString(JNIEnv *env, jstring str, const char *ref) {
...
+
+void JSS_SECStatusToExceptionMessage(JNIEnv *env, SECStatus result, PRErrorCode code, const char *message) {
+    if (result == SECSuccess) {
+        // We ignore PRErrorCode here. However, while we could clear an
+        // exception if it occurred, we don't; that's the caller's choice to
+        // make.
+        return;
+    }
+
+    if (result == SECFailure) {
```
edewata: Can the `result` have any value other than `SECSuccess` and `SECFailure`? If so should we ignore it, or throw a generic exception?

org/mozilla/jss/util/jssutil.c
```diff
@@ -750,6 +752,106 @@ void JSS_DerefJString(JNIEnv *env, jstring str, const char *ref) {
...
+            case SEC_ERROR_EXPIRED_CERTIFICATE:
+                JSS_throwMsgPrErrArg(env, CERTIFICATE_EXPIRED_EXCEPTION, message, code);
+                break;
+            case SEC_ERROR_BAD_DER:
+                JSS_throwMsgPrErrArg(env, CERTIFICATE_PARSING_EXCEPTION, message, code);
+                break;
+            case SEC_ERROR_REVOKED_CERTIFICATE:
+                JSS_throwMsgPrErrArg(env, CERTIFICATE_REVOKED_EXCEPTION, message, code);
+                break;
+        }
```
edewata: Do we need a default case and throw a generic exception?

org/mozilla/jss/util/jssutil.c
cipherboy: I've thought about that, but I don't know a good generic error code. I suppose you could throw `SEC_ERROR_BASE`, but I'm not sure what'd happen if you did.

The remaining candidates I see are:

 - `SEC_ERROR_LIBRARY_FAILURE`
 - `SEC_ERROR_BAD_DATA`
 - `SEC_ERROR_INVALID_ARGS`

In most cases, outside of what gets eventually shown to the user, I don't think NSS will interpret the error code...

org/mozilla/jss/util/jssutil.c
cipherboy: There's `SECWouldBlock`, but that's rarely used and I'm not sure how it interacts with `PORT_SetError()`. Do you know a good exception to return for that?

edewata: Unfortunately.. 乁( ⁰͡  Ĺ̯ ⁰͡ ) ㄏ

org/mozilla/jss/util/jssutil.c
edewata: OK, I guess we probably should just leave it for now, but could you add a comment in the code to record this issue so maybe someone can fix the it properly in the future? I'll ACK this PR. Thanks!

cipherboy: I realized that NSPR might have the answer. My suggestion would be `PR_UNKNOWN_ERROR` actually.

org/mozilla/jss/util/jssutil.c
cipherboy: Added.

org/mozilla/jss/util/jssutil.c
cipherboy: Okie dokie, let's leave this one alone then. Other two comments are addressed. :)

PR #271, opened 2019-09-27
Support using SymmetricKey in JSSMacSpi

cipherboy:
`JSSMacSpi` differs from `JSSCipherSpi` in that it requires the caller to
pass a `SecretKeyFacade` instance (which merely wraps a `SymmetricKey`) and
are typically created by `SecretKeyFactory`. However, the caller might
have an existing `SymmetricKey` and wish to use that with `JSSMacSpi`; check
for that case and handle it gracefully. Unlike `JSSCipherSpi`, we don't
use the `SecretKeyFactory` to clone the underlying key.


cipherboy:
@jmagne This is in response to your comment on #259. I think it is perfectly reasonable to use a `SymmetricKey` here, so I think this'll "just work".

PR #272, opened 2019-09-27
SSL Client Certificates

cipherboy:
This PR adds support for SSL client certificates in `org.mozilla.jss.nss`, which will eventually enable their use in the SSLEngine (#150). In particular, we:

 - Add a new field to `SSLFDProxy`, `clientCert`,
 - Add a setter,
 - Add a new callback, `JSSL_SSLFDCertSelectionCallback` which returns the single cert we've already selected for use. 

Later enhancements can add other certificate selection mechanisms, but this should interact well with our current SSLEngine design.

### DIFF
--- a/org/mozilla/jss/nss/SSL.c
+++ b/org/mozilla/jss/nss/SSL.c
@@ -12,6 +12,7 @@
 #include "jssutil.h"
 #include "pk11util.h"
 #include "PRFDProxy.h"
+#include "SSLFDProxy.h"
 #include "SSLVersionRange.h"
 
 #include "_jni/org_mozilla_jss_nss_SSL.h"
@@ -408,6 +409,34 @@ Java_org_mozilla_jss_nss_SSL_PeerCertificateChain(JNIEnv *env, jclass clazz,
     }
 
     return JSS_PK11_wrapCertChain(env, &chain);
+}
+
+JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_nss_SSL_AttachClientCertCallback(JNIEnv *env, jclass clazz,
+    jobject fd)
+{
+    PRFileDesc *real_fd = NULL;
+    CERTCertificate *real_cert = NULL;
+
+    PR_ASSERT(env != NULL && fd != NULL);
+
+    if (JSS_PR_getPRFileDesc(env, fd, &real_fd) != PR_SUCCESS) {
+        return SECFailure;
+    }
+
+    if (JSS_NSS_getSSLClientCert(env, fd, &real_cert) != PR_SUCCESS) {
+        return SECFailure;
+    }
+
+    /* When the returned cert is empty and no error occurred, there was no
+     * client certificate specified via SSLFD.SetClientCert(...). Don't add
+     * the certificate selection callback handler. */
+    if (real_cert == NULL) {
+        return SECSuccess;
+    }
+
+    return SSL_GetClientAuthDataHook(real_fd, JSSL_SSLFDCertSelectionCallback,
+                                     (void *)real_cert);
 }
 
 JNIEXPORT jint JNICALL
--- a/org/mozilla/jss/nss/SSLFDProxy.c
+++ b/org/mozilla/jss/nss/SSLFDProxy.c
@@ -1,0 +1,86 @@
+#include <nspr.h>
+#include <nss.h>
+#include <ssl.h>
+#include <pk11pub.h>
+#include <jni.h>
+
+#include "java_ids.h"
+#include "jssutil.h"
+#include "pk11util.h"
+#include "SSLFDProxy.h"
+
+PRStatus
+JSS_NSS_getSSLClientCert(JNIEnv *env, jobject sslfd_proxy, CERTCertificate **cert)
+{
+    jclass sslfdProxyClass;
+    jfieldID certField;
+    jobject certProxy;
+
+    PR_ASSERT(env != NULL && sslfd_proxy != NULL && cert != NULL);
+
+    /* Resolve the clientCert field on a SSLFDProxy object. */
+    sslfdProxyClass = (*env)->GetObjectClass(env, sslfd_proxy);
+    if (sslfdProxyClass == NULL) {
+        return PR_FAILURE;
+    }
+
+    certField = (*env)->GetFieldID(env, sslfdProxyClass,
+                                   SSLFD_PROXY_CLIENT_CERT_FIELD,
+                                   SSLFD_PROXY_CLIENT_CERT_SIG);
+    if (certField == NULL) {
+        return PR_FAILURE;
+    }
+
+    certProxy = (*env)->GetObjectField(env, sslfd_proxy, certField);
+
+    if (certProxy == NULL) {
+        *cert = NULL;
+        return PR_SUCCESS;
+    }
+
+    /* Get the underlying CERTCertificate pointer from the clientCert
+     * (of type PK11Cert) object. */
+    return JSS_PK11_getCertPtr(env, certProxy, cert);
+}
+
+SECStatus
+JSSL_SSLFDCertSelectionCallback(void *arg,
+                                PRFileDesc *fd,
+                                CERTDistNames *caNames,
+                                CERTCertificate **pRetCert,
+                                SECKEYPrivateKey **pRetKey)
+{
+    CERTCertificate *cert = arg;
+    PK11SlotList *slotList;
+    PK11SlotListElement *slotElement;
+    SECKEYPrivateKey *privkey = NULL;
+
+    /* Certificate selection for client auth requires that we pass both the
+     * certificate (in *pRetCert) and its key (in *pRetKey). We iterate over
+     * all slots that the certificate is in, looking for a private key in
+     * any of them. Once found, we return the private key. */
+    slotList = PK11_GetAllSlotsForCert(cert, NULL /* unused arg */);
+    if (slotList == NULL) {
+        return SECFailure;
+    }
+
+    for (slotElement = slotList->head; slotElement; slotElement = slotElement->next) {
+        privkey = PK11_FindPrivateKeyFromCert(slotElement->slot, cert, NULL /* pinarg */);
+        if (privkey != NULL) {
+            break;
+        }
+    }
+
+    /* Always free the slot list. */
+    PK11_FreeSlotList(slotList);
+
+    /* If the certificate isn't found, return SECFailure. */
+    if (privkey == NULL) {
+        return SECFailure;
+    }
+
+    *pRetCert = cert;
+    *pRetKey = privkey;
+    return SECSuccess;
+}
+
--- a/lib/jss.map
+++ b/lib/jss.map
@@ -398,6 +398,7 @@ JSS_4.6.2 {
     global:
 Java_org_mozilla_jss_pkcs11_PK11PrivKey_getPublicKey;
 Java_org_mozilla_jss_CryptoManager_importDERCertNative;
+Java_org_mozilla_jss_nss_SSL_AttachClientCertCallback;
     local:
         *;
 };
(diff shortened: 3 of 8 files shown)


org/mozilla/jss/nss/SSL.c
```diff
@@ -410,6 +411,34 @@ Java_org_mozilla_jss_nss_SSL_PeerCertificateChain(JNIEnv *env, jclass clazz,
...
+
+    if (JSS_PR_getPRFileDesc(env, fd, &real_fd) != PR_SUCCESS) {
+        return SECFailure;
+    }
+
+    if (JSS_NSS_getSSLClientCert(env, fd, &real_cert) != PR_SUCCESS) {
+        return SECFailure;
+    }
+
+    /* When the returned cert is empty, but no error occurred when fetching
```
cipherboy: Clean up comment wording.

org/mozilla/jss/nss/SSL.c
```diff
@@ -410,6 +411,34 @@ Java_org_mozilla_jss_nss_SSL_PeerCertificateChain(JNIEnv *env, jclass clazz,
...
+
+    if (JSS_NSS_getSSLClientCert(env, fd, &real_cert) != PR_SUCCESS) {
+        return SECFailure;
+    }
+
+    /* When the returned cert is empty and no error occurred, there was no
+     * client certificate specified via SSLFD.SetClientCert(...). Don't add
+     * the certificate selection callback handler. */
+    if (real_cert == NULL) {
+        return SECSuccess;
```
edewata: Does this mean the SSL handshake will proceed without client auth? In this case the caller specifically wants to do a client auth, so shouldn't we return a `SECFailure` if for whatever reason the `real_cert` is `NULL`?

org/mozilla/jss/nss/SSLFDProxy.c
```diff
@@ -0,0 +1,86 @@
...
+    PK11_FreeSlotList(slotList);
+
+    /* If the certificate isn't found, return SECFailure. */
+    if (privkey == NULL) {
+        return SECFailure;
+    }
+
+    *pRetCert = cert;
+    *pRetKey = privkey;
+    return SECSuccess;
```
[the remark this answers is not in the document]
cipherboy: We do that. See [lines 77 - 80](https://github.com/cipherboy/jss/blob/ssl-client-cert/org/mozilla/jss/nss/SSLFDProxy.c#L77-L80) above: 

```java
    /* If the certificate isn't found, return SECFailure. */
    if (privkey == NULL) {
        return SECFailure;
    }
```

org/mozilla/jss/nss/SSL.c
cipherboy: > Does this mean the SSL handshake will proceed without client auth? 

Yes, or rather, the handshake will fail if the server requires client auth because none would be provided from the client. 

> In this case the caller specifically wants to do a client auth, so shouldn't we return a SECFailure if for whatever reason the real_cert is NULL?

There's two parts to enabling client auth: selecting a certificate and adding the callback handler. Presently, there's only one way of doing each:
 
 1. Choose a certificate. A single static certificate is selected on the SSLFDProxy object.
 2. Attach a callback handler. Since there's a single static certificate, we really can have only one callback implementation: return the cert and its associated private key.

NSS (in general) allows the callback to make a more informed decision if it wishes: it can process say, the list of CAs who have been indicated as trusted by the server. Our existing [client certificate selection logic](https://github.com/cipherboy/jss/blob/ssl-client-cert/org/mozilla/jss/ssl/callbacks.c#L121-L294) lets the SSL user choose which particular certificate to use based on which are trusted and some callback which checks against a Java function.

If we implement something like that, except better suited for a KeyManager, in the future we'd then have the same two-step client auth process:

 1. Set up a certificate selection mechanism on the SSLFDProxy. 
    - Specify a single certificate,
    - Attach a KeyManager,
    - ... some other mechanism ... 
 2. Attach the correct callback.

2 could remain a single function call (`SSL.EnableClientAuthentication` -- perhaps we could rename to `SSL.AttachClientCertCallback`) that based on the information on SSLFDProxy, detects the correct one to use. Then one could envision, if step 1 isn't performed and there's no certificate, no callback is added. So, step 2 could *always* be called (doing nothing instead of erring in the case when step 1 isn't performed, and only erring when attaching the callback failed), and step 1 could be called *only* when a cert selection mechanism is desired (e.g., when the client has certificates it might want to use or when the client knows ahead of time that it wants to do client authentication).

So, my suggestion would be to rename EnableClientAuthentication -> AttachClientCertCallback.

Does the above make sense?

cipherboy: What this simplifies is the logic a little. Rather than having something like:

```java
void setKeyManager(KeyManager km) {
  if (km != null) {
    ssl_fd.setKeyManager(km);
    SSL.AttachClientCertCallback(ssl_fd);
  }
}
```

You'd have:

```java
void setKeyManager(km) {
  ssl_fd.setKeyManager(km);
  SSL.AttachClientCertCallback(ssl_fd);
}
```

(I think setKeyManager always gets called from ParseParams(...) but I could be wrong; I haven't actually looked at the Java source code for that). 

My intent with the `org.mozilla.jss.nss` namespace is to provide a mechanism for others to write their own SSL-enabled dodad (whether that's a SSLEngine, a SSLSocket, or whatever the next iteration of this is in Java 50) in the future, without needing to modify as much of JSS. All of the core bindings should be there, and if, for whatever reason, our JSSEngine doesn't work for their use case, it should be easier to create their own. My 2c.

org/mozilla/jss/nss/SSLFDProxy.c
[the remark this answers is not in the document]
edewata: Huh.. somehow I missed that :)

org/mozilla/jss/nss/SSL.c
edewata: > > Does this mean the SSL handshake will proceed without client auth?
> 
> Yes, or rather, the handshake will fail if the server requires client auth because none would be provided from the client.

Let's say the client auth is optional (that's how we configure PKI server), would there be any difference between:

1. someone trying to authenticate with a client cert, but for some reason it didn't get set properly, and the client still proceeds without the client cert
2. someone intentionally doesn't want to use a client cert (i.e. anonymous user)

I think in scenario 1 the client should fail (before even starting the handshake) because it cannot do what the user wants, instead of leaving the server to reject the request or to treat the user as anonymous user. This probably unlikely to happen or has only minor consequence so I'll ACK this. I'll let you decide if you want to push this first and address it further later.

About attaching the callback, this is probably a separate issue. I don't know all the details, but let's say in PKI server we need to create multiple SSL connections to different backends (e.g. DS, MongoDB), and one needs a client cert, but the other needs a different client cert, or doesn't require at all, will JSS/NSS support these scenarios at the same time in the same JVM?

cipherboy: > would there be any difference between...

The difference would have to be at a higher level, above the SSLEngine level. In particular, the SSLEngine doesn't particularly care about that. So here's the difficulty:

 - `SSLEngine` doesn't [have a public constructor](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html#SSLEngine--). Well, they're both protected. Any SSLEngine gets constructed from a [`SSLContext`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLContext.html) in its `createSSLEngine()` method.
    - This means that `SSLEngine` initialization isn't the right time to check if the intent of the user matches what they're getting.
    - A `SSLContext` is a Java-provider thing, which makes it really restrictive because you only get to implement [`SSLContextSpi`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLContextSpi.html). You get called with [`engineInit(...)`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLContextSpi.html#engineInit-javax.net.ssl.KeyManager:A-javax.net.ssl.TrustManager:A-java.security.SecureRandom-) and that's about it. So, if you want to enforce a policy, you have to do it on `SSLContext` creation time, i.e., via the `JSSProvider` and subclasses. 
    - In short, this means that you could have a `SSLContext$RequireClientAuth` policy you could instantiate whenever you want to force client auth on the client side. 
 - [`SSLEngine.setNeedClientAuth()`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html#setNeedClientAuth-boolean-) is for the server only. 
 - [`SSLEngine.setWantClientAuth()`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html#setWantClientAuth-boolean-) is also server-only. 
 - We don't necessarily control the KeyManager. We mostly [enforce](https://github.com/cipherboy/jss/blob/ssl-engine/org/mozilla/jss/provider/javax/net/JSSContext.java#L26-L42) that it is [your KeyManager](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/provider/javax/crypto/JSSKeyManager.html). In any rate, [`chooseClientAlias`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/provider/javax/crypto/JSSKeyManager.java#L56) and [`chooseServerAlias`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/provider/javax/crypto/JSSKeyManager.java#L69) return `null`.
    - What I've been relying on currently for my JSSEngine implementation is a new member, [`getCertificate(alias)`](https://github.com/cipherboy/jss/blob/ssl-engine/org/mozilla/jss/provider/javax/crypto/JSSKeyManager.java#L72-L80), and [passing](https://github.com/cipherboy/jss/blob/ssl-engine/org/mozilla/jss/ssl/javax/JSSEngine.java#L514-L520) it via [`JSSParameters`](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/ssl/javax/JSSParameters.html#setAlias-java.lang.String-).
    - We could however, extend `JSSKeyManager` to make it a little more useful and set a default alias so that `choose{Client,Server}Alias(...)` work as expected. The issue with that though is we don't construct them directly, they come from a [`KeyManagerFactory`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/KeyManagerFactory.html) instance... which doesn't understand aliases, and we can only implement the [`KeyManagerFactorySpi`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/KeyManagerFactorySpi.html)  interface and so can't control the `KeyManagerFactory` directly either. Yuck!
    - At any rate, if a bad alias is provided, it (`getCertificate(...)`) won't return `null` but will raise a `RuntimeException`. So the answer to 1 I guess is "yes" -- you provide an alias, it eventually gets to `JSSKeyManager` via `JSSParameters`, and a `RuntimeException` will occur if the cert doesn't exist under that nickname. So you really shouldn't get a `null` cert passed to `SetClientCert` unless you don't give a nickname.
    - The end result is, outside of JSSParameters (which requires modifying the least number of things), we don't have a good way of setting policy or querying keys even here. NSS does provide a requested Issuers parameters (in the form of `CERTCertificate *`'s), so if we find ourselves really wanting a better cert callback implementation, we can probably do so.

cipherboy: > About attaching the callback, this is probably a separate issue. I don't know all the details, but let's say in PKI server we need to create multiple SSL connections to different backends (e.g. DS, MongoDB), and one needs a client cert, but the other needs a different client cert, or doesn't require at all, will JSS/NSS support these scenarios at the same time in the same JVM?

Callback attachment is per ssl-enabled `PRFileDesc *`, so yes, you can support all these scenarios, and more (if you write other callbacks and choose which one to enable)!

edewata: This is getting more complicated than I expected :) but it sounds like the issue should not happen. Feel free to push. Thanks for the info!

PR #273, opened 2019-09-27
Fix callback alert handling

cipherboy:
In 871231d20f6e6956134c670d57087f121ffe0563 I incorrectly handled an
unused variable warning, where rc is unused when compiling outside of
debug mode. As part of this fix, I incorrectly returned based on the
same values we were asserting -- for 3 of the 4 values. Instead, I
should've returned when these conditionals were negated.

Candidate to backport to `v4.5.x`. Not necessary to backport to `v4.4.x` to my knowledge. 

@geetikakay Do you mind checking the resulting build to see if this happens to fix the SSLAlert issue you were mentioning in a [previous BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1733651)?

### DIFF
--- a/org/mozilla/jss/ssl/callbacks.c
+++ b/org/mozilla/jss/ssl/callbacks.c
@@ -313,8 +313,8 @@ JSSL_AlertReceivedCallback(const PRFileDesc *fd, void *arg, const SSLAlert *aler
     PR_ASSERT(env != NULL);
 
     /* Fast return when assumptions are incorrect. */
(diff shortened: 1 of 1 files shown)


geetikakay:
@cipherboy   Can you share the copr build that we can take and test

cipherboy:
@geetikakay Hi!

I shared this in the Bugzilla. But the COPR is here: https://copr.fedorainfracloud.org/coprs/g/pki/master/

Thanks,
Alex

PR #275, opened 2019-09-30
Native proxy improvements

cipherboy:
I thought about this while I was implementing SSLAlert handling, so I figured I'd attempt it here.

We extend the default NativeProxy implementation in three directions:

 - Make it `AutoCloseable`,
 - Add a `clear()` method which supports clearing the value of the
   pointer without calling `releaseNativeResources(...)`, and
 - Save a stack trace when assertions are enabled.

By implementing `AutoCloseable`, we enable the common Java pattern of
try-with-resources:

    try (NativeProxy item = ...) {
        ...
    }

and have the item automatically be closed (in our case, `finalized(...)`)
at the end of its life cycle. We've added additional logic to prevent
double-free bugs.

Additionally, by adding `clear(...)`, we both add protection from
double-free bugs and enable resource cleanup from native handlers that
removes entries in `NativeProxy`'s internal tracking.

Lastly, we save stack traces when assertions are enabled in the JVM and 
JSS is built in `DEBUG` mode (with `DEBUG` defined).

We enable its use on Buffers and PRFDProxys. The latter is an interesting case since NativeProxy still thinks these `PRFileDesc`'s leak on creation due to our use of `SSL.ImportFD(...)` (which subsumes the socket, meaning we don't have to close it, and it handles closing it for us internally). We could get around that by providing a single "Create PRFileDesc and import FD" method, but...

### DIFF
--- a/org/mozilla/jss/util/NativeProxy.java
+++ b/org/mozilla/jss/util/NativeProxy.java
@@ -4,9 +4,13 @@
 
 package org.mozilla.jss.util;
 
-import java.util.Enumeration;
-import java.util.Hashtable;
-import java.util.Random;
+import java.util.HashSet;
+
+import java.lang.AutoCloseable;
+import java.lang.Thread;
+import java.util.Arrays;
+
+import org.mozilla.jss.CryptoManager;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -20,9 +24,10 @@ import org.slf4j.LoggerFactory;
  * @author nicolson
  * @version $Revision$ $Date$
  */
-public abstract class NativeProxy
+public abstract class NativeProxy implements AutoCloseable
 {
     public static Logger logger = LoggerFactory.getLogger(NativeProxy.class);
+    private static final boolean saveStacktraces = assertsEnabled() && CryptoManager.JSS_DEBUG;
 
     /**
      * Create a NativeProxy from a byte array representing a C pointer.
@@ -35,8 +40,12 @@ public abstract class NativeProxy
      */
     public NativeProxy(byte[] pointer) {
 		assert(pointer!=null);
-        registryIndex = register();
+        registry.add(this);
         mPointer = pointer;
+
+        if (saveStacktraces) {
+            mTrace = Arrays.toString(Thread.currentThread().getStackTrace());
+        }
     }
 
     /**
@@ -68,8 +77,10 @@ public abstract class NativeProxy
      * Subclasses of NativeProxy must define this method to clean up
      * data structures in C code that are referenced by this proxy.
      * releaseNativeResources() will usually be implemented as a native method.
-     * <p>You don't call this method; NativeProxy.finalize() calls it for you.
-     * <p>You must declare a finalize() method which calls super.finalize().
+     * <p>You don't call this method; NativeProxy.finalize() or close() calls
+     * it for you.</p>
+     *
+     * If you free these resources explicitly, call clear(); instead.
      */
     protected abstract void releaseNativeResources();
 
@@ -88,18 +99,53 @@ public abstract class NativeProxy
      *      }
      * }
      *
-     * @deprecated finalize() in Object has been deprecated
+     * @deprecated finalize() in Object has been deprecated. Use close(...)
+     * from the AutoCloseable interface instead.
      */
     @Deprecated
     protected void finalize() throws Throwable {
-        unregister(registryIndex);
-        releaseNativeResources();
+        close();
+    }
+
+    /**
+     * Close this NativeProxy by releasing its native resources if they
+     * haven't otherwise been freed.
+     *
+     * See comment in finalize.
+     */
+    public final void close() throws Exception {
+        try {
+            if (registry.remove(this)) {
+                releaseNativeResources();
+            }
+        } finally {
+            mPointer = null;
+        }
+    }
+
+    /**
+     * Call clear(...) to clear the value of the pointer, setting it to null.
+     *
+     * This should be used when the pointer has been freed by another means.
+     * Similar to finalize(...) or close(...), except that it doesn't call
+     * releaseNativeResources(...).
+     *
+     * See also: JSS_clearPtrFromProxy(...) in jssutil.h
+     */
+    public final void clear() {
+        this.mPointer = null;
+        registry.remove(this);
     }
 
     /**
      * Byte array containing native pointer bytes.
      */
     private byte mPointer[];
+
+    /**
+     * String containing backtrace of pointer generation.
+     */
+    private String mTrace;
 
     /**
      * <p><b>Native Proxy Registry</b>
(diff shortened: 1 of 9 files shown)


org/mozilla/jss/util/NativeProxy.java
```diff
@@ -92,15 +103,56 @@ public boolean equals(Object obj) {
...
+
+    /**
+     * Call this to clear the value of the pointer, setting it to null.
+     *
+     * This should be used when the pointer has been free'd by another means.
+     * Similar to finalize(...) except that it doesn't call
+     * releaseNativeResources(...).
+     */
+    public final void clear() {
+        this.mPointer = null;
```
edewata: The `this` is not necessary.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -114,7 +166,7 @@ protected void finalize() throws Throwable {
      * releaseNativeResources() gets called.
      */
     private long registryIndex;
-    static Hashtable<Long, Long> registry;
+    static Hashtable<Long, NativeProxy> registry;
```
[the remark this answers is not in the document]
cipherboy: Yeah this should be fine. I don't think anyone is really using our registry's public API at all, so I've simplified it a bit.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -92,15 +103,56 @@ public boolean equals(Object obj) {
...
+                unregister(registryIndex);
+                registryIndex = -1;
+
+                releaseNativeResources();
+            }
+        } catch (Throwable t) {
+            throw t;
+        } finally {
+            mPointer = null;
+        }
```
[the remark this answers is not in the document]
cipherboy: I don't think `finalize()` is going anywhere any time soon. It is still in the [JDK 13 docs](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/lang/Object.html#finalize()). 

But I have done this.

cipherboy:
Thanks @edewata! I think this will be a cleaner `NativeProxy` interface.

The one issue with using a `HashSet` over `NativeProxy`'s directly is that multiple `NativeProxy` instancess might share a single pointer value (because there's no way to create a `NativeProxy` without registering it, even if separate `NativeProxy`s might "own" the same pointer).

In the future, I think we can add a constructor that doesn't register the instance, but that's for another time.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -88,19 +99,54 @@ public boolean equals(Object obj) {
...
+     *
+     * See comment in finalize.
+     */
+    public final void close() throws Exception {
+        try {
+            if (unregister(this)) {
+                releaseNativeResources();
+            }
+        } catch (Exception e) {
+            throw e;
```
edewata: I think we can drop the catch clause since it's simply rethrowing the exception.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -145,28 +181,20 @@ private synchronized static long register() {
      * @param index The index of this object in the registry, as returned
      * from the previous call to register().
```
edewata: The `@param index` is no longer needed.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -145,28 +181,20 @@ private synchronized static long register() {
...
-    }
-
-    /**
-     * @return A list of the indices in the registry. Each element is a Long.
-     * @see NativeProxy#getRegistryIndex
-     */
-    public synchronized static Enumeration<Long> getRegistryIndices() {
-        return registry.keys();
+    private synchronized static boolean unregister(NativeProxy instance) {
+        return registry.remove(instance);
```
edewata: We probably can make it a no-arg instance method and call:
```
return registry.remove(this);
```

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -109,34 +155,24 @@ protected void finalize() throws Throwable {
...
-        registry = new Hashtable<>();
-        indexGenerator = new Random();
+        registry = new HashSet<NativeProxy>();
     }
 
     /**
      * Register a NativeProxy instance.
      *
-     * @return The unique index of this object in the registry.
+     * @return The NativeProxy instance to track in the registry.
```
edewata: The `@return` can be removed.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -176,10 +204,16 @@ public long getRegistryIndex() {
      * is thrown.
      */
     public synchronized static void assertRegistryEmpty() {
```
edewata: Any idea who's calling this method? The method doesn't throw exception or return anything, so this is not really an assertion.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -68,8 +77,10 @@ public boolean equals(Object obj) {
      * Subclasses of NativeProxy must define this method to clean up
      * data structures in C code that are referenced by this proxy.
      * releaseNativeResources() will usually be implemented as a native method.
-     * <p>You don't call this method; NativeProxy.finalize() calls it for you.
-     * <p>You must declare a finalize() method which calls super.finalize().
+     * <p>You don't call this method; NativeProxy.finalize() calls it for you.</p>
+     * <p>You must declare a finalize() method which calls super.finalize().</p>
```
edewata: Should we change this doc to call close() instead? Existing code will still call finalize(), but new code should call close() or use AutoClosable.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -109,34 +155,24 @@ protected void finalize() throws Throwable {
...
      */
-    private long registryIndex;
-    static Hashtable<Long, Long> registry;
-    static Random indexGenerator;
+    static HashSet<NativeProxy> registry;
 
     static {
-        registry = new Hashtable<>();
-        indexGenerator = new Random();
+        registry = new HashSet<NativeProxy>();
```
edewata: This line can be merged into line 163 so the static block can be removed.

org/mozilla/jss/util/NativeProxy.java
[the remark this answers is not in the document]
edewata: Right, finalize() is probably not going to be removed by Java anytime soon, but existing code should be updated to use close() instead.

org/mozilla/jss/util/NativeProxy.java
cipherboy: Since they're private functions I've actually removed both. `HashSet` creation is much simpler than `Hashtable`.

org/mozilla/jss/util/NativeProxy.java
cipherboy: So it doesn't throw an assertion any more but it does log output. In master, [you're the one](https://github.com/dogtagpki/jss/commit/e0911cf5f30afba57afef522da8b7a63a46e0137) who last touched it to use `logger` instead of `Debug.trace(...)`.

Nothing that [isn't commented out](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/tests/SelfTest.java#L221) (which isn't run through JSSTests yet) calls it formally. I've been thinking about adding `assertRegistryEmpty(...)` to the test suite for SSLEngine stuff though. I've added it back in and called it a few times in the TestBufferPRFD; that's where the enhancements for Backtracing NativeProxy creation came from.

There's only [four](https://www.google.com/search?client=firefox-b-1-d&q=assertRegistryEmpty) results for it on Google.

I don't think it'll be missed if we do drop it (or better -- quit registering objects outside of "Debug" mode so we only take its small performance hit there), but FWIW, it is kinda useful for catching leaks.

org/mozilla/jss/util/NativeProxy.java
cipherboy: Is this really correct anywho? I'm kinda under the impression that it isn't strictly necessary. 

 - If you're not freeing / clearing / &c. anything else, I don't see the point in overriding finalize(); won't the child's freeNativeResources(...) be called since NativeProxy doesn't define one?
 - If you are freeing something else, then it is obvious you need to override close(...) / finalize(...).

Thoughts?

org/mozilla/jss/util/NativeProxy.java
edewata: Yeah, if we want to keep it we probably can attach it into the server shutdown code such that we get a report whether the resources are cleaned up properly. We can take care of this another time.

org/mozilla/jss/util/NativeProxy.java
edewata: As discussed on IRC we can drop the comment.

PR #279, opened 2019-10-02
Introduce GlobalRefProxy

cipherboy:
Java JNI has the notion of a global reference to an object, instead of
the traditional local reference created by the JVM when calling any
JNI'd function. Since all global references have the same type (`jobject`)
and release handler (`(*env)->DeleteGlobalRef(...)`), it makes sense to
track these as a single JNI type.

This ensures that these resources are released, allowing the referenced
object to be garbage collected.

### DIFF
--- a/lib/jss.map
+++ b/lib/jss.map
@@ -400,6 +400,7 @@ Java_org_mozilla_jss_pkcs11_PK11PrivKey_getPublicKey;
 Java_org_mozilla_jss_CryptoManager_importDERCertNative;
 Java_org_mozilla_jss_nss_SSL_AttachClientCertCallback;
 Java_org_mozilla_jss_CryptoManager_getJSSDebug;
+Java_org_mozilla_jss_util_GlobalRefProxy_releaseNativeResources;
     local:
         *;
 };
--- a/org/mozilla/jss/util/GlobalRefProxy.c
+++ b/org/mozilla/jss/util/GlobalRefProxy.c
@@ -1,0 +1,70 @@
+#include <nspr.h>
+#include <nss.h>
+#include <jni.h>
+
+#include "java_ids.h"
+#include "jssutil.h"
+
+jobject
+JSS_PR_wrapGlobalRef(JNIEnv *env, jobject *ref)
+{
+    jbyteArray pointer = NULL;
+    jclass proxyClass;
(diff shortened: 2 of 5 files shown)


GlobalRefProxy.c
```diff
@@ -0,0 +1,70 @@
...
+    if (constructor == NULL) {
+        ASSERT_OUTOFMEM(env);
+        goto finish;
+    }
+
+    /* Call the constructor. */
+    refObj = (*env)->NewObject(env, proxyClass, constructor, pointer);
+
+finish:
+    if (refObj == NULL && *ref != NULL) {
```
edewata: Is `*ref != NULL` still needed? There's an assertion for it in line 16.

cipherboy: "eh" -- `PR_ASSERT` behaves like `assert(...)` in C (which only triggers when `DEBUG` is defined; we don't define it in release modes). I think `PR_ASSERT` triggers when `DEBUG` or `PR_DEBUG` is defined (the latter we don't define anywhere). This is similar to `assert(...)` in Java, where it only gets triggered when `-ea` is passed to the running JVM. Otherwise, they'll be ignored.

Most everything (in NSS or JSS) handles a `NULL` free just fine. I'm not sure if `DeleteGlobalRef` (which comes from the JVM) does. It isn't clear from [the docs](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html) if it handles `NULL` safely.

cipherboy:
Fixed obvious path error made while splitting this into a separate commit/PR: 

    GlobalRefProxy.{c,h,java} -> org/mozilla/jss/utils/GlobalRefProxy{c,h,java}

GlobalRefProxy.c
edewata: Ah, OK. I'm wondering if we should continue to use PR_ASSERTs or convert them into real code. Regardless, feel free to push.

cipherboy: Yeah, at some point in the future, I'd migrate to `PR_ASSERT`'s (so we get a nice crash we can attach to GDB with when built in debug mode) and `if` statements with proper exceptions. Some of these we don't need to assert though (`env != NULL` is a really weird edge case).

PR #281, opened 2019-10-02
Support freeing SSLFDProxy resources

cipherboy:
`SSLFDProxies` might have additional resources associated with them
besides just the `PRFileDesc` pointer. This form of `PR.Close(...)` will
first close the underlying `PRFileDesc` and then call `fd.close()`
explicitly, to handle any other resources.

@edewata Does this generally look like a sane decision? SSL-wrapped `PRFileDesc`'s otherwise behave like  `PRFileDesc`'s, since they implement the `PRFileDesc` interface thingy; you'd mostly interact with their fd-facing portion with NSPR calls. Only the SSL-specific stuff (cipher types, callbacks, &c) are handled by NSS. We're looking to do something similar, except at a higher level. Since we can't give NSS data to hold (and have a callback process freeing that data when `PR_Close(...)` is called on the underlying socket), we have to attach it to `SSLFDProxy` and then have a mechanism for handling freeing it. That data is on the `SSLFDProxy` class and not on the underlying `PRFileDesc`, so there's no use-after-free bug. 

### DIFF
--- a/org/mozilla/jss/nss/PR.java
+++ b/org/mozilla/jss/nss/PR.java
@@ -73,6 +73,18 @@ public class PR {
     public static native int Close(PRFDProxy fd);
 
     /**
+     * Close an existing SSLFDProxy.
+     *
+     * See also: org.mozilla.jss.nss.PR.Close
+     *           org.mozilla.jss.nss.SSLFD.releaseNativeResources
+     */
+    public static int Close(SSLFDProxy fd) throws Exception {
+        int ret = PR.Close((PRFDProxy) fd);
+        fd.close();
+        return ret;
+    }
+
+    /**
      * Shutdown an existing PRFDProxy.
      * This is usually used with TCP modes.
      *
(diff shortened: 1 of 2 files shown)


org/mozilla/jss/nss/PR.java
```diff
@@ -72,6 +72,18 @@ public static native PRFDProxy NewBufferPRFD(BufferProxy read_buf,
...
 
+    /**
+     * Close an existing SSLFDProxy.
+     *
+     * See also: org.mozilla.jss.nss.PR.Close
+     *           org.mozilla.jss.nss.SSLFD.releaseNativeResources
+     */
+    public static int Close(SSLFDProxy fd) throws Exception {
+        int ret = PR.Close((PRFDProxy) fd);
+        fd.close();
```
edewata: Usually we clean up the subclass first then the super class. If we run `fd.close()` before `PR.Close(PRFDProxy)` would that cause any problem? I'll let you decide since I'm not familiar with NSS / NSPR API at this level. Regardless, the code looks fine to me, ACK.

cipherboy: The first close simply closes the socket and frees resources related to the socket in the NSPR layer. The second `close(...)` calls `releaseNativeResources(...)` from `NativeProxy`. If the order is swapped, the first call will succeed, but the second will fail, since there's no pointer to the file descriptor, which means that the NSPR resources won't get freed and thus leak.

PR #284, opened 2019-10-15
CVE-2019-14823: Fix root certificate validation

cipherboy:
When the Leaf and Chain OCSP checking policy is enabled in
CryptoManager, JSS will switch to alternative certificate verification
logic in `JSSL_DefaultCertAuthCallback`. In this method, the root
certificate was incorrectly trusted without being verified to exist in
the trust store.

This patch cleans up the logic in `JSSL_verifyCertPKIX` and makes it
more explicit in addition to fixing the error.

Fixes: [CVE-2019-14823](https://nvd.nist.gov/vuln/detail/CVE-2019-14823)

### DIFF
--- a/org/mozilla/jss/ssl/common.c
+++ b/org/mozilla/jss/ssl/common.c
@@ -904,7 +904,6 @@ finish:
 }
 
 /* Get the trusted anchor for pkix */
-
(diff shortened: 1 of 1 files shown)


cipherboy:
See #285 for a reproducer.

PR #285, opened 2019-10-15
Add optional test case against badssl.com

cipherboy:
badssl.com maintains a number of subdomains with valid and invalid TLS
configurations. A number of these test certificates which fail in
certain scenarios (revoked, expired, etc). Add a test runner which
validates SSLSocket's implementation against badssl.com.

### DIFF
--- a/org/mozilla/jss/tests/BadSSL.java
+++ b/org/mozilla/jss/tests/BadSSL.java
@@ -1,0 +1,218 @@
+package org.mozilla.jss.tests;
+
+import org.mozilla.jss.CryptoManager;
+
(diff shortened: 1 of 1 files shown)


cipherboy:
To execute this script: 

1. First build and run the existing test suite
2. Use `tools/common_roots.sh` to add a NSS DB with roots from the system trust store, or add pk11-kit-trust to an existing NSS DB.
3. cd `build` and run `./run_test.sh results/cadb` or `./run_test.sh results/cadb LEAF_AND_CERT`.

Allowed failures will have no output after printing the test name; test cases which passed will print "ok", non-allowed failures will otherwise say so and state why they failed.

PR #292, opened 2019-10-16
Switch to suggesting default-jdk

cipherboy:
With the advent of JDK 11+ being the default JDK in Debian testing,
we should stop suggesting that JDK 8 is the only supported JDK for JSS.
Since we are compatible with JDK 8+ and Debian provides a default-jdk
installation target, use that instead.

This change was suggested by @emaldona.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ This project has the following dependencies:
     - Minimum version: 3.30
     - A c and c++ compiler such as [gcc](ttps://gcc.gnu.org/)
     - [zlib](https://zlib.net/)
- - [OpenJDK 1.8.0](https://openjdk.java.net/)
+ - [OpenJDK 1.8.0 or newer](https://openjdk.java.net/)
  - [CMake](https://cmake.org/)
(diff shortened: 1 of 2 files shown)


emaldona:
Bad copy and paste, ignore the bottom part.

cipherboy:
So this warning is a red herring:
> CMake Warning at cmake/JSSConfig.cmake:349 (message):
>   Your NSS version doesn't support CMAC; some features of JSS won't work.

It is fine that your NSS version doesn't support CMAC. As I [mentioned in the release notes](https://github.com/dogtagpki/jss/releases/tag/v4.6.2), NSS won't support it until 3.47. That'll be released [on the 18th](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/NSS_3.47_release_notes).

I think the X509CertTest failure is the same one the `ubuntu_jdk8` image has. I don't think JDK11 introduces any new failures there, but I'll double check as time permits.

emaldona:
@cipherboy thanks for the clarification

PR #299, opened 2019-10-28
Fix NativeProxy reference tracker

cipherboy:
In eb5df01003d74b57473eacb84e538d31f5bb06ca, I introduced a bug by
setting `mPointer` after trying to add `NativeProxy` to the registry. In
most instances this won't matter, however, if another instance exists in
the `HashSet` with the same hash value, the equals comparator will be
used, triggering a NPE.


edewata:
Not sure what happened, but it fixed the issue I had with MongoDatabase connection. Thanks!

PR #301, opened 2019-10-31
Fix sandbox builds of JSS

cipherboy:
Previously, we wouldn't include `NSS_LIBRARIES` in `LD_LIBRARY_PATH`, in part because it wasn't always a directory -- sometimes it was a path to a the library.

This fixes sandbox builds when NSS isn't installed on the system, or when the sandboxed NSS otherwise rejects the system NSS's library versions.


cipherboy:
Sandbox build error unrelated (nss:master doesn't yet support FIPS mode TLS v1.3). Passes CI, so I'm going ahead and merging.

PR #306, opened 2019-12-06
NIST SP800-108 KBKDF

cipherboy:
Add NIST SP800-108 KBKDF PKCS#11 Support
    
This introduces initial support for NIST SP800-108 KBKDFs via PKCS#11.
Support for all three NIST KBKDFs are included: counter mode, feedback
mode, and double pipeline mode. Additionally, NSS includes support for
data based derivations: variants of the KDFs which derive data objects
instead of keys. This allows them to be used for SCP03 challenges.
    
This builds upon previous commits which add support for the underlying
PKCS#11 constants and data parameters.

---

Task List:

 - [x] Implement PKCS#11 Attributes
 - [x] Finish implementing parameters
 - [x] Clean up PKCS#11 Constants diff
 - [x] Save parameters to/from native layer
 - [x] Generate keys
 - [x] Add general content tests
 - [x] Add additional derived keys tests
 - [x] Wait for NSS to merge KBKDF upstream.
 - [x] Documentation in #359.


org/mozilla/jss/crypto/KBKDF.c
```diff
@@ -0,0 +1,115 @@
...
+#include "pk11pub.h"
+#include "pkcs11n.h"
+#include "pkcs11t.h"
+#include "jni.h"
+
+#include "jssutil.h"
+#include "java_ids.h"
+#include "StaticVoidPointer.h"
+
+JNIEXPORT void JNICALL
```
cipherboy: @edewata here.

org/mozilla/jss/crypto/KBKDFByteArrayParam.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Arrays;
+
+import org.mozilla.jss.util.NativeEnclosure;
+
+/**
+ * A KBKDF Byte Array Parameter is a type of KBKDF Data Parameter that
+ * contains a single byte array that gets passed to the KBKDF to be inserted
+ * into the PRF input stream.
+ */
+public class KBKDFByteArrayParam extends KBKDFDataParameter {
```
cipherboy: @edewata here.

org/mozilla/jss/util/NativeEnclosure.java
```diff
@@ -0,0 +1,42 @@
+package org.mozilla.jss.util;
```
cipherboy: @edewata here.

org/mozilla/jss/pkcs11/PK11KeyGenerator.java
```diff
@@ -192,6 +195,27 @@ public SymmetricKey generate()
                     Password.wipeBytes(pwbytes);
                 }
             }
+        } else if (is_kbkdf) {
```
cipherboy: @edewata here.

org/mozilla/jss/pkcs11/PK11KeyGenerator.java
```diff
@@ -192,6 +195,27 @@ public SymmetricKey generate()
...
+            KBKDFParameterSpec kps = (KBKDFParameterSpec) parameters;
+
+            try {
+                kps.open();
+            } catch (Exception e) {
+                throw new RuntimeException(e.getMessage(), e);
+            }
+
+            SymmetricKey result = generateKBKDF(token, algorithm, kps.ptr,
```
cipherboy: AI @cipherboy: This needs to go into a `try { ... } finally { ... }` block.

org/mozilla/jss/util/NativeEnclosure.java
```diff
@@ -0,0 +1,42 @@
...
+ *      // ... do something with obj ...
+ * }
+ * </pre>
+ */
+public abstract class NativeEnclosure implements AutoCloseable {
+    public NativeProxy ptr;
+    public long ptr_size;
+
+    public final void open() throws Exception {
+        acquireNativeResources();
```
edewata: Any concern about concurrency? Do we need some kind of locking mechanism?

org/mozilla/jss/crypto/KBKDFByteArrayParam.java
```diff
@@ -0,0 +1,51 @@
...
+
+    public void setByteArray(byte[] data) throws IllegalArgumentException {
+        if (data == null) {
+            throw new IllegalArgumentException("Expected non-null byte array!");
+        }
+
+        this.data = Arrays.copyOf(data, data.length);
+    }
+
+    protected void acquireNativeResources() throws Exception {
```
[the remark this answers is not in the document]
cipherboy: Well, this is to implement the abstract methods from `NativeEnclosure`... The goal of `NativeEnclosure` is to wrap all access to the class at the native level into a single pointer lookup. It's essentially a `Serializable` interface (of sorts), except that it contains a member variables (`ptr` and `ptr_size`) and the requirement that `open(...)` get called before `ptr`/`ptr_size` are accessed, and `close(...)` gets called after all accesses are done. 

Also, you're not going to a real serialization format (XML / Base64 / JSON / ...), you're going to well, a pointer of some type.

So, IMO, having Native in the name 1) points the developer back to `NativeEnclosure` as the source of these methods and 2) is correct and won't become obsolete.

This is also mostly about defining where the boundary is between the Java and the C code is. That boundary can change over time, and could shift so that the abstract class `NativeEnclosure` isn't necessary, but until NSS is rewritten from scratch in Java (and JSS becomes obsolete), you'll still have to interface with C / "native" code somewhere. :-)

org/mozilla/jss/crypto/KBKDFParameterSpec.java
```diff
@@ -0,0 +1,115 @@
...
+
+import java.util.Arrays;
+import java.util.ArrayList;
+
+import org.mozilla.jss.util.NativeEnclosure;
+
+public abstract class KBKDFParameterSpec extends NativeEnclosure implements AlgorithmParameterSpec, KeySpec {
+    protected PKCS11Algorithm prf;
+    protected KBKDFDataParameter[] params;
+    protected KBKDFDerivedKey[] additional_keys;
```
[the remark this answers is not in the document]
cipherboy: Yeah... This is where I say that the native layer is tough. So what's happening is these are stored on the class itself, right? In C pseudo-code, we want to eventually do:

```c
CK_PRF_DATA_PARAM data_params = calloc(this.params.length, sizeof(CK_PRF_DATA_PARAM)); 
for (size_t index = 0; index < this.params.length; index++) {
    // Copy data from the KBKDFDataParameter class to the actual C struct. 
    data_params[index] = *this.params[index].ptr;
}

CK_SP800_108_KDF_PARAMS kdf_params = { this.params.length, data_params,  ... }

call_this_kdf(kdf_params);
``` 

(Where `this` refers to the `KBKDFParameterSpec` Java class instance and `this.params` is shorthand for actually accessing that array from the JNI layer. It isn't that simple sadly). 

By using `NativeEnclosure`, each class/parameter/key/... worries about how it gets serialized in a vacuum. It allocates memory for itself, it frees its own memory, it transfers whatever data it needs to between the allocated memory and the Java class.

We can then blindly copy the contents of / use those pointers from the JNI C layer (after accessing the value of the pointer) -- which makes that part really easy. 

Anyhow, using `ArrayList` makes the whole "accessing" part harder and requires jumping through more hoops to access the data. Definitely possible, just more work. I'd prefer to simplify the JNI layer as much as possible, hence a straight array.

--- 

And to answer your question, "eh", probably not. It'd probably be fine to only add a 'setParams' / `setAdditionalKeys` method and not an `addParameter` method. But I don't know how it'll get used in the end, so...

(We also don't need any additional keys so that'll largely go unused from our portion -- its just parameters that we care about for SCP03).

org/mozilla/jss/util/NativeEnclosure.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Deprecated
+    protected void finalize() throws Throwable {
+        close();
```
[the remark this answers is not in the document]
cipherboy: Yes, it if wasn't, it'd result in double frees. It has the same semantics as `close(...)`/releaseNativeResources(...)` on `NativeProxy`.

org/mozilla/jss/util/NativeEnclosure.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public final void close() throws Exception {
+        releaseNativeResources();
+        ptr.close();
+    }
+
+    protected abstract void acquireNativeResources() throws Exception;
+    protected abstract void releaseNativeResources() throws Exception;
```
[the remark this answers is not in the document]
cipherboy: We could, see discussion above.

org/mozilla/jss/util/NativeEnclosure.java
cipherboy: Wouldn't these objects go inside the lock? Not really sure what the concurrency model would be here anyways. At least, with our KDF, concurrency shouldn't be an issue since it'll all be set up serially (and the actual call looks like `open(...) / call_jni_kdf() / close(...)` deep within the inner details of JSS implementation.

This `NativeEnclosure` isn't really a public-facing abstract class; it is mroe of an implementation detail of how we do stuff.

org/mozilla/jss/crypto/KBKDFByteArrayParam.java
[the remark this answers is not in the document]
edewata: OK

org/mozilla/jss/util/NativeEnclosure.java
edewata: Actually, maybe not concurrency. Is open() idempotent as well? Is there any possibility that `open()` is called multiple times on the same object?

cipherboy: We could do:

```java
public final void open() throws Exception {
    if (ptr == null) {
        acquireNativeResources();
    }
}
```

And do similar for `close()` if that'd be preferable. Then if the user wishes to call `open()` multiple times, they'd have to call `close()` first, thus freeing the memory.

edewata: OK

cipherboy:
I've rebased this patchset on top of #327, but I'm waiting for that to merge first.

cipherboy:
@jmagne This is ready for review.

cipherboy:
Got [merged](https://bugzilla.mozilla.org/show_bug.cgi?id=1599603) in [NSS](https://phabricator.services.mozilla.com/D54821)! \o/

org/mozilla/jss/crypto/KBKDFCounterParams.java
```diff
@@ -0,0 +1,204 @@
...
+            throw new IllegalArgumentException(msg);
+        }
+
+        if (prf != PKCS11Constants.CKM_SHA_1_HMAC &&
+                prf != PKCS11Constants.CKM_SHA256_HMAC &&
+                prf != PKCS11Constants.CKM_SHA384_HMAC &&
+                prf != PKCS11Constants.CKM_SHA512_HMAC &&
+                prf != PKCS11Constants.CKM_AES_CMAC) {
+            String msg = "Expected Pseudo-Random Function to be SHA1-HMAC, ";
+            msg += "SHA2-HMAC, or AES-CMAC, but got unknown PKCS#11 ";
```
jmagne: Here is the error message exactly correct? For instance it doesn't list all the HMACs it is checking for above..

cipherboy: A [`SHA2`](https://en.wikipedia.org/wiki/SHA-2)-HMAC is any of `CKM_SHA256_HMAC`, `CKM_SHA384_HMAC`, or `CKM_SHA512_HMAC`.

(Technically it also includes `CKM_SHA224_HMAC`, but JSS doesn't support that anywhere so its ommission shouldn't be surprising to the caller.)

jmagne: OK, I thought maybe there was shorthand going on there, just double checking..

org/mozilla/jss/pkcs11/PK11KeyGenerator.c
```diff
@@ -476,3 +477,62 @@ Java_org_mozilla_jss_pkcs11_PK11KeyGenerator_nativeClone
     return cloneObj;
 }
 
+/***********************************************************************
+ *
+ * PK11KeyGenerator.generateKBKDF
+ *
+ */
```
jmagne: Just wondering here, since from java land we have the ParameterSpec object available , why not just pass that instead of passing it AND a bunch of other members of the object separately???

cipherboy: Some fields, such as `strength` are members of the `ParameterSpec`, but are separate parameters passed to `PK11_DeriveWithFlagsPerm`. Most of the other parameters (like the `KBKDFDataParameters`) get wrapped in an object and passed in the "mechanism parameters" (`param_item`) slot.

If we leave everything in the `ParameterSpec` object, this method looks like:

```c
generateKBKDF(parameterSpec) {
 int strength = getFieldFromJavaObject(parameterSpec, strength);
 long mechanism = getFieldFromJavaObject(parameterSpec, derivedKeyAlgorithm);
 ...
}
```

Each time you call `getFieldFromJavaObject`, you incur some overhead with the JNI call. Whereas if you pass them as arguments to `generateKBKDF` (like I am), Java handles all that setup for you, so theoretically the overhead should be better (and at worst, the same).

You also have a much longer method with more error handling logic. Java takes care of all that for you. So the sorter method, to me, wins. My 2c.

cipherboy:
@jmagne said:
> There are millions of constructors for every possible use for these classes. 

Should I drop the constructors? They're just boilerplate, and I think in most cases, the incremental building of things makes more sense than using constructors, but it is helpful for when you have partial data.

Thoughts?

cipherboy:
Note that this is blocked by the following Mozilla BZs:
 - https://bugzilla.mozilla.org/show_bug.cgi?id=1611209 (awaiting final approval from Bob)

I'd propose to hold off until the following BZ is also merged, but I'm not sure when that is:
 - https://bugzilla.mozilla.org/show_bug.cgi?id=1608245 (awaiting confirmation that Coverity is satisfied)

cipherboy:
The blocking bug has been merged upstream. I tested this locally with a sandbox build as Fedora Rawhide is [currently broken](https://bugzilla.redhat.com/show_bug.cgi?id=1801882).

PR #312, opened 2019-12-09
Support CK_ prefixed PKCS#11 constants

cipherboy:
PKCS#11 includes several constants with the `CK_` prefix, including
`CK_TRUE`, `CK_FALSE`, `CK_INVALID_HANDLE`, and others. With PKCS#11 v3.0 in
draft state, this namespace also includes various SP800-108 (KBKDF)
specific constants. We need a blacklist to handle certain function
pointer macros, but otherwise import all `CK_` constants that we can.


cipherboy:
Ah need to fix Python 2 by importing the `print` function, d'oh.

PR #313, opened 2019-12-10
Move javac command to file lists

cipherboy:
`javac` supports parsing options from a file using the `@/path/to/arguments`
syntax. This lets us pass all of our sources to `javac` in a single file
(containing all of the Java files we wish to build in this pass) instead
of passing them individually on the command line. This reduces the size
of the output significantly when there's an error in the build.

### DIFF
--- a/cmake/JSSCommon.cmake
+++ b/cmake/JSSCommon.cmake
@@ -37,6 +37,17 @@ macro(jss_build_globs)
         endif()
     endforeach()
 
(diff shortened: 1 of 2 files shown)


cipherboy:
Thanks @edewata ! Yeah, I looked at those files at the time but I didn't catch the file list. I don't recall any more why I didn't just import your Java module into JSS, but I thought I had some reason.... I forget now what that was.

PR #315, opened 2019-12-11
Allow `PK11KeyPairGenerator` use of `RSAKeyGenParameterSpec`

cipherboy:
Java provides the `java.security.spec.RSAKeyGenParameterSpec` that is the
same as our `RSAParameterSpec`, sans casing on one function call. Update
`RSAParameterSpec` to extend `RSAKeyGenParameterSpec` for backwards
compatibility and let `PK11KeyPairGenerator` use `RSAKeyGenParameterSpec`
instead.

--- 

This gets tested through the existing test cases, so I think it should be fine. Also note that PKI doesn't actually use this class. Go figure.

### DIFF
--- a/org/mozilla/jss/crypto/RSAParameterSpec.java
+++ b/org/mozilla/jss/crypto/RSAParameterSpec.java
@@ -4,12 +4,13 @@
 package org.mozilla.jss.crypto;
 
 import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.RSAKeyGenParameterSpec;
 import java.math.BigInteger;
 
 /**
  * This class specifies the parameters used for generating an RSA key pair.
  */
-public class RSAParameterSpec implements AlgorithmParameterSpec {
+public class RSAParameterSpec extends RSAKeyGenParameterSpec {
 
     /**
      * Creates a new RSAParameterSpec with the specified parameter values.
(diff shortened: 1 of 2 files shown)


org/mozilla/jss/crypto/RSAParameterSpec.java
```diff
@@ -4,12 +4,13 @@
...
 
 import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.RSAKeyGenParameterSpec;
 import java.math.BigInteger;
 
 /**
  * This class specifies the parameters used for generating an RSA key pair.
  */
-public class RSAParameterSpec implements AlgorithmParameterSpec {
+public class RSAParameterSpec extends RSAKeyGenParameterSpec implements AlgorithmParameterSpec {
```
cipherboy: Since [`RSAKeyGenParameterSpec`](https://docs.oracle.com/javase/8/docs/api/java/security/spec/RSAKeyGenParameterSpec.html) implements `AlgorithmParameterSpec`, we can drop the `implements AlgorithmParameterSpec` here.

edewata:
Should we deprecate and eventually drop `RSAParameterSpec` in the future?

cipherboy:
We probably could, since we're not relying on it for anything.

PR #317, opened 2019-12-11
Drop Apache Commons Codec Dependency

cipherboy:
Since Base64 is included in the standard library since JDK 8, the
minimum JDK version JSS `master` supports, and we already have hex
encoding in `netscape.security.utils.Utils`, we can drop all of our
dependencies on Apache Commons Codec fairly easily. Doing so reduces the
footprint of the JSS library and otherwise is beneficial since there's
likely to be little need for it in the future.

### DIFF
--- a/org/mozilla/jss/netscape/security/util/Utils.java
+++ b/org/mozilla/jss/netscape/security/util/Utils.java
@@ -32,11 +32,10 @@ import java.io.PrintWriter;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.text.SimpleDateFormat;
+import java.util.Base64;
 import java.util.Date;
 import java.util.StringTokenizer;
 import java.util.Vector;
-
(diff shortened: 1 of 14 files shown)


org/mozilla/jss/netscape/security/util/Utils.java
```diff
@@ -374,7 +373,7 @@ public static String base64encode(byte[] bytes, boolean chunked) {
      * @return base-64 encoded data
      */
     public static String base64encodeMultiLine(byte[] bytes) {
-        return new Base64(64).encodeToString(bytes);
+        return Base64.getEncoder().encodeToString(bytes);
```
edewata: I think we need to use `Base64.getMimeEncoder()` here to generate multi-line output.

cipherboy: Ah, good catch, thanks @edewata! I'll update that tomorrow and re-review the rest of these places.

Issue #298, opened 2019-10-28
Validate RSA Key Parameters


cipherboy:
This is a more systemic underlying issue with NSS: it doesn't validate parameters outside of FIPS mode and will otherwise happily ignore standard constraints imposed on public keys. Take for instance:

        KeyFactory keyFactory = KeyFactory.getInstance("RSA", "Mozilla-JSS");
        String publickey = "11770017164903779372574829180998783872141652386640045965285794535346240910202920038507767689925485723074614440813280690507719304032614007177220836643131577";
        BigInteger modulus = new BigInteger(publickey);

        BigInteger exponent = new BigInteger("3");

        RSAPublicKeySpec keySpec = new RSAPublicKeySpec(modulus, exponent);
        PublicKey publicKey = keyFactory.generatePublic(keySpec);

        String signature = "GlRa+zjMjKnHH0ipUM2ojozppxDZqb2hHQnAYvsVrmhaY3RAlUy0w24zOuUYU6vgT3Lpp9OMGd4yJIEvYyjOIw==";
        byte[] data = Base64.getDecoder().decode(signature);

        Signature signer = Signature.getInstance("SHA256withRSA", "Mozilla-JSS");
        signer.initVerify(publicKey);
        signer.update(new byte[] { 0x61, 0x62, 0x63, 0x64 });

        if (!signer.verify(data)) {
            throw new RuntimeException("Expected this to be a valid signature");
        }

A 512-bit key with `e=3` should never be silently accepted without explicitly allowing unsafe cryptography. 

We should add basic validation to both RSA and ECC based keys, when possible.

cipherboy:
Fixed in #319.

PR #319, opened 2019-12-11
Minimum RSA Key Sizes

cipherboy:
Note that this doesn't update EC (to support limiting curves) or DSA or DH -- only RSA at this point in time.  One of the issues with DSA is that NSS only supports up to 1024-bit DSA key generation, but policy requires at least 2048 bits! To fix this, we need to switch to the NSS PQG v2 API instead of the v1 API. 

---- 

Add `jss.crypto.Policy` to reflect local policy

This queries NSS for a few policy suggestions so that applications using
JSS can begin obeying local system policy.

----

Update RSA key generation to obey local policy

This uses the newly added `jss.crypto.Policy` class to validate local
parameter selection against what the system policy enforces. In
particular, this allows for validating the exponent (currently a
hard-coded sane default) and the RSA key size.

Resolves: GitHub JSS Issue #298

### DIFF
--- a/org/mozilla/jss/crypto/Policy.c
+++ b/org/mozilla/jss/crypto/Policy.c
@@ -1,0 +1,54 @@
+#include "_jni/org_mozilla_jss_crypto_Policy.h"
+
+#include <nspr.h>
+#include <nss.h>
+#include <jni.h>
+
+jint
+nearest_power_of_two(jint value)
+{
+    for (jint exponent = 8; exponent < 20; exponent++) {
+        if ((1 << exponent) >= value) {
+            return 1 << exponent;
+        }
+    }
+
+    return value;
+}
+
+JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_crypto_Policy_getRSAMinimumKeySize(JNIEnv *env, jclass clazz)
+{
+    PRInt32 value = 0;
+    SECStatus ret = NSS_OptionGet(NSS_RSA_MIN_KEY_SIZE, &value);
+    if (ret != SECSuccess) {
+        PR_ASSERT(PR_FALSE);
+    }
+
+    return nearest_power_of_two(value);
+}
+
+JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_crypto_Policy_getDHMinimumKeySize(JNIEnv *env, jclass clazz)
+{
+    PRInt32 value = 0;
+    SECStatus ret = NSS_OptionGet(NSS_DH_MIN_KEY_SIZE, &value);
+    if (ret != SECSuccess) {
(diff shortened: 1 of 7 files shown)


org/mozilla/jss/crypto/Policy.java
```diff
@@ -0,0 +1,32 @@
...
+import java.math.BigInteger;
+
+/**
+ * This class helps JSS callers align with local system cryptographic policy.
+ */
+public class Policy {
+    /**
+     * Minimum RSA key length in bits permitted by local policy.
+     */
+    public static final int RSA_MINIMUM_KEY_SIZE = getRSAMinimumKeySize();
```
edewata: Is the policy likely to change while the server is running? If so we might want to call the method directly instead of storing it as a constant. Otherwise changing the policy will require restarting all running instances.

cipherboy: Per the manual entry for `CRYPTO-POLICIES(7)`: 

> If the system administrator changes the system-wide policy level with the update-crypto-policies(8) command it is advisable to restart the system as the individual back-end libraries read the configuration files usually during their initialization. The changes in the policy level thus take place in most cases only when the applications using the back-end libraries are restarted.

So it is generally fine to leave this as is. However, it could be construed as a breaking change. I'm inclined to make these `public static int` constants (thus dropping the `final`) and document that the way to... knowing circumvent policy is to override the constant of choice.

cipherboy:
This breaks on Debian and Ubuntu, probably due to a lack of Crypto Policies, with no good fallback. I'll need to investigate more.

org/mozilla/jss/crypto/Policy.c
```diff
@@ -0,0 +1,54 @@
+#include "_jni/org_mozilla_jss_crypto_Policy.h"
+
+#include <nspr.h>
+#include <nss.h>
+#include <jni.h>
+
+jint
+nearest_power_of_two(jint value)
+{
+    for (jint exponent = 8; exponent < 20; exponent++) {
```
cipherboy: @edewata On Ubuntu 19.10, `NSS_OptionGet(NSS_RSA_MIN_KEY_SIZE, &value)` returns `1023`. When this is passed to RSA key generation, it fails because it isn't a power of two. The minimum RSA key sizes supported is either 128 iirc, but the minimum DSA key size is 256, hence choosing `1 << 8 == 256` as the lower bound in case `value == 0`.

This correctly then moves 1023 -> 1024 and handles other scenarios.

On Fedora with `crypto-policies`, the value returned by `NSS_OptionGet(NSS_RSA_MIN_KEY_SIZE, &value)` is a power of two.

edewata: LGTM.

PR #322, opened 2019-12-13
Introduce NativeEnclosure and StaticVoidPointer

cipherboy:
A `NativeEnclosure` is an abstract class which stores a `NativeProxy`
instance and wraps both the pointer and the size of the memory pointed
to by said pointer. When used in combination with `StaticVoidPointer`, a
pointer which isn't tracked and freed like the usual `NativeProxy` is,
this can be used as a form of serialization, especially for converting
parameters to and from the JNI layer.

### DIFF
--- a/org/mozilla/jss/util/NativeEnclosure.java
+++ b/org/mozilla/jss/util/NativeEnclosure.java
@@ -1,0 +1,82 @@
+package org.mozilla.jss.util;
+
+/**
+ * Classes extending NativeEnclsoure wrap a single NativeProxy instance,
+ * allowing it to be accessed from the JNI layer but be allocated and scoped
+ * from the Java layer.
+ *
+ * Because this class implements AutoCloseable, it is suggested to add
+ * constructors to derived classes which call open; this'll allow a single
+ * try-with-resources block to scope the lifetime of this object:
+ *
+ * <pre>
+ * try (NEC obj = new NEC(...)) {
+ *      // ... do something with obj ...
+ * }
+ * </pre>
+ *
+ * Extending classes implement acquireNativeResources() and
+ * releaseNativeResources(). Before this instance is passed to the JNI layer,
+ * open() should be called, allocating all necessary resources. After making
+ * all necessary JNI calls, close() should be called to free resources.
+ * Ideally, open() and close() should be called close to the JNI calls,
+ * wrapped by the developer to limit accidental memory leaks.
+ */
+public abstract class NativeEnclosure implements AutoCloseable {
+    /**
+     * Enclosed NativeProxy reference.
+     */
+    public NativeProxy mPointer;
+
+    /**
+     * Size of enclosed mPointer.
+     */
+    public long        mPointerSize;
+
+    /**
+     * Allocate and initialize mPointer with its enclosed value.
+     *
+     * Note that this method prevents you from accidentally leaking memory;
+     * to call open() twice, call close() first.
+     */
+    public final void open() throws Exception {
+        if (mPointer == null) {
+            acquireNativeResources();
+        }
+    }
+
+    @Deprecated
+    protected void finalize() throws Throwable {
+        close();
+    }
+
+    /**
+     * Deinitialize and free mPointer.
+     *
+     * Must be called to prevent memory leaks.
+     */
+    public final void close() throws Exception {
+        if (mPointer != null) {
+            releaseNativeResources();
+            mPointer.close();
+        }
+
+        mPointer = null;
+        mPointerSize = 0;
+    }
+
+    /**
+     * Allocate native resources, setting mPointer and mPointerSize as
+     * appropriate.
+     */
+    protected abstract void acquireNativeResources() throws Exception;
+
+    /**
+     * Called to deallocate native resources; note that mPointer.close()
+     * is called afterwards.
+     *
+     * If mPointer.close() should be a no-op, extend from StaticVoidRef and
+     * do any required cleanup here.
+     */
+    protected abstract void releaseNativeResources() throws Exception;
+}
--- a/org/mozilla/jss/util/NativeProxy.java
+++ b/org/mozilla/jss/util/NativeProxy.java
@@ -31,7 +31,7 @@ public abstract class NativeProxy implements AutoCloseable
 
     /**
      * Create a NativeProxy from a byte array representing a C pointer.
-     * This is the only way to create a NativeProxy, it should be called
+     * This is the primary way of creating a NativeProxy; it should be called
      * from the constructor of your subclass.
      *
      * @param pointer A byte array, created with JSS_ptrToByteArray, that
@@ -39,12 +39,27 @@ public abstract class NativeProxy implements AutoCloseable
      * NativeProxy instance acts as a proxy for that native data structure.
      */
     public NativeProxy(byte[] pointer) {
-        assert(pointer!=null);
-        mPointer = pointer;
-        registry.add(this);
+        this(pointer, true);
+    }
 
-        if (saveStacktraces) {
-            mTrace = Arrays.toString(Thread.currentThread().getStackTrace());
+    /**
+     * Create a NativeProxy from a byte array representing a C pointer.
+     * This allows for creating an untracked NativeProxy instance (when
+     * track=false), which allows for creating NativeProxy instances out
+     * of stack-allocated variables and/or creating NativeProxies which
+     * aren't freed.
+     */
+    protected NativeProxy(byte[] pointer, boolean track) {
+        mPointer = pointer;
+
+        if (track) {
+            assert(pointer != null);
+
+            registry.add(this);
+
+            if (saveStacktraces) {
+                mTrace = Arrays.toString(Thread.currentThread().getStackTrace());
+            }
         }
     }
 
--- a/org/mozilla/jss/util/NativeEnclosure.c
+++ b/org/mozilla/jss/util/NativeEnclosure.c
@@ -1,0 +1,69 @@
+#include <nspr.h>
+#include <jni.h>
+
+#include "java_ids.h"
+#include "NativeEnclosure.h"
+
+PRStatus
+JSS_PR_LoadNativeEnclosure(JNIEnv *env, jobject this, jobject *ptrObj, jlong *ptrSize)
+{
+    jclass this_class = NULL;
+    jfieldID field_id = NULL;
+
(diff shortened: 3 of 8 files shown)


org/mozilla/jss/util/NativeProxy.java
```diff
@@ -31,20 +31,35 @@
...
+    }
+
+    /**
+     * Create a NativeProxy from a byte array representing a C pointer.
+     * This allows for creating an untracked NativeProxy instance (when
+     * track=false), which allows for creating NativeProxy instances out
+     * of stack-allocated variables and/or creating NativeProxies which
+     * aren't freed.
+     */
+    public NativeProxy(byte[] pointer, boolean track) {
```
cipherboy: @edewata  Should I make this protected to only allow its use by `StaticVoidPointer` and `NativeProxy`?

edewata: It's really up to you. Basically if you're not really sure the method will be safe or useful for general usage then you can set it to protected, at least for now. If you're pretty sure it's safe/useful then it can be public. And we can always change or deprecate it later.

edewata: There are other ways too. Instead of providing an option (i.e. `track` parameter) in `NativeProxy`, you could put the code in a method (e.g. `track()`) then override it in the subclass to change its behavior for that particular class.

It's also possible to check which class is executing the code (sounds familiar?) and then run the appropriate code. This is not recommended though. :)

org/mozilla/jss/util/NativeEnclosure.java
```diff
@@ -0,0 +1,46 @@
...
+ *      // ... do something with obj ...
+ * }
+ * </pre>
+ */
+public abstract class NativeEnclosure implements AutoCloseable {
+    public NativeProxy ptr;
+    public long ptr_size;
+
+    public final void open() throws Exception {
+        if (ptr == null) {
```
edewata: How does `ptr` get initialized? Should we make it more explicit like this?
```
if (ptr == null) {
    ptr = acquireNativeResources();
}

if (ptr != null) {
    releaseNativeResources();
    ptr.close();
    ptr = null;
}
```

Just thinking, does it make sense to move `acquireNativeResources()` and `releaseNativeResources()` into `NativeProxy`?

cipherboy: A `NativeProxy` is a way of allocating, tracking, and giving control of that pointer to Java. The constructor for `NativeProxy` takes a pointer, tracks it, and provides a `close(...)` method to free it. That means--on construction of the object--this pointer needs to be non-null and provided. In order to release resources, `NativeProxy` has an abstract `releaseNativeResources()` method which calls `free(...)` or whatever else is correct in this situation. Derived classes implement that method. Examples of `NativeProxy` include [`GlobalRefProxy`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/util/GlobalRefProxy.java), [`TokenProxy`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/pkcs11/TokenProxy.java), and all of the PK11 key types.

Typically, the proxy is allocated by something else -- `CryptoManager` creates a number of `PK11Token` instances based on the local modules and stores them in the `TokenSupplier` singleton.

---

A `StaticVoidRef` inverts the first property of `NativeProxy`s: that they implement free. In particular, the `releaseNativeResources()` call is a no-op. This is useful when you have C code like:

```
{
    InsaneType obj;
    staticvoidref = wrap(&obj);
    call_some_jni(staticvoidref);
}
```

In this case, since InsaneType is scoped and doesn't need a free call (with the assumption that the JNI call doesn't store a copy of the reference, it just needs it for whatever reason), this is safe and fine.

Another example of its usage would be in the JSSTrustManager code with the legacy SSLSocket: you allocate `CERTCertificate`s in the NSS certificate handler layer, but you need to pass them to the JSSTrustManager which doesn't store them and instead uses them only for the lifetime of the JNI call. You'd ideally keep the scope of the `CERTCertificate` in the NSS certificate handler, and never free it in the JSSTrustManager because you'd have no way of knowing if it was free'd. Using a `StaticVoidRef` could work there (other than the slight fact issue that it doesn't type check but that's a different issue). 

Lastly, because it isn't something that needs to be free'd later, it can't be tracked. Sometimes a void pointer is useful in a return parameter in C code. In that case, you'd allocate a `StaticVoidRef` object with a `NULL` pointer, pass it to JNI, but since `mPointer == NULL`, tracking could fail and might involve a collision. That's not good! So we extend `NativeProxy` to disable tracking temporarily.

--- 

Lastly we get to `NativeEnclosure`: a `NativeProxy` is a way for C code to give up a pointer it created and tell Java that it gets to deal with the scope of it. However, _something_ still needs to tell C to create said pointer. That's where `NativeEnclosure` comes in: it is a contract for Java to tell C, hey, do your thing. It "encloses" the scope of the pointer: something calls `open(...)` to create `ptr` and then something calls `close(...)` to free it, but tell Java when you do either.

We use `NativeEnclosure` in conjunction with `StaticVoidRef` to create the following situation:

 - Data parameters are Java classes with constructors and setters. (this is while `ptr == null`)
 - Data parameters need to get wrapped into a C pointer "just in time" for the `derive_key` call. (this is `open(...)`)
 - Occasionally, before the C pointer is free'd, we need to exfiltrate data; this is a necessity with additional derived keys. That's why `this.releaseNativeResources(...)` is called called head of `ptr.close()`. 
     - However, since we're already getting the value of `ptr`, we might as well clean it up and free it while were here. That's why we use `StaticVoidRef` -- to prevent a double free.
 - We also can't double-free. If ptr is non-null and is of type other than `StaticVoidRef`, we'd want to track that. Hence `close(...)` only works if `ptr != null`.
 - We also can't lose memory. If `ptr` is null, you'd first have to call `close(...)` before `open(...)` works again.

Hope that makes some sense?

cipherboy: Concretely: 

> How does `ptr` get initialized? 

`ptr` gets initialized from `acquireNativeResources(...)` which is called only by open. `acquireNativeResources(...)` is an implementation detail of the derived enclosing type.

`ptr` needs to be public though, so that other (JNI) methods can access its value. This happens with e.g., `KBKDFCounterParam`, where it has an array of `KBKDFDataParameters` which it calls `open()` on and in turn calls `close()` on when it gets free'd.

org/mozilla/jss/util/NativeProxy.java
cipherboy: I've changed it to protected.

org/mozilla/jss/util/NativeEnclosure.java
cipherboy: I guess for an analogy:

`NativeEnclosure` is like curly braces...

```c
{ // Begin enclosure
int something;
void *ptr; 

... do some things ... 

// Now you have a choice: what do you put inside the enclosure?
// Do you put a reference to something statically allocated?
ptr = &something;

... do some other things ... 

// or do you dynamically allocate memory?
ptr = malloc(sizeof(int));

... do some more things ... 

// ... but if you did, you better call free:
free(ptr);

... do something else entirely ... 

} // End Enclosure 
```

`acquireNativeResouces(...)` behaves like an assignment or initialization function (e.g., `CMAC_Init(...)`). `open(...)` protects you from calling `acquireNativeResources(...)` multiple times without calling `close(...)`. `releaseNativeResources(...)` acts like a `free(...)` call / custom deconstruction handler (e.g., `CMAC_Destroy(...)`). `close(...)` protects you from calling `releaseNativeResources(...)` and `free(...)` multiple times.

edewata: Thanks for the detailed explanation! I can't say I understand everything :) but I'm just wondering, if `acquireNativeResources()` is responsible to initialize `ptr`, should we change it to return the value of the `ptr` like this?
```
ptr = acquireNativeResources();
```
This will ensure the the `ptr` will be set properly, avoiding bugs in the subclass.

org/mozilla/jss/util/NativeEnclosure.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Deprecated
+    protected void finalize() throws Throwable {
+        close();
+    }
+
+    public final void close() throws Exception {
+        if (ptr != null) {
+            releaseNativeResources();
+            ptr.close();
```
edewata: Similarly, who is responsible to nullify `ptr` after `ptr.close()`, or does it matter? Do we need to add this?
```
ptr = null;
```

edewata:
Regardless, feel free to merge. These are just implementation details.

org/mozilla/jss/util/NativeEnclosure.java
cipherboy: You're right here, I forgot to `null` it.

org/mozilla/jss/util/NativeEnclosure.java
cipherboy: There's two fields though:

```Java
public abstract class NativeEnclosure implements AutoCloseable {
    public NativeProxy ptr;
    public long ptr_size;
...
}
```

---

(aside)

In Java, you don't particularly care about the size of your object. When size is important (e.g., if its an List or an Array or something), you have attributes to find it. Also, except for a few primitive types (`int`, `int[]`, `boolean`, &c), Everything is an object.

In C, you almost always need to know the size of your pointer and what type of pointer it is, especially for memory management. The two exceptions are NULL-terminated C strings and calling `free(...)`. Regular arrays almost always have a size attached to them, pointers to structs might have a size attached to make sure that you weren't passed garbage, and a myriad of other uses.

Also, there's two "types" of pointers: a reference to the address space of a variable, typically used to return a value out of a function, and a reference to a contiguous address space of `n` elements, i.e., an array.

`NativeProxy` is just a proxy for the pointer and so it lacks size information.

`NativeEnclosure` wraps a pointer (via `NativeProxy`) _and_ includes the size information for that pointer, so you can determine if you got a 1-sized array or an n-sized array, _assuming_ you know the type of the pointer. :)

---

Anyhow, all this to say, we could return `ptr`, but we can't return both `ptr` and `ptr_size`. I can't do something like I would in C:

```
<error type> acquireNativeResources(&ptr, &ptr_size);
```

You could do something similar to the above if you changed `ptr_size` to a `Long`, but they're instance variables anyways, so I'm not sure that'd be very helpful...

I've added some comments to clarify this some more hopefully.

cipherboy:
Since this is largely meant as an internal implementation detail for JSS, I'm going ahead and merging.

PR #324, opened 2019-12-17
Add CK_ATTRIBUTE classes

cipherboy:
The new class `org.mozilla.jss.pkcs11.attrs.CKAttribute` represents a
single PKCS#11 `CK_ATTRIBUTE` in Java. These attributes can be serialized
to the native layer (via `NativeEnclosure`). Each attribute is handled in
a derived class. Currently four types of attributes are supported:

 - `CKA_CLASS` in `jss.pkcs11.attrs.CKAClass`
 - `CKA_KEY_TYPE` in `jss.pkcs11.attrs.CKAKeyType`
 - `CKA_{ENCRYPT,DECRYPT,...}` usage attributes in
   `jss.pkcs11.attrs.CKAUsage`
 - `CKA_VALUE_LEN` in `jss.pkcs11.attrs.CKAValueLen`

These attributes are useful for constructing key templates for
additional derived keys in PKCS#11's NIST SP800-108 KBKDF interface.

### DIFF
--- a/org/mozilla/jss/pkcs11/attrs/CKAClass.java
+++ b/org/mozilla/jss/pkcs11/attrs/CKAClass.java
@@ -1,0 +1,119 @@
+package org.mozilla.jss.pkcs11.attrs;
+
+import org.mozilla.jss.pkcs11.PKCS11Constants;
+
+/**
+ * CKAClass is an instance of a PKCS#11 CK_ATTRIBUTE with type = CKA_CLASS.
+ */
+public class CKAClass extends CKAttribute {
+    private long value;
+
+    /**
+     * Representation of a PKCS#11 CK_ATTRIBUTE with type CKA_CLASS and a
+     * custom value.
+     *
+     * Note: it is generally recommended to use the subclasses of this class
+     * instead of providing a custom value.
+     */
+    public CKAClass(long value) {
+        super(PKCS11Constants.CKA_CLASS);
+        setValue(value);
+    }
+
+    /**
+     * Set the value of this CKA_CLASS attribute.
+     */
+    public void setValue(long value) {
+        this.value = value;
+    }
+
+    /**
+     * Get the value of this CKA_CLASS attribute.
+     */
+    public long getValue() {
+        return value;
+    }
+
+    protected native void acquireNativeResources();
+    protected native void releaseNativeResources();
+
+    /**
+     * Representation of a PKCS#11 CK_ATTRIBUTE with type CKA_CLASS and value
+     * CKO_DATA.
+     */
+    public static class Data extends CKAClass {
+        public Data() {
+            super(PKCS11Constants.CKO_DATA);
+        }
+    }
+
+    /**
+     * Representation of a PKCS#11 CK_ATTRIBUTE with type CKA_CLASS and value
+     * CKO_CERTIFICATE.
+     */
+    public static class Certificate extends CKAClass {
+        public Certificate() {
+            super(PKCS11Constants.CKO_CERTIFICATE);
+        }
+    }
+
+    /**
+     * Representation of a PKCS#11 CK_ATTRIBUTE with type CKA_CLASS and value
+     * CKO_PUBLIC_KEY.
+     */
+    public static class PublicKey extends CKAClass {
+        public PublicKey() {
+            super(PKCS11Constants.CKO_PUBLIC_KEY);
+        }
+    }
+
+    /**
+     * Representation of a PKCS#11 CK_ATTRIBUTE with type CKA_CLASS and value
+     * CKO_PRIVATE_KEY.
+     */
+    public static class PrivateKey extends CKAClass {
+        public PrivateKey() {
+            super(PKCS11Constants.CKO_PRIVATE_KEY);
+        }
+    }
+
+    /**
+     * Representation of a PKCS#11 CK_ATTRIBUTE with type CKA_CLASS and value
+     * CKO_SECRET_KEY.
+     */
+    public static class SecretKey extends CKAClass {
+        public SecretKey() {
+            super(PKCS11Constants.CKO_SECRET_KEY);
+        }
+    }
+
+    /**
+     * Representation of a PKCS#11 CK_ATTRIBUTE with type CKA_CLASS and value
+     * CKO_HW_FEATURE.
+     */
+    public static class HWFeature extends CKAClass {
+        public HWFeature() {
+            super(PKCS11Constants.CKO_HW_FEATURE);
+        }
+    }
+
+    /**
+     * Representation of a PKCS#11 CK_ATTRIBUTE with type CKA_CLASS and value
+     * CKO_DOMAIN_PARAMETERS.
+     */
+    public static class DomainParameters extends CKAClass {
+        public DomainParameters() {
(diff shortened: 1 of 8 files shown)


org/mozilla/jss/pkcs11/attrs/CKAttribute.java
```diff
@@ -0,0 +1,7 @@
+package org.mozilla.jss.pkcs11.attrs;
+
+import org.mozilla.jss.util.NativeEnclosure;
+
```
cipherboy: @cipherboy - Needs a doc message here.

org/mozilla/jss/pkcs11/attrs/CKAttribute.c
```diff
@@ -0,0 +1,271 @@
...
+JNIEXPORT void JNICALL
+Java_org_mozilla_jss_pkcs11_attrs_CKAUsage_releaseNativeResources(JNIEnv *env, jobject this)
+{
+    CK_ATTRIBUTE_PTR attr = NULL;
+
+    if (JSS_PK11_UnwrapAttribute(env, this, &attr) != PR_SUCCESS || attr == NULL) {
+        return;
+    }
+
+    /* Since this is always a reference to JSS_CK_TRUE, don't free it! */
```
cipherboy: @cipherboy: clarify comment. This always _contains_ a reference to `JSS_CK_TRUE`; don't free the reference.

org/mozilla/jss/pkcs11/attrs/CKAttribute.c
```diff
@@ -0,0 +1,271 @@
...
+        goto failure;
+    }
+
+    if (size != sizeof(CK_ATTRIBUTE)) {
+        goto failure;
+    }
+
+    return PR_SUCCESS;
+failure:
+    *attr = NULL;
```
edewata: If `attr` is NULL, this code might cause segfault.

cipherboy:
For reference of how this is used, see [`TestKBKDF.java`](https://github.com/cipherboy/jss/blob/pkcs11-kbkdf/org/mozilla/jss/tests/TestKBKDF.java#L48-L52).

This essentially mirrors what'd happen in C, for example, [in NSS's `fipstest.c`](https://github.com/cipherboy/nss/blob/nist-800-108-kbkdf/cmd/fipstest/fipstest.c#L8307-L8313).

org/mozilla/jss/pkcs11/attrs/CKAUsage.java
```diff
@@ -0,0 +1,115 @@
+package org.mozilla.jss.pkcs11.attrs;
+
+import org.mozilla.jss.pkcs11.PKCS11Constants;
+
+/**
+ * CKAUsage is a collection of PKCS#11 CK_ATTRIBUTES which have common value
+ * (CK_TRUE).
+ */
+public class CKAUsage extends CKAttribute {
+    public CKAUsage() {}
```
edewata: This constructor allows `CKAUsage` to be created with uninitialized/default `type`. If this is not allowed this constructor probably should be removed.

org/mozilla/jss/pkcs11/attrs/CKAttribute.java
```diff
@@ -0,0 +1,27 @@
...
+ * and its size respectively. They get allocated when open() is called, and
+ * freed when close() is called.
+ *
+ * The value (numbers 2 and 3 above) get determined by the extending type,
+ * and its corresponding native methods. Some use statically allocated values,
+ * like CKAUsage's classes. Others are dynamically allocated pointers, such as
+ * when referring to CK_ULONG values (as with CKAValueLen for instance).
+ */
+public abstract class CKAttribute extends NativeEnclosure {
+    public long type;
```
edewata: If `type` is required, we probably should add a constructor that takes a `type` parameter.

org/mozilla/jss/pkcs11/attrs/CKAttribute.c
cipherboy: `attr` should be non-null by calling conventions, but I can add a `PR_ASSERT(...)` at the top of the method if you'd like.

edewata: The code in line 66 checks for null `attr`, so if that is a possibility we should check for null here too. But if that's not possible maybe the code in line 66 can be simplified?

cipherboy: Ah that was a mistake. That should be `*attr == NULL`. I want to check that the returned attribute was non-NULL, not that the address space for the returned attribute was non-NULL.

org/mozilla/jss/pkcs11/attrs/CKAUsage.java
cipherboy: Ah good catch. I forgot to add the `setType(...)` helper for this. There's a few NSS extensions to PKCS#11 that could be set here, that I haven't added classes for. So `CKAUsage() + setType()` should be possible. I don't want to add `setType(...)` to CKAttribute as it really shouldn't be necessary to set the type on other attributes.

cipherboy: Actually, I take that back. The one argument constructor is probably best here. I'll drop the `setType(...)` and no-argument constructor.

org/mozilla/jss/pkcs11/attrs/CKAttribute.java
cipherboy: Sure, I think that makes the most sense.

org/mozilla/jss/pkcs11/attrs/CKAUsage.java
jmagne: Or we could just add a default value, if applicable.

org/mozilla/jss/pkcs11/attrs/CKAClass.java
```diff
@@ -0,0 +1,123 @@
...
+
+import org.mozilla.jss.pkcs11.PKCS11Constants;
+
+/**
+ * CKAClass is an instance of a PKCS#11 CK_ATTRIBUTE with type = CKA_CLASS.
+ */
+public class CKAClass extends CKAttribute {
+    private long value;
+
+    public CKAClass() {
```
edewata: One more comment :) This constructor allows creating `CKAClass` without initializing the value. If this is not allowed probably this constructor should be removed as well. Same thing with other subclasses of `CKAAttribute`.

cipherboy: Should we drop it or make it private? I suppose it is better to have the one argument constructor rather than a no-argument constructor... And should I drop the `setValue(...)` thingies then? Currently it is using the no-argument constructor in all of the subclasses and calling `setvalue(...)`, though I could switch that to calling `super(...)` which'd probably be cleaner... I'll do that.

org/mozilla/jss/pkcs11/attrs/CKAUsage.java
cipherboy: Most of the time there isn't a default value. Only `CKA_VALUE_LENGTH` really has a default value.

edewata:
LGTM. No more comments from me :)

PR #327, opened 2019-12-17
PKCS#11: KBKDF Parameters

cipherboy:
Underlying all three KBKDFs are a common set of parameters represented
in the base `KBKDFParameterSpec` class: information about the underlying
PRF, information about the input to the PRF, and information about the
output of the PRF.

The input to the PRF are a series of `KBKDFDataParameters`, which there
are four types of:

1. `KBKDFByteArrayParam`, a fixed byte array which gets passed to the
KDF,
2. `KBKDFDKMLengthParam`, an indication of how to compute and encode the
length of the PRF output stream,
3. `KBKDFIterationVariableParam`, an indication of where to place the
iteration information (whether it is the required counter in Counter
Mode or the previous PRF's output in Feedback and Double Pipeline
modes),
4. `KBKDFOptionalCounterParam`, an indication of how to encode an
optional counter in Feedback and Double Pipeline modes.

These parameters are index sensitive: they're added to the PRF input
stream in the order they're specified.

There's also a series of derived keys. First is the implicit derived
key, which gets returned by the call to `generateKey()` -- however, other
derived keys can get specified by adding `KBKDFDerivedKey` objects to the
KBKDF parameters. No two keys may overlap in the same PRF segment.

Note that this commit only adds the data parameters and the required
`NativeEnclosure` serializations; no KDF operations are performed.

---

This is the last PR before #306 can be rebased and reviewed.


org/mozilla/jss/crypto/KBKDFDerivedKey.java
```diff
@@ -0,0 +1,107 @@
...
+
+        ArrayList<CKAttribute> arl;
+        if (attrs == null) {
+            arl = new ArrayList<CKAttribute>(1);
+        } else {
+            arl = new ArrayList<CKAttribute>(Arrays.asList(attrs));
+        }
+        arl.add(attr);
+
+        this.attrs = arl.toArray(new CKAttribute[0]);
```
edewata: Minor. Generally the array passed to `list.toArray()` should be created with `list.size()` elements:
```
this.attrs = arl.toArray(new CKAttribute[arl.size()]);
```

org/mozilla/jss/crypto/KBKDFDerivedKey.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Class for supporting additional derived keys from PKCS#11 v3.0's
+ * SP800-108 KBKDF implementation.
+ */
+public class KBKDFDerivedKey extends NativeEnclosure {
+    private CKAttribute[] attrs;
+    private long handle;
+
+    public KBKDFDerivedKey() {}
```
edewata: Is this actually allowed? The `acquireNativeResources()` will fail since `attrs` will be null.

org/mozilla/jss/crypto/KBKDF.c
```diff
@@ -0,0 +1,743 @@
...
+    jobject ptr_object = NULL;
+    jlong size = 0;
+
+    PR_ASSERT(env != NULL && this != NULL && param != NULL && param_length != NULL);
+
+    if (JSS_PR_LoadNativeEnclosure(env, this, &ptr_object, &size) != PR_SUCCESS) {
+        goto failure;
+    }
+
+    if (JSS_PR_getStaticVoidRef(env, ptr_object, (void **)param) != PR_SUCCESS || param == NULL) {
```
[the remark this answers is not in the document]
jmagne: I think edewata is correct. The actual param var has already been verified to be not NULL. Here we are looking for the returned pointer from JSS_Pr_getStaticVoidRef.

org/mozilla/jss/crypto/KBKDFDerivedKey.java
cipherboy: You can either call:

```
attrs = new CKAttribute[] { ... };
kdk = new KBKDFDerivedKey(attrs);
```

or:

```
kdk = new KBKDFDerivedKey();
kdk.addAttribute(...);
...
```

And incrementally add attributes for your derived key. `acquireNativeResources(...)` checks to make sure you have, but you have enough rope to make the code look pretty :-) 

Its the same thing with the `KBKDFParameterSpec` class. There's a `validateParameters(...)` method that'll get implemented, along with setters for each of the types... And an empty constructor if you want to build it all up. Else the derived class (`KBKDFCounterParams` / ...) has the constructors which do all of the assignments internally.

cipherboy:
@jmagne I've addressed review feedback and updated the patch.

There are no memory leaks when I run it under `valgrind` that originate from the KBKDF code.

PR #332, opened 2019-12-18
Switch SHA2/HMAC to PKCS#11 interface

cipherboy:
This enables the new KBKDF mechanisms to specify the underlying MAC
algorithm by using high-level instances (of `CMACAlgorithm` and
`HMACAlgorithm`). However, all of these need to be PKCS#11 interface and
can't be `SEC_OID`s.

---

Note that `CKM_*_HMAC` were already added to `Algorithm.{c,Java}`, so this just switches over to using it.


cipherboy:
@jmagne or @ladycfu -- do either of you remember why the SEC_OIDs were chosen over the PKCS#11 interface? I don't think it matters that much, perhaps a little higher performance overhead in calls, but nothing too significant I'd imagine.

cipherboy:
Per discussion on IRC, this has been tested and already is used for SHA-1. There's little risk in merging this, so I'm going ahead and doing so.

PR #337, opened 2019-12-18
Fix FIPS-mode TLS tests with NSS sandbox build

cipherboy:
NSS currently has partial support for TLS 1.3 in FIPS mode; this is
broken when doing a debug, sandbox build due to various internal
asserts triggering. Disable this test for now.


cipherboy:
Finally we got the sandbox build green again!

PR #339, opened 2019-12-18
Handle MIME-encoded Base64 in base64decode

cipherboy:
`jss.netscape.security.util.Utils.base64decode` previously could accept
MIME-encoded Base64 because Apache Commons Codec didn't care about the
difference. In 8de4440c5652f6f1af5b4b923a15730ba84f29e1, support was
erroneously removed. This commit re-introduces support.


cipherboy:
OK, this has been updated with review feedback from @jmagne and retested against PKI (which discovered the bug in the initial PR).

PR #342, opened 2020-01-06
Pass NSS error on CryptoManager initialization

cipherboy:
When the `CryptoManager` fails to initialize, we throw an exception. This
should include the underlying NSS error, when available, so use
`JSS_throwMsgPrErr` instead of `JSS_throwMsg`.


cipherboy:
Note: this probably warrants a more thorough review of all call sites, but this is the one I ran into.

cipherboy:
CI failure on rawhide is unrelated. Merging.

PR #346, opened 2020-01-07
SSL Alert Handling

cipherboy:
This PR enables asynchronous access to SSL Alerts events from NSS. When the `JSSL_SSLFDAlertReceivedCallback` callback is triggered, it looks up the `inboundAlerts` member of the `SSLFDProxy` and appends the new alert.

### DIFF
--- a/org/mozilla/jss/nss/SSL.java
+++ b/org/mozilla/jss/nss/SSL.java
@@ -5,9 +5,12 @@ package org.mozilla.jss.nss;
  * and handles the usage of NativeProxy objects.
  */
 
+import java.util.ArrayList;
+
 import org.mozilla.jss.pkcs11.PK11Cert;
 import org.mozilla.jss.pkcs11.PK11PrivKey;
 
+import org.mozilla.jss.ssl.SSLAlertEvent;
 import org.mozilla.jss.ssl.SSLVersionRange;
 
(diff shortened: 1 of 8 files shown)


org/mozilla/jss/nss/SSL.java
```diff
@@ -188,6 +191,24 @@ public static native int ConfigServerSessionIDCache(int maxCacheEntries,
...
     public static native int AttachClientCertCallback(SSLFDProxy fd) throws Exception;
 
+    /**
+     * Enable recording of alerts in the SSLFDProxy object.
+     *
+     * See also: SSL_AlertReceivedCallback in /usr/include/nss3/ssl.h,
+     *           SSL_AlertSentCallback in /usr/include/nss3/ssl.h
+     */
+    public static int EnableAlertLogging(SSLFDProxy fd) {
+        fd.inboundAlerts = new ArrayList<SSLAlertEvent>();
```
cipherboy: @edewata Is there a better option here?

I want a data structure that let's me:

 - Track my position in the list
 - Push new elements
 - Know when a new one has been seen
 - But also lets me access previous elements if I need to.

Up until the last requirement, that's kinda looking like a stock queue (and calling `.pop()` and `.push(...)`) but...

PR #348, opened 2020-01-07
Enable SSLAlertEvent to raise SSLExceptions

cipherboy:
When using `SSLAlertEvents` inside an `SSLEngine`, we need to raise a
`SSLException` of the correct type. There's one issue here though:

While we could add support for telling if this is an inbound or
an outbound alert, we don't know _what_ caused said alert (because
NSS created it) -- an issue on the client or an issue on the server.
This means that we can't report a `SSLKeyException` and only report
`SSLProtocolExceptions`.

Considering NSS is the one sending the alert, and we're just exposing
it happened to the calling `SSLEngine`, this is a fine trade-off.

--- 

Note that most of the time that a `SSLKeyException` would occur, it could've been detected sooner, when the key is provided to the `SSLEngine` itself and raised as an IllegalArgumentException. I think one of the few times is when the peer rejects our key. I'm not sure that'll matter much. Thoughts?

### DIFF
--- a/org/mozilla/jss/ssl/SSLAlertDescription.java
+++ b/org/mozilla/jss/ssl/SSLAlertDescription.java
@@ -4,55 +4,70 @@
 
 package org.mozilla.jss.ssl;
 
+import javax.net.ssl.SSLException;
+import javax.net.ssl.SSLHandshakeException;
+import javax.net.ssl.SSLPeerUnverifiedException;
+import javax.net.ssl.SSLProtocolException;
+
 public enum SSLAlertDescription {
 
     // see lib/ssl/ssl3prot.h in NSS
     CLOSE_NOTIFY                    (0),
     END_OF_EARLY_DATA               (1), // TLS 1.3
-    UNEXPECTED_MESSAGE              (10),
-    BAD_RECORD_MAC                  (20),
-    DECRYPTION_FAILED               (21), // RFC 5246
-    RECORD_OVERFLOW                 (22), // TLS only
-    DECOMPRESSION_FAILURE           (30),
-    HANDSHAKE_FAILURE               (40),
-    NO_CERTIFICATE                  (41), // SSL3 only, NOT TLS
-    BAD_CERTIFICATE                 (42),
-    UNSUPPORTED_CERTIFICATE         (43),
-    CERTIFICATE_REVOKED             (44),
-    CERTIFICATE_EXPIRED             (45),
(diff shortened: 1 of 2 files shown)


org/mozilla/jss/ssl/SSLAlertDescription.java
```diff
@@ -4,57 +4,71 @@
...
+    MISSING_EXTENSION               (109, SSLHandshakeException.class),
+    UNSUPPORTED_EXTENSION           (110, SSLHandshakeException.class),
+    CERTIFICATE_UNOBTAINABLE        (111, SSLPeerUnverifiedException.class),
+    UNRECOGNIZED_NAME               (112, SSLHandshakeException.class),
+    BAD_CERTIFICATE_STATUS_RESPONSE (113, SSLPeerUnverifiedException.class),
+    BAD_CERTIFICATE_HASH_VALUE      (114, SSLPeerUnverifiedException.class),
+    NO_APPLICATION_PROTOCOL         (120, SSLHandshakeException.class);
 
     private int id;
+    private Class exception;
```
edewata: Does this generate a warning in Eclipse? Usually a `Class` has to be specified as `Class<?>`.

org/mozilla/jss/ssl/SSLAlertEvent.java
```diff
@@ -98,6 +100,20 @@ public void setDescription(SSLAlertDescription description) {
         this.description = description.getID();
     }
 
+    public void throwException() throws SSLException {
+        if (levelEnum == SSLAlertLevel.FATAL) {
+            Class exception = descriptionEnum.getExceptionClass();
+
+            try {
+                throw (SSLException) exception.getConstructor(String.class).newInstance(this.toString());
+            } catch (Exception e) {
```
edewata: I suppose this will catch the `SSLException` itself, so you might want to create the instance in the `try-catch` block, but throw it outside the block. The `catch` block should catch the exception thrown by `getConstructor()` and `newInstance()` and chain it to a new `SSLException`.

cipherboy: Ah, yeah, good catch.

org/mozilla/jss/ssl/SSLAlertDescription.java
cipherboy: @edewata I've updated this to be `Class<? extends SSLException>` since that's what actually happens... :-)

PR #354, opened 2020-01-13
Fix client auth handling in JSSParameters

cipherboy:
`SSLParameters` and `SSLEngine` impose an interesting duality between the
"want client auth" and "need client auth" options: they're mutually
exclusive (and setting one clears the other). This wasn't correctly
handled in `JSSParameters` because NSS has a different notion: requesting
peer auth and requiring peer auth.

Fix `JSSParameters` to align better with Java's expectations.


cipherboy:
This mirrors new logic in [`SSLEngine`](https://github.com/cipherboy/jss/blob/ssl-engine/org/mozilla/jss/ssl/javax/JSSEngine.java#L529-L530).

PR #369, opened 2020-01-16
Track SSLCipher supported status

cipherboy:
This allows consumers of `SSLCipher` to know whether or not a given
cipher suite is supported in various conditions:

 - With TLSv1.3, or
 - In FIPS mode, or
 - With RSA or ECDSA certificates.

---- 

@edewata This doesn't quite bring full crypto-policies compliance, just FIPS compliance. I'm not sure how to do the former.

### DIFF
--- a/org/mozilla/jss/ssl/SSLCipher.c
+++ b/org/mozilla/jss/ssl/SSLCipher.c
@@ -1,0 +1,39 @@
+#include <nspr.h>
+#include <nss.h>
+#include <ssl.h>
+#include <pk11pub.h>
+#include <jni.h>
+
+#include "_jni/org_mozilla_jss_ssl_SSLCipher.h"
+
+JNIEXPORT jboolean JNICALL
+Java_org_mozilla_jss_ssl_SSLCipher_checkSupportedStatus(JNIEnv *env, jclass clazz, jint cipher_suite)
+{
+    SSLCipherSuiteInfo info = { 0 };
+    jboolean found = JNI_FALSE;
+
+    for (PRUint16 index = 0; index < SSL_NumImplementedCiphers; index++) {
+        if (SSL_ImplementedCiphers[index] == cipher_suite) {
+            found = JNI_TRUE;
+            break;
+        }
+    }
+
+    if (found == JNI_FALSE) {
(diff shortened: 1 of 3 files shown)


org/mozilla/jss/ssl/SSLCipher.c
```diff
@@ -0,0 +1,31 @@
...
+
+    if (found == JNI_FALSE) {
+        return found;
+    }
+
+    if (SSL_GetCipherSuiteInfo(cipher_suite, &info, sizeof(info)) != SECSuccess || info.length < sizeof(info)) {
+        return JNI_FALSE;
+    }
+
+    return ((!PK11_IsFIPS() || (info.isFIPS == 0)) && (info.nonStandard == 0)) ? JNI_TRUE : JNI_FALSE;
```
edewata: Shouldn't that be `info.isFIPS != 0`?

This expression is also a bit hard to read. Could we break it down maybe like this?
```
if (info.nonStandard != 0) {
    // non-standard ciphers are not supported
    return JNI_FALSE;
}
if (info.isFIPS == 0 && PKI_IsFIPS()) {
    // non-FIPS ciphers are not supported in FIPS mode
    return JNI_FALSE;
}
return JNI_TRUE;
```

org/mozilla/jss/ssl/SSLCipher.java
```diff
@@ -365,10 +379,36 @@ public boolean isECC() {
...
+    }
+
     public static SSLCipher valueOf(int id) {
         for (SSLCipher cipher : SSLCipher.class.getEnumConstants()) {
             if (cipher.id == id) return cipher;
         }
         return null;
     }
+
+    public static SSLCipher fromName(String name) {
```
edewata: Enums have built-in valueOf() so this method is redundant.
https://www.tutorialspoint.com/java/lang/enum_valueof.htm

cipherboy: Ah sorry. I was thinking I was going to need to add aliases here (between IANA and JDK names -- kinda like `SSLVersion` has), but I don't think I need to actually. So I can remove this.

PR #373, opened 2020-01-17
Log invalid Base64 and return empty data

cipherboy:
Apache Commons Codec's base64 parser will happily return invalid,
nonsensical base64 data. Add logging to `Utils` to catch these inputs,
and return an empty data array. This makes `Utils.base64decode(...)` behave
like Apache Commons Codec (in that it doesn't raise an exception), but
diverges in behavior since Base64 blobs with invalid prefixes will
return empty data, a safer alternative than continuing to parse and
returning garbage.

### DIFF
--- a/org/mozilla/jss/netscape/security/util/Utils.java
+++ b/org/mozilla/jss/netscape/security/util/Utils.java
@@ -37,7 +37,12 @@ import java.util.Date;
 import java.util.StringTokenizer;
 import java.util.Vector;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class Utils {
+    public static Logger logger = LoggerFactory.getLogger(Utils.class);
+
     /**
      * Checks if this is NT.
      */
@@ -395,7 +400,12 @@ public class Utils {
      * @return byte array
      */
     public static byte[] base64decode(String string) {
-        return Base64.getMimeDecoder().decode(string);
+        try {
+            return Base64.getMimeDecoder().decode(string);
+        } catch (IllegalArgumentException iae) {
+            logger.warn("Invalid base64: [" + string + "]: " + iae);
+            return new byte[0];
+        }
     }
 
     /**


cipherboy:
Prior to this fix, the error trace would look like:

```
2020-01-17 12:03:57 [https-jsse-nio-8443-exec-16] SEVERE: ProfileSubmitServlet: error in processing request: Last unit does not have enough valid bits
java.lang.IllegalArgumentException: Last unit does not have enough valid bits
        at java.util.Base64$Decoder.decode0(Base64.java:734)
        at java.util.Base64$Decoder.decode(Base64.java:526)
        at java.util.Base64$Decoder.decode(Base64.java:549)
        at org.mozilla.jss.netscape.security.util.Utils.base64decode(Utils.java:398)
        at org.dogtagpki.server.authentication.AuthToken.getInByteArray(AuthToken.java:113)
        at org.dogtagpki.server.authentication.AuthToken.getInStringArray(AuthToken.java:198)
        at com.netscape.cms.servlet.cert.CertProcessor.setAuthTokenIntoRequest(CertProcessor.java:133)
        at com.netscape.cms.servlet.cert.CertProcessor.populateRequests(CertProcessor.java:336)
```

If we look at the caller:

```java
    public byte[] getInByteArray(String name) {
        String value = getInString(name);
        if (value == null) {
            return null;
        }
        return Utils.base64decode(value);
    }

    public String[] getInStringArray(String name) {
        String[] stringValues;

        byte[] byteValue = getInByteArray(name);
        if (byteValue == null) {
            return null;
        }
        try {
            DerInputStream in = new DerInputStream(byteValue);
            DerValue[] derValues = in.getSequence(5);
            stringValues = new String[derValues.length];
            for (int i = 0; i < derValues.length; i++) {
                stringValues[i] = derValues[i].getAsString();
            }
        } catch (IOException e) {
            return null;
        }
        return stringValues;
    }
```

We see that it does handle `null` correctly. However, since Apache Commons Codec doesn't return `null` on invalid Base64 decode -- it returns an empty array -- we should do the same. 

In this case, the empty array gets passed to `DerInputStream`, which presumably will raise an IOException when trying to get some values, and well, fail badly. What the rest of this fallback logic does, I haven't traced. But at least at the level of `getInStringArray`, we have the correct behavior.

cipherboy:
Thanks Jack!

PR #393, opened 2020-01-30
Change to dynamic-first libjss4.so linking

cipherboy:
JSS master branch has changed to prefer a dynamic approach to finding
libjss4.so: use the value of -Djava.library.path and then fallback to
hard-coded paths if the native portion of JSS isn't found. Update the
JSS v4.4.x code to follow this behavior.

Resolves: rh-bz#1796642

### DIFF
--- a/org/mozilla/jss/CryptoManager.java
+++ b/org/mozilla/jss/CryptoManager.java
@@ -1477,24 +1477,20 @@ public final class CryptoManager implements TokenSupplier
      */
     synchronized static void loadNativeLibraries()
     {
-        if( ! mNativeLibrariesLoaded ) {
-            try { // 64 bit rhel/fedora
-                System.load( "/usr/lib64/jss/libjss4.so" );
-                Debug.trace(Debug.VERBOSE, "64-bit jss library loaded");
+        if (!mNativeLibrariesLoaded) {
+            try { // java.library.path
+                System.loadLibrary("jss4");
+                Debug.trace(Debug.VERBOSE, "jss library loaded from java.library.path");
                 mNativeLibrariesLoaded = true;
-            } catch( UnsatisfiedLinkError e ) {
-                try { // 32 bit rhel/fedora
-                    System.load( "/usr/lib/jss/libjss4.so" );
-                    Debug.trace(Debug.VERBOSE, "32-bit jss library loaded");
+            } catch (UnsatisfiedLinkError ule) {
+                try { // 64-bit library
+                    System.load("/usr/lib64/jss/libjss4.so");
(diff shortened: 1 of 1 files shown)


cipherboy:
Per meeting, this should get merged.

cipherboy:
@edewata mind doing a review? I wanted to limit the scope of this change as much as possible, so I didn't complete copy the changes from `master` (including the `static` scope). However, this does introduce a new [`UnsatisfiedLinkError`](https://docs.oracle.com/javase/8/docs/api/java/lang/UnsatisfiedLinkError.html) exception in some cases, which IMO, is a good thing.

cipherboy:
Correct my understanding if I'm wrong but... 

Doesn't 42cbe4fda805c48eb0ae15dba591486458e53760 require the library to be present and loaded when the application start? Whereas this behavior requires it to be present only if/when `CryptoManager.initialize(...)` is called?

There's a small subset of usage (mostly of the ASN.1 parser and the libraries under the `netscape.security` namespace) that don't strictly require NSS to be initialized. This would change the behavior of any small utilities which depend on this behavior. It is just like the new dependencies: slf4j is always required to be present, but the other runtime dependencies are only strictly required when calling into `netscape.security`. If you know you aren't doing so, you can safely leave them off the classpath.

While I in general agree with the behavior of 42cbe4fda805c48eb0ae15dba591486458e53760, I'll note it happened at a new minor version release of JSS (and corresponded to the release of RHEL 8; such a JSS version likely won't ever run on RHEL 7). It was a behavior change after all.

My 2c.

edewata:
Yes, the static code block is executed when the class is being referenced (i.e. loaded into JVM) for the first time. If the class is never loaded, the static code block is never executed.

Without the static code block it's possible to run certain methods in `CryptoManager` or other JSS methods without loading the native library. However, that will require us to inspect the code or test to see if the method is safe to call and then keep an accurate list of safe methods for end users to use. Also, we need to be careful when making changes not to accidentally introduce a new dependency to the native library. So it's possible, but it's a very fragile state with probably limited usefulness, and we're not doing the precautionary steps above anyway. Besides, the native library will be there anyway regardless if it's loaded or not. I'm not sure there's really a good reason not to load the native library when running JSS since the native library has specifically been written and built for JSS.

With the static code block in `CryptoManager`, and since `CryptoManager` is sort of the entry point for JSS, most of JSS code is generally safe to call without requiring us to inspect the code. However, there might be exceptions since some code may not necessarily depend on `CryptoManager`. In the future we can add a similar static code block to all classes that contain native methods, or we can introduce a true entry point for JSS that will load the native library for all.

PR #394, opened 2020-01-31
Reintroduce legacy PKIX from JSS v4.2.x

cipherboy:


cipherboy:
Note that this doesn't quite work as expected, as the relevant classes cannot be used easily or interchangeably with the non-legacy classes. This patch requires more work before it can be used like this.

PR #398, opened 2020-02-03
Advanced SSLCipher feature detection

cipherboy:
Mimic the behavior in NSS's `ssl3_config_match_init` to ensure we limit
ourselves to ciphers that NSS will actually negotiate. In particular,
validate that there exists a token that could perform the given
authentication, hash, or key type if it was required of us.

This largely mimics code in NSS's `ssl3con.c`.

### DIFF
--- a/org/mozilla/jss/ssl/SSLCipher.c
+++ b/org/mozilla/jss/ssl/SSLCipher.c
@@ -6,26 +6,118 @@
 
 #include "_jni/org_mozilla_jss_ssl_SSLCipher.h"
 
+/* Copied from NSS's ssl3con.c. */
+static const CK_MECHANISM_TYPE auth_alg_defs[] = {
+    CKM_INVALID_MECHANISM, /* ssl_auth_null */
+    CKM_RSA_PKCS,          /* ssl_auth_rsa_decrypt */
+    CKM_DSA, /* ? _SHA1 */ /* ssl_auth_dsa */
+    CKM_INVALID_MECHANISM, /* ssl_auth_kea (unused) */
+    CKM_ECDSA,             /* ssl_auth_ecdsa */
+    CKM_ECDH1_DERIVE,      /* ssl_auth_ecdh_rsa */
+    CKM_ECDH1_DERIVE,      /* ssl_auth_ecdh_ecdsa */
+    CKM_RSA_PKCS,          /* ssl_auth_rsa_sign */
+    CKM_RSA_PKCS_PSS,      /* ssl_auth_rsa_pss */
+    CKM_NSS_HKDF_SHA256,   /* ssl_auth_psk (just check for HKDF) */
+    CKM_INVALID_MECHANISM  /* ssl_auth_tls13_any */
+};
+PR_STATIC_ASSERT(PR_ARRAY_SIZE(auth_alg_defs) == ssl_auth_size);
+
+/* Copied from NSS's ssl3con.c. */
+static const CK_MECHANISM_TYPE kea_alg_defs[] = {
+    CKM_INVALID_MECHANISM, /* ssl_kea_null */
+    CKM_RSA_PKCS,          /* ssl_kea_rsa */
+    CKM_DH_PKCS_DERIVE,    /* ssl_kea_dh */
+    CKM_INVALID_MECHANISM, /* ssl_kea_fortezza (unused) */
+    CKM_ECDH1_DERIVE,      /* ssl_kea_ecdh */
+    CKM_ECDH1_DERIVE,      /* ssl_kea_ecdh_psk */
+    CKM_DH_PKCS_DERIVE,    /* ssl_kea_dh_psk */
+    CKM_INVALID_MECHANISM, /* ssl_kea_tls13_any */
+};
+PR_STATIC_ASSERT(PR_ARRAY_SIZE(kea_alg_defs) == ssl_kea_size);
+
+#if (NSS_VMAJOR >= 3) && (NSS_VMINOR >= 43)
+/* Not present in ssl3con.c. */
+static const CK_MECHANISM_TYPE hash_alg_defs[] = {
+    CKM_INVALID_MECHANISM, /* ssl_hash_none */
+    CKM_MD5,               /* ssl_hash_md5 */
+    CKM_SHA_1,             /* ssl_hash_sha1 */
+    CKM_SHA224,            /* ssl_hash_sha224 */
+    CKM_SHA256,            /* ssl_hash_sha256 */
+    CKM_SHA384,            /* ssl_hash_sha384 */
+    CKM_SHA512,            /* ssl_hash_sha512 */
+};
+#endif
+
 JNIEXPORT jboolean JNICALL
 Java_org_mozilla_jss_ssl_SSLCipher_checkSupportedStatus(JNIEnv *env, jclass clazz, jint cipher_suite)
 {
     PRInt32 allowed;
+    SSLCipherSuiteInfo info = { 0 };
 
+    /* Fetch information about whether or not this cipher is allowed by local
+     * policy. */
     if (SSL_CipherPolicyGet(cipher_suite, &allowed) != SECSuccess) {
+        /* When the previous call fails, it means that the specified cipher
+         * suite wasn't found; calls to enable it will also fail. Mark it as
+         * unsupported. */
         return JNI_FALSE;
     }
 
-    if (!PK11_IsFIPS() || allowed != SSL_ALLOWED) {
-        return (allowed == SSL_ALLOWED) ? JNI_TRUE : JNI_FALSE;
+    if (allowed != SSL_ALLOWED) {
+        /* If the cipher suite isn't allowed by policy, reject it early. */
+        return JNI_FALSE;
+    }
+
+    /* Fetch extended information about this particular cipher suite. */
+    if (SSL_GetCipherSuiteInfo(cipher_suite, &info, sizeof(info)) != SECSuccess) {
+        /* Since we know the cipher suite is good (because the call to
+         * SSL_CipherPolicyGet(...) succeeded), we know that this is because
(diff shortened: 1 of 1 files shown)


org/mozilla/jss/ssl/SSLCipher.c
```diff
@@ -6,26 +6,112 @@
...
+            info.keaType != ssl_kea_null &&
+            info.keaType != ssl_kea_tls13_any &&
+            !PK11_TokenExists(kea_alg_defs[info.keaType])) {
+        return JNI_FALSE;
+    }
+
+    /* Only check if NSS >= 3.43. Note that when this condition holds at
+     * compile time, and we're executing under an older NSS version, we're
+     * safe as the length condition will fail to hold. */
+#if (NSS_VMAJOR >= 3) && (NSS_VMINOR >= 43)
```
edewata: According to [this](https://github.com/nss-dev/nss/blob/master/lib/ssl/sslt.h#L429-L430):
```
Callers must ensure the application uses a version of NSS that
isn't older than the version used at compile time.
```
So IIUC while this code might technically work with NSS older than the one used at build time, that is not something recommended by NSS.

My concern is actually more about the consistency of a particular JSS version. Imagine JSS x.y.z was built with an older NSS on platform A, and the same JSS version was built with NSS 3.43 on platform B. If later the NSS on platform A is upgraded to 3.43, the behavior of JSS on these platforms will be inconsistent although both platforms have the same JSS and NSS versions, and that will be harder to troubleshoot.

So I suggest we specify the NSS version that provides the feature we want to use in this particular JSS version by adding this to `jss.spec`:
```
BuildRequires: nss >= 3.43
Requires: nss >= 3.43
```
I think in general the C macros should be used to resolve differences across platforms instead of introducing one. So if a platform does not provide the required NSS version, it should use an older JSS that does not require that feature, or we compensate for it in JSS using the C macros. What do you think?

cipherboy: It's tricky, but these aren't valid concerns. :-) 

> According to this:
> 
> > Callers must ensure the application uses a version of NSS that
> > isn't older than the version used at compile time.
> 
> So IIUC while this code might technically work with NSS older than the one used at build time, that is not something recommended by NSS.

The full context is important:

>     /* On return, SSL_GetCipherSuitelInfo sets |length| to the smaller of
>      * the |len| argument and the length of the struct used by NSS.
>      * Callers must ensure the application uses a version of NSS that
>      * isn't older than the version used at compile time. */

Structures in C are essentially large blocks of contiguous memory. Accessing a value is done by going to a fixed offset in this memory block, according to the order and sizes of the fields.

This contract is saying that, if your NSS version is older than what you used to compile with (_and_ the size of the `SSLCipherSuiteInfo` struct changes sizes), not all fields will be populated.

This is demonstrated by the actual NSS code for the [function in question](https://github.com/nss-dev/nss/blob/master/lib/ssl/sslinfo.c#L331-L356): it just does a `memcpy`! 

So there's three cases:

1. Sizes are the same.
2. JSS's compiled NSS is newer than system NSS.
3. JSS's compiled NSS is older than system NSS. 

1 is trivial, it just works.

When JSS's NSS exceeds the size of the system NSS (scenario 2 over a scenario where the size of the struct actually changes), NSS will bail out. I don't think there's really anything we can do here, so we have two options:

 1. Disallow the cipher suite. However, this is a global problem that affects every cipher suite. Disallowing all cipher suites indicates that there's a problem _somewhere_. We could alternatively raise an exception here.
2. Allow the cipher suite and ignore these checks. It ignores the problem, and SSL handshake could fail (if we're doing something like the test suite where we're testing one suite at a time). It turns "supported" into a blunt instrument based only upon local policy. 

When JSS's compiled NSS is older than the system NSS, we'll get a partial struct, up to the amount of info we know about. Concretely, this means, if we don't know about `kdfHash`, we don't get it. (The rest of the fields that we calculate are really, really old). 

So the question is: "can we ignore `kdfHash` if it is missing"? The answer is probably yes. System NSS supports TLSv1.3, but we don't necessarily know about it. That means we're unlikely to negotiate it, because we don't have the correct constant identifiers to name it. We'd have bounded the range by TLSv1.2.

Is this something that should fail compilation? Nah. Should I change the behavior of item 2 above? Probably.

org/mozilla/jss/ssl/SSLCipher.c
```diff
@@ -6,26 +6,112 @@
...
+        return JNI_FALSE;
+    }
+
+    have_auth_type = info.length >= (((void *)&info.authType - (void *)&info) + sizeof(info.authType));
+    have_kea_type = info.length >= (((void *)&info.keaType - (void *)&info) + sizeof(info.keaType));
+
+    if (allowed != SSL_ALLOWED) {
+        /* If the cipher suite isn't allowed by policy, reject it early. */
+        return JNI_FALSE;
     }
```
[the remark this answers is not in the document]
cipherboy: This behavior is intentional.

If the policy doesn't allow a cipher suite (and in the unlikely case where the NSS DB is in FIPS mode but the system isn't), we shouldn't negotiate that cipher suite. There's a few overlapping items here that make this more comprehensive of a "supported" check:

 - Does local policy allow us to use this cipher?
 - Can we use this cipher in FIPS mode (if the system or NSS DB is in it)?
 - Do we have all of the mechanisms necessary to support this cipher?

org/mozilla/jss/ssl/SSLCipher.c
edewata: It's not about buildability, but more about the relevancy of build time dependency. A version number of a package should normally reflect the functionality provided by the package. The runtime dependency may affect the availability of some functionality, but build time dependency usually should not be relevant.

For example, here's something that we might say in the release notes:
* JSS 4.6.3 supports TLSv1.3. NSS 3.43 or later is required in all cases.
* JSS 4.6.3 supports TLSv1.3 with NSS 3.43 or later. It can run with older NSS but it will be limited to TLSv1.2.

But probably not this:
* JSS 4.6.3 supports TLSv1.3 if it's built with NSS 3.43 or later. It's limited to TLSv1.2 if it was build with older NSS regardless of actual NSS used at runtime.

The first 2 cases will guarantee the availability of TLSv1.3 by installing JSS 4.6.3 and NSS 3.43 or later, which maybe useful for system administrators trying to meet system requirements. However, in the third case it may be difficult to figure out which NSS was used to build JSS because usually that information is not well published, so even if JSS 4.6.3 and NSS 3.43 were installed, there is no guarantee that TLSv1.3 will be available, which could be misleading.

@jmagne, what's your take on this?

org/mozilla/jss/ssl/SSLCipher.c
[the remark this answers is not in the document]
edewata: OK.

org/mozilla/jss/ssl/SSLCipher.c
cipherboy: Keep in mind -- this "supported" check isn't a final whitelist / blacklist. It is a _heuristic_ designed to inform our decision of [`SSLEngine.getSupportedCipherSuites(...)`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html#getSupportedCipherSuites--). NSS is ultimately the one who chooses the negotiated cipher suite, based on whether or not the programmer calls [`SSLEngine.setEnabledCipherSuites(...)`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html#setEnabledCipherSuites-java.lang.String:A-) and what the peer sends in their cipher suite selection.

We just need to be reasonably sure that we _could_ potentially negotiate these cipher suites, if we had to. So if we're to err, we should err on the side allowing too many ciphers than allowing too few (at the risk of having to blacklist them in the test suite). 

This is complicated a little by the fact that Tomcat uses this as an absolute filter -- so if the cipher isn't reported in `SSLEngine.getSupportedCipherSuites(...)`, Tomcat won't pass it to `SSLEngine.setEnabledCipherSuites(...)`

org/mozilla/jss/ssl/SSLCipher.c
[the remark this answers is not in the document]
cipherboy: I agree... I think we could move this up though.

In particular, if `SSL_CipherPolicyGet` fails, it shows that we've passed an invalid cipher suite (which shouldn't happen in most cases, unless we add a new one to the `SSLCipher` enum and the underlying NSS doesn't support it). We can then check allowed status and return `JNI_FALSE` if it isn't allowed at all. Then when `SSL_GetCipherSuiteInfo` is called -- the only failure case remaining is when the `cipher_suite` struct we pass exceeds the size of what the system NSS library supports. We can return `JNI_TRUE` instead of `JNI_FALSE` -- because it indicates an internal error and we can't get more precise information.

org/mozilla/jss/ssl/SSLCipher.c
cipherboy: I just tested building with NSS v3.42 (which has a smaller struct size). Then I rebuilt NSS to version v3.49 and tried running the tests again:

```
58/72 Test #51: SSLEngine_RSA .....................................   Passed  105.50 sec
70/72 Test #52: SSLEngine_ECDSA ...................................   Passed    8.56 sec
......
```

So I think I agree with that last statement ("JSS 4.6.3 supports TLSv1.3 if it's built with NSS 3.43 or later. It's limited to TLSv1.2 if it was build with older NSS regardless of actual NSS used at runtime.") and I think it is desired. 

Namely, if you need TLSv1.3, you need a new enough NSS version which supports it -- both at build time and at run time. I don't think artificially inflating the build version to get TLSv1.3 support does much for us... ...because you'd have the same problem with the SSLSocket interface, in my opinion, and that doesn't have this artificial constraint.

cipherboy:
Thanks @edewata!

PR #408, opened 2020-02-13
Mark JSS Provider and PKCS#11 classes as public

cipherboy:
This enables javadoc generation for these classes. While the JSSProvider
clases lack useful javadocs, their existence helps developers check the
supported interfaces. Additionally, the PKCS#11 `PrivateKey` interfaces
should be made public to mirror their `PublicKey` counterparts, in the
rare instances where they're used instead of the generic Java
interfaces.

### DIFF
--- a/org/mozilla/jss/pkcs11/PK11DSAPrivateKey.java
+++ b/org/mozilla/jss/pkcs11/PK11DSAPrivateKey.java
@@ -6,7 +6,7 @@ import java.math.BigInteger;
 import java.security.interfaces.DSAParams;
 import java.security.interfaces.DSAPrivateKey;
 
-class PK11DSAPrivateKey
+public class PK11DSAPrivateKey
     extends PK11PrivKey implements DSAPrivateKey
 {
 
--- a/org/mozilla/jss/pkcs11/PK11ECPrivateKey.java
+++ b/org/mozilla/jss/pkcs11/PK11ECPrivateKey.java
@@ -7,7 +7,7 @@ import java.math.BigInteger;
 import org.mozilla.jss.crypto.PrivateKey;
 import org.mozilla.jss.util.EC;
 
-class PK11ECPrivateKey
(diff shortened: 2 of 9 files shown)


org/mozilla/jss/pkcs11/PK11DSAPrivateKey.java
```diff
@@ -6,7 +6,7 @@
 import java.security.interfaces.DSAParams;
 import java.security.interfaces.DSAPrivateKey;
 
-class PK11DSAPrivateKey
+public class PK11DSAPrivateKey
     extends PK11PrivKey implements DSAPrivateKey
```
edewata: This is separate from this PR. Can we rename `PK11PrivKey` to `PK11PrivateKey` since it will appear in javadoc? I wouldn't worry about backward compatibility as long as it's done in a new minor version (i.e. 4.x).

cipherboy: `PK11PrivKey` [already exists in Javadocs](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/pkcs11/PK11PrivKey.html), without this change.

While a historical artifact, I think the naming makes some sense: the core classes have `PK11PrivKey` / `PK11PubKey` / `PK11Cert` / `PK11SymKey`, but while they're what gets used internally to JSS most often, as most applications (algorithms which use these keys) are agnostic of actual key type. 

However, occasionally more data is necessary and it is necessary to extract details about the (public) keys. Hence why the end classes are necessary. That happens, e.g., in the JSS Provider SPIs.

edewata: I'm not sure about symkey or cert, but shortening private and public into priv and pub seems to be uncommon, unnecessary, and inconsistent with the interfaces they're implementing (i.e. `PrivateKey` and `PublicKey`).

PR #410, opened 2020-02-24
Merge IVParameterSpec into IvParameterSpec

cipherboy:
[`org.mozilla.jss.crypto.IVParameterSpec`](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/crypto/IVParameterSpec.html) is an artifact from the dark ages when Java didn't support much crypto.

Now there's [`javax.crypto.spec.IvParameterSpec`](https://docs.oracle.com/javase/8/docs/api/javax/crypto/spec/IvParameterSpec.html). We've supported both for a while. To finally deprecate the former, make it inherit from the latter, removing its implementation completely.

There's two commits here:

1. Merge the implementations.
2. Combine all usages of the former into only caring about the latter. 

If you have better one-line commit messages, do tell me, else I could just do it in one commit.

### DIFF
--- a/org/mozilla/jss/crypto/Algorithm.java
+++ b/org/mozilla/jss/crypto/Algorithm.java
@@ -112,18 +112,29 @@ public class Algorithm {
      * @return Returns <code>true</code> if the given Object can be used as a parameter.
      */
     public boolean isValidParameterObject(Object o) {
-        if( o == null ) {
+        if (o == null) {
+            // It is only valid to pass a null parameter if the length of the
+            // expected class list is zero. Otherwise, the usage is invalid.
             return (parameterClasses.length == 0);
         }
-        if( parameterClasses.length == 0 ){
-            return false;
-        }
-        Class<?> c = o.getClass();
-        for( int i = 0; i < parameterClasses.length; ++i) {
-            if( c.equals( parameterClasses[i] ) ) {
+
+        // For the remaining instances, if the object's class is a subclass
+        // of the specified parameter class, allow it. This allows the
+        // construct:
+        //
+        //   IVParameterClass -child-of-> IvParameterClass
+        //
+        // by only specifying the parent, IvParameterClass.
+        for (int i = 0; i < parameterClasses.length; i++) {
+            if (parameterClasses[i].isInstance(o)) {
                 return true;
             }
         }
+
+        // If we got an object parameter, but weren't expecting any, it
+        // is invalid. Also, if our class didn't match our expectations,
+        // it is also invalid.
+
         return false;
     }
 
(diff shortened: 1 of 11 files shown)


org/mozilla/jss/crypto/Algorithm.java
```diff
@@ -112,18 +112,31 @@ public OBJECT_IDENTIFIER toOID() throws NoSuchAlgorithmException {
...
+            // expected class list is zero. Otherwise, the usage is invalid.
             return (parameterClasses.length == 0);
         }
-        if( parameterClasses.length == 0 ){
+
+        if (parameterClasses.length == 0) {
+            // If we got an object parameter, but weren't expecting any, it
+            // is invalid.
             return false;
         }
```
edewata: This case is actually covered by the subsequent loop. It's up to you if you want to remove it.

cipherboy:
@edewata correct me if I'm wrong, but AFAICT, there's only one place (in the commented out section of `AuthenticatedSafes.java`) we're iterating over the class of the _parameter_ itself, and comparing it to the `IvParameterSpec` with `equals(...)`.

Otherwise, we're always iterating over the _algorithm's_ accepted parameter whitelist, from which we've removed `IVParameterSpec`, and left only `IvParameterSpec`. We're just checking if the algorithm takes an `IvParameterSpec`, and if so, constructing a new one with a random IV. IMO, equals is fine there.

cipherboy:
> Could you replace the former and the latter with actual class names since someone might be reading this commit message without knowing about the earlier commit?

Ah sorry Endi, I had reworded the commit to simplify it, removing one of the names. :-) Fixed!

edewata:
> Otherwise, we're always iterating over the _algorithm's_ accepted parameter whitelist, from which we've removed `IVParameterSpec`, and left only `IvParameterSpec`. We're just checking if the algorithm takes an `IvParameterSpec`, and if so, constructing a new one with a random IV. IMO, equals is fine there.

I was just wondering if someone created a subclass of `Algorithm` using the old `IVParameterSpec`, maybe the code will stop working? Or maybe that's very unlikely?

Regardless, I think we should mark `IVParameterSpec` class deprecated with annotations.

cipherboy:
> I was just wondering if someone created a subclass of Algorithm using the old IVParameterSpec, maybe the code will stop working? Or maybe that's very unlikely?

In order to create a new type of Algorithm, I think you'd need to:

 - Add new constants to Algorithm.
 - Create a new subclass of Algorithm, to define the interface desired.
 - (Optionally) create a new algorithm interface.
 - Create a new implementation under PKCS#11.
 - Update the JSS Provider to have the new interface/...

IMO, that relegates it to a custom fork of JSS because it is so much work. I don't think it is possible to have JSS behave nicely with said algorithm if it isn't part of the JSS tree. They can update when possible.

Also, the most common constructs with an IV are already covered under this merge request. Things like HMACs typically don't take an IV, etc.

So I think it is unlikely. I don't think their code will stop working, and ours shouldn't stop working either. Notably, the name `org.mozilla.jss.crypto.IVParameterSpec` still exists, and obeys the original specification for the class.

PR #411, opened 2020-02-25
Fix base64-encoding of CSRs

cipherboy:
In 8de4440c5652f6f1af5b4b923a15730ba84f29e1, the base64 encoder was
changed from `apache-commons-codec` to the Java standard library to drop
a dependency. However, the behavior changed as a result: the Java
standard library doesn't include a final line separator, whereas
`apache-commons-codec` did. This results in malformed CSRs:

> YWRPxyBKvFAOB29fwPwBJLZksrwQ0xAs7sooc+qF-----END NEW CERTIFICATE REQUEST-----

Resolves: https://pagure.io/freeipa/issue/8199

### DIFF
--- a/org/mozilla/jss/netscape/security/util/Utils.java
+++ b/org/mozilla/jss/netscape/security/util/Utils.java
@@ -378,7 +378,10 @@ public class Utils {
      * @return base-64 encoded data
      */
     public static String base64encodeMultiLine(byte[] bytes) {
-        return Base64.getMimeEncoder().encodeToString(bytes);
+        // When switching from apache-commons-codec to the standard library,
+        // the standard library does not include a final line separator at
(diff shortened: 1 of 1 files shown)


cipherboy:
@flo-renaud - this was the fix. I want to test it with PKI and RHCS first, then I'll merge and it'll be available in COPR.

flo-renaud:
@cipherboy 
thanks for the quick fix. freeipa non-reg against @pki/master are automatically launched on Sundays, but I will be able to run on a test machine when the fix is available in the copr repo. Just let me know :)

flo-renaud:
@cipherboy 
I did a quick test with jss-4.6.3-1.20200226223702.f580a535.fc31.x86_64 from the copr repo, LGTM. Thanks for the fix :)

PR #413, opened 2020-02-26
Remove deprecated netscape.security.acl

cipherboy:
The corresponding JDK interfaces under the `java.security.acl` package
namespace have been marked deprecated since Java 1.2. The implementation
of these interfaces isn't used by JSS or Dogtag and is thus removed.


edewata:
I'm not familiar with these classes so I'll let @jmagne decide. I have no objection removing it.

Just to clarify, it's actually deprecated only in JDK 9, but the replacement is available since JDK 1.2. See:
* https://docs.oracle.com/javase/8/docs/api/java/security/acl/package-summary.html
* https://docs.oracle.com/javase/9/docs/api/java/security/acl/package-summary.html

cipherboy:
@edewata Thanks, I've updated my commit.

cipherboy:
This code is also present under the `sun.security.acl` namespace of the JDK. Anyone wishing to use this can use JDK 8 and change classes to that version instead.

PR #418, opened 2020-03-02
Finish releasing slots after use

cipherboy:
Slots in NSS are reference counted; it is important to release our copy
of the reference when we're done using it. This fixes various instances
of slots either not being released or being released in the wrong order.

This should help to address [`rh-bz#676083`](https://bugzilla.redhat.com/show_bug.cgi?id=676083). It should be tested with additional subsystems to ensure it doesn't break Dogtag. 


cipherboy:
Tested in my Ansible environment. Hopefully this won't break any other instances, but merging early in this next release cycle should help us to find any.

PR #419, opened 2020-03-05
Query and Set Default SSL configuration

cipherboy:
When splitting `JSSEngine` into an abstract class and a reference implementation (in #150), I've left the `SSLFDProxy` instance, `ssl_fd`, in the reference implementation, but moved all the cipher suite configuration to JSSEngine.

This resulted in the need to query default configuration, instead of asking `ssl_fd` how it was configured, though technically this need already existed (as `ssl_fd` could've been `null`, which we largely ignored).

This introduces wrappers for `SSL_CipherPrefSetDefault`, `SSL_CipherPrefGetDefault`, `SSL_VersionRangeGetDefault`, and `SSL_VersionRangeSetDefault`.

### DIFF
--- a/org/mozilla/jss/nss/SSL.c
+++ b/org/mozilla/jss/nss/SSL.c
@@ -278,6 +278,66 @@ Java_org_mozilla_jss_nss_SSL_VersionRangeGet(JNIEnv *env, jclass clazz,
     return JSS_SSL_wrapVersionRange(env, vrange);
 }
 
+JNIEXPORT int JNICALL
+Java_org_mozilla_jss_nss_SSL_VersionRangeSetDefaultNative(JNIEnv *env, jclass clazz,
+    jint variant_ssl, jint min_ssl, jint max_ssl)
+{
+    SSLVersionRange vrange;
+    SSLProtocolVariant variant;
+
+    PR_ASSERT(env != NULL);
+    PR_SetError(0, 0);
+
+    if (min_ssl < 0 || min_ssl >= JSSL_enums_size ||
+            max_ssl < 0 || max_ssl >= JSSL_enums_size ||
+            variant_ssl < 0 || variant_ssl >= JSSL_enums_size)
+    {
+        char buf[200];
+        snprintf(buf, 200,
+                 "SSL.VersionRangeSetDefaultNative(): for min=%d max=%d "
+                 "variant=%d failed - out of range for array JSSL_enums "
+                 "size: %d", min_ssl, max_ssl, variant_ssl, JSSL_enums_size);
+        JSSL_throwSSLSocketException(env, buf);
+        return SECFailure;
+    }
+
+    vrange.min = JSSL_enums[min_ssl];
+    vrange.max = JSSL_enums[max_ssl];
+    variant = JSSL_enums[variant_ssl];
+
+    return SSL_VersionRangeSetDefault(variant, &vrange);
+}
+
+JNIEXPORT jobject JNICALL
+Java_org_mozilla_jss_nss_SSL_VersionRangeGetDefaultNative(JNIEnv *env,
+    jclass clazz, jint variant_ssl)
+{
+    SSLVersionRange vrange;
+    SSLProtocolVariant variant;
+
+    PR_SetError(0, 0);
+
+    if (variant_ssl < 0 || variant_ssl >= JSSL_enums_size) {
+        char buf[200];
+        snprintf(buf, 200,
+                 "SSL.VersionRangeGetDefaultNative(): for variant=%d failed: "
+                 "out of range for array JSSL_enums size: %d", variant_ssl,
+                 JSSL_enums_size);
+        JSSL_throwSSLSocketException(env, buf);
+        return NULL;
+    }
+
+    variant = JSSL_enums[variant_ssl];
+
+    if (SSL_VersionRangeGetDefault(variant, &vrange) != SECSuccess) {
+        JSS_throwMsg(env, INVALID_PARAMETER_EXCEPTION,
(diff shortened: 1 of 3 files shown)


org/mozilla/jss/nss/SSL.c
```diff
@@ -278,6 +278,66 @@ Java_org_mozilla_jss_nss_SSL_VersionRangeGet(JNIEnv *env, jclass clazz,
...
+
+    PR_ASSERT(env != NULL);
+    PR_SetError(0, 0);
+
+    if (min_ssl < 0 || min_ssl >= JSSL_enums_size ||
+            max_ssl < 0 || max_ssl >= JSSL_enums_size ||
+            variant_ssl < 0 || variant_ssl >= JSSL_enums_size)
+    {
+        char buf[148];
+        snprintf(buf, 148,
```
edewata: Is this the exact length of the current string? I'd give it a larger size (e.g. 200) to accommodate the variables and the possibilities of someone updating the text without updating the buffer size.

org/mozilla/jss/ssl/SSLProtocolVariant.java
```diff
@@ -16,5 +16,5 @@
 
     private SSLProtocolVariant(int val) { _enum = val; }
 
-    int getEnum() { return _enum; }
+    public int getEnum() { return _enum; }
```
edewata: If we want to mimic Java Enum style, we probably can add a `value()` that returns `_enum`.

org/mozilla/jss/nss/SSL.c
```diff
@@ -192,6 +192,34 @@ Java_org_mozilla_jss_nss_SSL_CipherPrefGet(JNIEnv *env, jclass clazz,
...
+    jint cipher)
+{
+    int enabled = false;
+
+    PR_ASSERT(env != NULL);
+    PR_SetError(0, 0);
+
+    if (SSL_CipherPrefGetDefault(cipher, &enabled) != SECSuccess) {
+        JSS_throwMsg(env, ILLEGAL_ARGUMENT_EXCEPTION,
+            "Unknown cipher suite to get or getting its value failed");
```
edewata: Is there any error code/message from NSS that we can include?

cipherboy: Yes and no:

```c
SECStatus
SSL_CipherPrefGetDefault(PRInt32 which, PRBool *enabled)
{
    SECStatus rv;

    if (!enabled) {
        PORT_SetError(SEC_ERROR_INVALID_ARGS);
        return SECFailure;
    }
    if (ssl_IsRemovedCipherSuite(which)) {
        *enabled = PR_FALSE;
        rv = SECSuccess;
    } else {
        rv = ssl3_CipherPrefGetDefault((ssl3CipherSuite)which, enabled);
    }
    return rv;
}

/* return the user preference for this suite */
SECStatus
ssl3_CipherPrefGetDefault(ssl3CipherSuite which, PRBool *enabled)
{
    const ssl3CipherSuiteCfg *suite;
    PRBool pref;
    SECStatus rv;

    suite = ssl_LookupCipherSuiteCfg(which, cipherSuites);
    if (suite) {
        pref = suite->enabled;
        rv = SECSuccess;
    } else {
        pref = SSL_NOT_ALLOWED;
        rv = SECFailure; /* err code was set by Lookup. */
    }
    *enabled = pref;
    return rv;
}

/* Find the cipher configuration struct associate with suite */
/* XXX This does a linear search.  A binary search would be better. */
static ssl3CipherSuiteCfg *
ssl_LookupCipherSuiteCfgMutable(ssl3CipherSuite suite,
                                ssl3CipherSuiteCfg *suites)
{
    int i;

    for (i = 0; i < ssl_V3_SUITES_IMPLEMENTED; i++) {
        if (suites[i].cipher_suite == suite)
            return &suites[i];
    }
    /* return NULL and let the caller handle it.  */
    PORT_SetError(SSL_ERROR_UNKNOWN_CIPHER_SUITE);
    return NULL;
}

const ssl3CipherSuiteCfg *
ssl_LookupCipherSuiteCfg(ssl3CipherSuite suite, const ssl3CipherSuiteCfg *suites)
{
    return ssl_LookupCipherSuiteCfgMutable(suite,
                                           CONST_CAST(ssl3CipherSuiteCfg, suites));
}
```

Failure case in `SSL_CipherPrefGetDefault` won't get hit unless something hits the fan and C passes a `NULL` value for a stack variable reference. Should never happen.

The only other remaining failure case is `SSL_ERROR_UNKNOWN_CIPHER_SUITE` -- that means the NSS version doesn't know our given cipher suite. Should rarely happen and is safe to ignore on the aggregate (like we do in JSSEngine).

org/mozilla/jss/nss/SSL.c
cipherboy: The length of the string is 108 without digit placeholders. Each number can be at most 11 digits (since it is a 32-bit signed integer), which means the result fits in 108 + 4*11 = 152 characters. So when numbers were negative, it was actually too small!

I've fixed this to 200 as suggested. However, note that this is unlikely, as it mostly connotes a mismatch between JSS Java and JSS Native components. (Or reflection).

org/mozilla/jss/ssl/SSLProtocolVariant.java
cipherboy: Yeah, a future improvement would be to make these actual enums, but for the time being, I think we'll leave it as-is.

cipherboy:
Thanks @edewata! I think my error message is as informative as the error code from NSS (it says the same thing), so I'm going ahead and merging.

PR #421, opened 2020-03-05
JSSProvider - java.security based loader

cipherboy:
When this works, it will eventually allow us to integrate more closely into the JDK ecosystem. In particular, it'll allow us to replace the providers in `java.security` with our own override list, which means FIPS mode will be more stable.

Eventually, this will also remove some of the need TomcatJSS's initialization logic. In particular, JSS will be already initialized by the JDK prior to starting TomcatJSS, so we can access the JSSProvider's SSLEngine already.

### DIFF
--- a/org/mozilla/jss/JSSProvider.java
+++ b/org/mozilla/jss/JSSProvider.java
@@ -3,6 +3,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package org.mozilla.jss;
 
+import java.lang.NullPointerException;
+import java.security.Provider;
+
+import java.io.InputStream;
+
 public final class JSSProvider extends java.security.Provider {
 
     private static final long serialVersionUID = 1L;
@@ -18,10 +23,58 @@ public final class JSSProvider extends java.security.Provider {
                                            (JSS_MINOR_VERSION * 100 +
                                             JSS_PATCH_VERSION)/10000.0;
 
+    private static JSSLoader loader = new JSSLoader();
+
     public JSSProvider() {
+        this(loader.loaded());
+    }
+
+    public JSSProvider(boolean initialize) {
         super("Mozilla-JSS", JSS_VERSION,
                 "Provides Signature, Message Digesting, and RNG");
 
+        if (initialize) {
+            initializeProvider();
+        }
+    }
+
+    public JSSProvider(String config_path) throws Exception {
+        this(false);
+
+        configure(config_path);
+    }
+
+    public JSSProvider(InputStream config) throws Exception {
+        this(false);
+
+        loader.init(config);
+        initializeProvider();
+    }
+
+    /**
+     * Configure this instance of JSSProvider with the specified path
+     * to a JSS configuration properties file.
+     *
+     * See JSSLoader's class description for a description of the JSS
+     * configuration properties file and known values.
+     *
+     * If the JSSProvider is already loaded, this is a no-op.
+     */
+    public Provider configure(String arg) {
+        try {
+            loader.init(arg);
+        } catch (NullPointerException npe) {
+            throw npe;
+        } catch (Exception e) {
+            throw new RuntimeException(e.getMessage(), e);
+        }
+
+        initializeProvider();
+
+        return this;
+    }
+
+    protected void initializeProvider() {
         /////////////////////////////////////////////////////////////
         // Signature
         /////////////////////////////////////////////////////////////
@@ -324,7 +377,7 @@ public final class JSSProvider extends java.security.Provider {
     }
 
     public String toString() {
-        String mozillaProviderVersion = JSS_MAJOR_VERSION + "." + 
+        String mozillaProviderVersion = JSS_MAJOR_VERSION + "." +
                                         JSS_MINOR_VERSION;
         if ( JSS_PATCH_VERSION != 0 ) {
             mozillaProviderVersion = mozillaProviderVersion + "." +
--- a/cmake/JSSCommon.cmake
+++ b/cmake/JSSCommon.cmake
@@ -191,15 +191,13 @@ macro(jss_build_jars)
     # JAR.
     add_custom_command(
         OUTPUT "${JSS_BUILD_JAR_PATH}"
-        COMMAND "${Java_JAR_EXECUTABLE}" cmf "${CMAKE_BINARY_DIR}/MANIFEST.MF" ${JSS_BUILD_JAR_PATH} org/*
-        WORKING_DIRECTORY "${CLASSES_OUTPUT_DIR}"
+        COMMAND "${Java_JAR_EXECUTABLE}" cmf "${CMAKE_BINARY_DIR}/MANIFEST.MF" "${JSS_BUILD_JAR_PATH}" -C "${CLASSES_OUTPUT_DIR}" org -C "${CLASSES_OUTPUT_DIR}" META-INF
         DEPENDS generate_java
     )
 
     add_custom_command(
         OUTPUT "${JSS_TESTS_JAR_PATH}"
-        COMMAND "${Java_JAR_EXECUTABLE}" cmf "${CMAKE_BINARY_DIR}/MANIFEST.MF" ${JSS_TESTS_JAR_PATH} org/*
-        WORKING_DIRECTORY "${TESTS_CLASSES_OUTPUT_DIR}"
+        COMMAND "${Java_JAR_EXECUTABLE}" cmf "${CMAKE_BINARY_DIR}/MANIFEST.MF" ${JSS_TESTS_JAR_PATH} -C "${TESTS_CLASSES_OUTPUT_DIR}" org
         DEPENDS generate_java
     )
 
--- a/cmake/JSSConfig.cmake
+++ b/cmake/JSSConfig.cmake
@@ -14,6 +14,9 @@ macro(jss_config)
     # Configure java-related flags
     jss_config_java()
 
+    # Configure test variables
+    jss_config_tests()
+
     # Template auto-generated files
     jss_config_template()
 
@@ -57,6 +60,7 @@ macro(jss_config_outputs)
     # Global variables representing various output files; note that these
     # directories are created at the end of this macro.
     set(CLASSES_OUTPUT_DIR "${CMAKE_BINARY_DIR}/classes/jss")
+    set(CONFIG_OUTPUT_DIR "${CMAKE_BINARY_DIR}/config")
     set(DOCS_OUTPUT_DIR "${CMAKE_BINARY_DIR}/docs")
     set(LIB_OUTPUT_DIR "${CMAKE_BINARY_DIR}/lib")
     set(BIN_OUTPUT_DIR "${CMAKE_BINARY_DIR}/bin")
@@ -97,6 +101,7 @@ macro(jss_config_outputs)
 
     # Create the *_OUTPUT_DIR locations.
     file(MAKE_DIRECTORY "${CLASSES_OUTPUT_DIR}")
+    file(MAKE_DIRECTORY "${CONFIG_OUTPUT_DIR}")
     file(MAKE_DIRECTORY "${DOCS_OUTPUT_DIR}")
     file(MAKE_DIRECTORY "${LIB_OUTPUT_DIR}")
     file(MAKE_DIRECTORY "${BIN_OUTPUT_DIR}")
@@ -313,6 +318,9 @@ macro(jss_config_java)
     set(JSS_BASE_PORT 2876)
     math(EXPR JSS_TEST_PORT_CLIENTAUTH ${JSS_BASE_PORT}+0)
     math(EXPR JSS_TEST_PORT_CLIENTAUTH_FIPS ${JSS_BASE_PORT}+1)
+
+    # Create META-INF directory for provider
+    file(MAKE_DIRECTORY "${CLASSES_OUTPUT_DIR}/META-INF/services")
 endmacro()
 
 macro(jss_config_template)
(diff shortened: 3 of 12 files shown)


org/mozilla/jss/JSSProvider.java
```diff
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 package org.mozilla.jss;
 
-public final class JSSProvider extends java.security.Provider {
+public class JSSProvider extends java.security.Provider {
```
cipherboy: @edewata We can't encapsulate JSSProviderLoader inside JSSProvider like this:

```java
public final class JSSProvider extends java.security.Provider {
    public final class Loader extends JSSProvider { }
}
``` 

Because it complains about final:

```java
cd /home/ascheel/GitHub/c2/jss/build && /etc/alternatives/java_sdk/bin/javac -classpath /usr/share/java/slf4j/api.jar:/usr/share/java/apache-commons-lang.jar:/usr/share/java/jaxb-api.jar -sourcepath /home/ascheel/GitHub/c2/jss -target 1.8 -source 1.8 -g -d /home/ascheel/GitHub/c2/jss/build/classes/jss -h /home/ascheel/GitHub/c2/jss/build/include/jss/_jni @/home/ascheel/GitHub/c2/jss/build/.targets/java.sources && touch /home/ascheel/GitHub/c2/jss/build/.targets/finished_generate_java
/home/ascheel/GitHub/c2/jss/org/mozilla/jss/JSSProvider.java:346: error: cannot inherit from final JSSProvider
    public final static class JSSProviderLoader extends JSSProvider {
                                                        ^
/home/ascheel/GitHub/c2/jss/org/mozilla/jss/provider/java/security/JSSProviderLoader.java:12: error: cannot inherit from final JSSProvider
public class JSSProviderLoader extends JSSProvider {
                                       ^
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: /home/ascheel/GitHub/c2/jss/org/mozilla/jss/ssl/SSLAlertDescription.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
2 errors
```

My preference would be to leave `JSSProvider` as final somehow. Should I try and add detection logic to `JSSProvider` to see if a CM has been initialized, and if not, call `JSSProviderLoader`'s `init(...)` method? That way we can keep the config-parsing logic separate from the `JSSProvider` itself, and point to `org.mozilla.jss.JSSProvider` as the actual provider in the `java.security` file.

cipherboy:
This was broken on JDK11 as they changed the way Providers were initialized. This introduces two things to fix that:

 - A file to help with the registration of the JSS provider (`META-INF/services/java.security.Provider`)
 - The new Config-based approach to provider loading in JDK9+

cipherboy:
@edewata This includes docs and is now ready for review. Thanks!

cipherboy:
@edewata  said:
> Since the parameters will be loaded from a file, it might be better to convert the doc in JSSProviderLoader into Markdown so end users can easily read it when they're customizing the configuration file.

Sure, I can do that. Most users will probably only need one or two parameters (nss db path and password). And [`InitializationValues`](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/InitializationValues.html) is really the authoritative reference for what these parameters do; this is just a thin mapping over the `Properties` file format into configuring those values. 

@edewata said:
>  Is there a doc/example on how to use JSSProviderLoader? 

Example is [in the tests](https://github.com/cipherboy/jss/blob/provider-loader/cmake/JSSTests.cmake#L427-L431): Pass `-Djava.security.properties=/path/to/java.security` with a [`java.security`](https://github.com/cipherboy/jss/blob/provider-loader/tools/java.security.in#L31-L40) file configured with a [`jss.cfg`](https://github.com/cipherboy/jss/blob/provider-loader/tools/jss.cfg.in). The latter (`jss.cfg`) is the config file that `JSSProviderLoader` reads. 

@edewata said:
> I don't think it should be inheriting from JSSProvider. IIUC it is not a new security provider, but it's only adding the ability to use Properties. So I think JSSProvider itself should support Properties, but it may delegate the functionality to JSSProviderLoader.

So my thoughts on this was we should separate the old way of doing this from the new way. I agree that it isn't a new provider, hence why I had `JSSProviderLoader` inherit from `JSSProvider` -- it is another `JSSProvider`, just extended to be loaded from `java.security`'s `security.provider.<n>` property directly instead of from `CryptoManager`. If I wanted an alternative provider, I would've inherited from `Provider` directly.

In this setup, `JSSProvider` controls the naming (`Mozilla-JSS`), versioning (`4.6.4`), and loading all the algorithms &c that JSS supports. `JSSProviderLoader` handles reading the config file to initialize `CryptoManager`, and then calling into its parent, `JSSProvider` to finish the init. 

In other words, `JSSProvider` is the "real" provider, and all `JSSProviderLoader` does is load the former. My 2c, but I think the separation is nice. 

--- 

The problem with merging `JSSProviderLoader` into `JSSProvider` is that we [break](https://github.com/candlepin/candlepin/blob/master/server/src/main/java/org/candlepin/pki/impl/JSSProviderLoader.java#L143) existing [usages](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/CryptoManager.java#L528) of the `JSSProvider`. It also becomes hard to detect who is calling us: are we having to handle the two-step init of JDK9+? Under JDK <= 8? Someone directly initializing the `JSSProvider`? `CryptoManager` adding us to the providers list? 

In the first two cases, we have to read the configuration file. Under JDK 9+, the no-argument constructor gets called ([see the corresponding comment](https://github.com/cipherboy/jss/blob/provider-loader/org/mozilla/jss/provider/java/security/JSSProviderLoader.java#L90-L94)), and then, if `isConfigured()` returns `false`, `configure("/path/to/config/file")` gets called. Under JDK <= 8, the one-argument String constructor gets called and we load our configuration right then. We have to initialize a CryptoManager before we can do anything else. 

In the latter two cases, we don't want to do any initialization. We just want to update our local Provider and not affect any global state.

The issue with merging them is cases 1 and 3/4 conflict: in 1, we explicitly want to hold off on configuring _any_ provider functionality until we've been configured, but in 3/4, we want to configure the provider. We assume someone else has taken care of `CryptoManager` initialization for us.

We could probably add some detection logic to `JSSProvider`, but IMO, that mixes the concerns of the two.

Thoughts?

edewata:
> So my thoughts on this was we should separate the old way of doing this from the new way. I agree that it isn't a new provider, hence why I had `JSSProviderLoader` inherit from `JSSProvider` -- it is another `JSSProvider`, just extended to be loaded from `java.security`'s `security.provider.<n>` property directly instead of from `CryptoManager`. If I wanted an alternative provider, I would've inherited from `Provider` directly.

It is still a new provider since we can only specify providers in the java.security file:
> security.provider.1=org.mozilla.jss.provider.java.security.JSSProviderLoader @JSS_CFG_PATH@

Anything inheriting from `Provider` either directly or indirectly is a provider. I think it should be like this:
> security.provider.1=org.mozilla.jss.provider.java.security.JSSProvider @JSS_CFG_PATH@

I was hoping we can maintain the separation with something like this (delegation instead of inheritance):
```
public class JSSProvider extends Provider {
    JSSProviderLoader loader = new JSSProviderLoader(this);

    public JSSProvider() {
        this(JAVA_VERSION < 9);
    }

    public JSSProvider(boolean initialize) {
        super("Mozilla-JSS", JSS_VERSION,
                "Provides Signature, Message Digesting, and RNG");

        if (initialize) {
            initializeProvider();
        }
    }

    public JSSProvider(String config_path) throws Exception {
        super(false);
        loader.configure(config_path);
    }

    public JSSProvider(InputStream istream) throws Exception {
        super(false);
        loader.init(istream);
    }

    public JSSProvider(String config_path, InputStream istream) throws Exception {
        super(false);

        if (istream == null) {
            loader.configure(config_path);
        } else {
            loader.init(istream);
        }
    }
}

public class JSSProviderLoader {
    JSSProvider provider;

    public JSSProviderLoader(JSSProvider provider) {
        this.provider = provider;
    }

    public void init(InputStream istream) throws Exception {
        ...
        provider.initializeProvider();
    }
}
```

I'm still trying to understand the situation, but IIUC if anybody is creating provider objects directly, they'd have to deal with the changes in JDK 9+ too, right? Can we expect them to change their code to call `configure()` in JDK 9+, not because we require it, but because the new JDK requires it? Can we delay `CryptoManager` initialization until either the provider is used or the `configure()` is called?

If we really can't update the existing `JSSProvider` to work with JDK 9+, we probably can rename `JSSProviderLoader` to become a new `JSSProvider` under a different package (e.g. `org.mozilla.jss`). So the java.security file will look like this:
> security.provider.1=org.mozilla.jss.JSSProvider @JSS_CFG_PATH@

People that uses the existing `JSSProvider` directly will have to change their code in order to use the new provider.

cipherboy:
@edewata said:
> It is still a new provider since we can only specify providers in the java.security file:
> ```
> security.provider.1=org.mozilla.jss.provider.java.security.JSSProviderLoader @JSS_CFG_PATH@
> ```
> 
> Anything inheriting from Provider either directly or indirectly is a provider. I think it should be like this:
> ```
> security.provider.1=org.mozilla.jss.provider.java.security.JSSProvider @JSS_CFG_PATH@
> ```

I don't think this is quite correct. The [`Provider`](https://docs.oracle.com/javase/8/docs/api/java/security/Provider.html) interface provides a way of looking up which algorithms a given provider supports. However, the actual _instance_ doesn't matter -- what matters is the _name_ field. In particular, the [`Security`](https://docs.oracle.com/javase/8/docs/api/java/security/Security.html) interface only really allows one Provider with a given name (as reported by a call to `Provider(String name, ...)` -- not the class name). 

---

If you specify the following in the `java.security` file:

```
security.provider.1=org.mozilla.jss.provider.java.security.JSSProviderLoader @JSS_CFG_PATH@
security.provider.2=org.mozilla.jss.JSSProvider
```

This will add the `Mozilla-JSS` provider twice via two different classes. (This is safe because the first instance already initializes the CryptoManager and so it won't be). 

The output from listing shows:

```
 - Mozilla-JSS | Mozilla-JSS version 4.6.3 | 1768305536
 - Mozilla-JSS | Mozilla-JSS version 4.6.3 | 1530388690
```

(The last is the system object identifier -- showing that they're fundamentally different instances.)

However, when you go to query it, there's no way to distinguish the two. They're expected to be interchangable. Consider that `System.getProvider(String name)` -- only returns a single instance. They're only expecting a single provider by that given name!

In particular, if you call `System.getProvider(String name)` or `Mac.getInstance("HmacSHA512", "Mozilla-JSS").getProvider()` -- it always returns the _first_ provider. In particular, if you make `JSSProvider` the first provider (and pass `initialize=false` so it doesn't advertise any algorithms) -- it'll fail to find any matching providers for `Mac.getInstance("HmacSHA512", "Mozilla-JSS").getProvider()`! This is because -- while the second `Mozilla-JSS` instance from JSSProviderLoader does advertise this algorithm -- the first instance is the only one found and it doesn't advertise anything!

---

@edewata said:
> If we really can't update the existing JSSProvider to work with JDK 9+, we probably can rename JSSProviderLoader to become a new JSSProvider under a different package (e.g. org.mozilla.jss). So the java.security file will look like this:

From my perspective, it doesn't matter what the inheritance tree looks like here -- IMO, `JSSProvider` **is** the `Mozilla-JSS` provider. I'd rather have `JSSProviderLoader` inherit from `JSSProvider` (if the former is going to be added to `java.security`) than have it be a completely separate Provider implementation. That shows that it is fundamentally the same provider (to users), just with a different class name to make it clear that it can be used from `java.security` on all JDK releases.

---

Now, we could make `JSSProviderLoader` be a separate class and have it behave like you mentioned above.

I think the condition here needs to be a little different:

```java
public JSSProvider() {
    this(JAVA_VERSION < 9);
}
```

I think it should be:

```java
public JSSProvider() {
    boolean initialize = false;
    try {
        initialize = (CryptoManager.getInstance() != null);
    } catch (Exception e) { } 

    this(initialize);
}
```

This handles:

 - JDK8 vs JDK9: JDK8 calls the one-argument constructor with the `String` config path when one is given. Otherwise it calls the no-argument constructor and Mozilla-JSS will be useless, not implementing any algorithms. JDK9 calls the no-argument constructor always, checks the value of `isConfigured(...)` and then calls `configure(...)` if an argument is given. Again, `Mozilla-JSS` is useless until configured.
    - This means that any explicit request to Mozilla-JSS will fail.
    - If they wish to call `configure(...)` afterwords, JDK8 people can cast and JDK9 people should be expecting it now.
 - A direct construction after `CryptoManager` is initialized does the right thing.
    - This means the `CryptoManager` property to initialize JSS behaves as expected.
    - Candlepin isn't affected in their current usage, but they can move to a new version.

So this is probably the path I'd take. Thoughts?

cipherboy:
> I'm still trying to understand the situation, but IIUC if anybody is creating provider objects directly, they'd have to deal with the changes in JDK 9+ too, right? Can we expect them to change their code to call configure() in JDK 9+, not because we require it, but because the new JDK requires it? Can we delay CryptoManager initialization until either the provider is used or the configure() is called?

To clarify this: I think `CryptoManager` should still be able to be initialized separately. We can delay until `configure(...)` is used theoretically, but I'm not sure what benefit that gives us. We have no way of knowing when the provider is "used" -- in particular, NSS initialization isn't cheap (though it only happens once), so we'd prefer it to happen before the app needs to know about it, i.e., very close to JVM startup. In order to tell when the provider is used, we have to add hooks _everywhere_ -- to every single algorithm we implement -- and check if CryptoManager is initialized and then initialize it if not.

That seems risky.

cipherboy:
@edewata This is updated. We're waiting on 3.50.0-2 hitting F30/F31 before tests are unblocked anyways. But it should use the method you proposed in your discussions.

PR #422, opened 2020-03-06
Move CryptoManager logging to debug

cipherboy:
When loading JSS as part of the default JDK setup, we don't really wish
to spam unsuspecting users with log messages. Move `logger.info` to
`logger.debug` in the `CryptoManager` initialization process.

----

@edewata / @jmagne / @ladycfu -- are there any arguments against this? It makes JSS initialization silent in most cases, which for putting it in `java.security`, I think is a good thing.

### DIFF
--- a/org/mozilla/jss/CryptoManager.java
+++ b/org/mozilla/jss/CryptoManager.java
@@ -53,21 +53,21 @@ public final class CryptoManager implements TokenSupplier
 
     static {
 
-        logger.info("CryptoManager: loading JSS library");
+        logger.debug("CryptoManager: loading JSS library");
 
         try {
             System.loadLibrary("jss4");
-            logger.info("CryptoManager: loaded JSS library from java.library.path");
+            logger.debug("CryptoManager: loaded JSS library from java.library.path");
 
         } catch (UnsatisfiedLinkError e) {
 
             try {
                 System.load("/usr/lib64/jss/libjss4.so");
-                logger.info("CryptoManager: loaded JSS library from /usr/lib64/jss/libjss4.so");
+                logger.debug("CryptoManager: loaded JSS library from /usr/lib64/jss/libjss4.so");
 
             } catch (UnsatisfiedLinkError e1) {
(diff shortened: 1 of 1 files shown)


cipherboy:
@edewata Yes:

```java
$ cat Sample.java
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public class Sample {
    public static Logger logger = LoggerFactory.getLogger(Sample.class);

    public static void main(String[] args) {
        System.out.println("Debug: " + logger.isDebugEnabled());
        System.out.println("Error: " + logger.isErrorEnabled());
        System.out.println("Info: " + logger.isInfoEnabled());
        System.out.println("Trace: " + logger.isTraceEnabled());
        System.out.println("Warn: " + logger.isWarnEnabled());
        logger.debug("DEBUG");
        logger.info("INFO");
        logger.trace("TRACE");
        logger.warn("WARN");
        logger.error("ERROR");
    }
}

```bash
$ cat main.sh
#!/bin/bash

classpath="/usr/share/java/slf4j/api.jar:/usr/share/java/slf4j/jdk14.jar:."

javac -cp "$classpath" Sample.java && java -cp "$classpath" Sample
$ ./main.sh
Debug: false
Error: true
Info: true
Trace: false
Warn: true
Mar 06, 2020 10:48:38 AM Sample main
INFO: INFO
Mar 06, 2020 10:48:38 AM Sample main
WARNING: WARN
Mar 06, 2020 10:48:38 AM Sample main
SEVERE: ERROR
```

edewata:
Thanks. It might be nice to show one line of `INFO` message just to indicate that JSS is running, but it's up to you.

ladycfu:
the idea makes sense.  I also like edewata's suggestion on showing one INFO message.

cipherboy:
Thanks @edewata and @ladycfu!

I've moved the info message to after JSS is fully initialized to be more of "this is here" versus a "this will be here if nothing goes wrong".

PR #425, opened 2020-03-12
Detect NSS with broken CMACs during configuration

cipherboy:
NSS versions v3.47 to v3.50 included swapped values for `CKM_AES_CMAC` and
`CKM_AES_CMAC_GENERAL`. This adds feature detection to JSS, disabling CMAC
and KBKDF at compile time for the broken NSS versions.

Related: [`moz-bz#1611209`](https://bugzilla.mozilla.org/show_bug.cgi?id=1611209)

---

Note that this doesn't add runtime detection. I'm still debating if that's worth adding, and what our response should be. Runtime detection would have to actually call the CMAC operation, which would likely involve generating a new key. In FIPS mode, that likely requires the token to be logged in already.

### DIFF
--- a/cmake/JSSConfig.cmake
+++ b/cmake/JSSConfig.cmake
@@ -345,4 +345,20 @@ macro(jss_config_symbols)
     if(NOT HAVE_NSS_KBKDF)
         message(WARNING "Your NSS version doesn't support NIST SP800-108 KBKDF; some features of JSS won't work.")
     endif()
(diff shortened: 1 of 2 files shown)


cipherboy:
`fedora_rawhide` shows a warning, but it isn't clear why; the checks passed. `fedora_sandbox` also passed, as does F31 and F30 now. Merging.

PR #428, opened 2020-03-13
Add documentation about JSSEngine usage and design

cipherboy:

---

This complements PR #150 and adds usage and design documentation about the SSLEngine.

This is still WIP because its missing documentation on:

 - [x] Configuring JSSEngine
 - [x] Non-blocking configuration (Buffer PRFD)
 - [x] Performance issues
 - [x] Handshake Theory
 - [x] PHA/rehandshake
 - [x] Large wrap/unwraps. 

View the documentation [here](https://github.com/cipherboy/jss/blob/jss-engine-docs/docs/usage/jssengine.md).

### DIFF
--- a/docs/usage/jssengine.md
+++ b/docs/usage/jssengine.md
@@ -17,6 +17,12 @@ There are two ways to use the `JSSEngine`: via the JCA Provider interface,
 or constructing a `JSSEngine` instance directly. The former method is
 preferred out of the two.
 
+Please refer to the [`JSSEngine` test cases][jss.jssengine.tests] or to
+the [Oracle `SSLEngine` demo][oracle.sslengine.demo] for information on
+using it. Usually a `SSLEngine` instance is asked by an external API
+(e.g., Tomcat) or used inside a `SSLSocket` implementation, such as the
+`SSLSocket` implementation provided by `Mozilla-JSS`'s SSLContext`.
+
 
 ### Via the JSSProvider
 
@@ -199,6 +205,10 @@ two important introductions:
     allows us to tie into NSS's hostname verification directly, instead of
     responding after the fact by closing the connection.
 
+Generally, using `SSLParameters` should be sufficient for most applications.
+Two exceptions are when we wish to explicitly select key material (e.g., from
+a certificate nickname) or when using NSS for SSL hostname validation.
+
 #### Session Control
 
 The `JSSEngine` lacks many of the session control functions other `SSLEngine`
@@ -237,8 +247,8 @@ We expect two primary implementations:
 
  - `JSSEngineReferenceImpl`, a reference implementation with more logging
    and debugging statements. This also includes port-based debugging, so
-   situations where a `SSLEngine` is without writing to the network can
-   still be tracked and analyzed in Wireshark. Each call to `wrap` or `unwrap`
+   situations where a `SSLEngine` isn't writing to the network can still
+   be tracked and analyzed in Wireshark. Each call to `wrap` or `unwrap`
    makes several JNI calls, incurring lots of overhead.
  - `JSSEngineOptimizedImpl`, an optimized, production-ready implementation.
    This one is harder to debug due to fewer logging statements, but does
@@ -246,6 +256,14 @@ We expect two primary implementations:
 
 
 ### Non-Blocking IO
+
+In order to implement `wrap` and `unwrap` on the SSLEngine, we use NSPR
+sockets configured in non-blocking mode. Data is held in buffers rather
+than being written to the network; in this way, data is passed from a
+`unwrap` call to `NSPR` so that `NSS` can decrypt the wire data, with
+the result being returned from `unwrap`.
+
+This is implemented as follows:
 
 NSPR introduces a platform-independent abstraction over C's file descriptors
 (usually an `int`) in the form of the `PRFileDesc` structure. This is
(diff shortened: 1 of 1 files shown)


cipherboy:
@edewata  / @jmagne This is ready for review.

docs/usage/jssengine.md
```diff
@@ -0,0 +1,521 @@
...
+contract for non-blocking TLS. Unlike the `SunJSSE` provider's `SSLEngine`
+(when using the `SunPKCS11-NSS` provider for primitives), this is built
+directly on NSS's high-level SSL module. This is better for FIPS compliance
+and HSM support (as keys never leave the NSS cryptographic module) and also
+means that we don't need to reimplement the underlying state machine. This
+approach is consistent with our [JSS `SSLSocket`][jss.ssl-socket].
+
+
+## Using `JSSEngine`
+
```
jmagne, line 15: This is great so far, but below it talks about the various ways to create an SSLEngine and configure it. Would it make any sense, in your opinion to put a little more up top as to why one would create one of these things to do what? Maybe a skeleton example cllient / server program. Of course to be fair, if someone wants to use this thing, they may possibly know already... Right now it just briefly describes that it is implements the contract for non-blocking TLS. Just a thought.

cipherboy: An interesting idea, yes. Outside of test cases, I'm not sure most people will use a SSLEngine directly for client/server handshakes. There's a number of higher-level APIs (Tomcat, Apache products, and some internal to the JDK) that take a SSLEngine instance and use it to do, say, TLS for HTTPS or LDAPS. In my opinion, most people probably won't be using SSLEngine directly to do client/server ops and just need to "configure" it to pass it to the relevant library/...

Should I just provide a link to the [test cases](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/tests/TestSSLEngine.java) where this is done? Otherwise, there is an [excellent Oracle example](https://docs.oracle.com/javase/8/docs/technotes/guides/security/jsse/samples/sslengine/SSLEngineSimpleDemo.java) that explains this as well.

jmagne: Yeah, the links sound good, but maybe just put a blurb or two explaining what this thing is for and how mostly it will be associated with higher level API's, just like you said and that should do it. Otherwise I"m just about finishing reading it and it looks great so far.

docs/usage/jssengine.md
```diff
@@ -0,0 +1,521 @@
...
+// SSLEngine inst;
+inst.setEnabledCipherSuites(new String[] { "TLS_AES_128_GCM_SHA256" });
+```
+
+will enable only a single TLSv1.3 cipher suite.
+
+Alternative methods are available that take JSS standard `SSLCipher`
+values as parameters; see the `JSSEngine` javadoc for more information.
+
+#### Using `JSSParameters`
```
jmagne, line 191: This section might be a little confusing to those initiated. Here is just says it's mostly like SSLParameters but with added properties. Would one know how and when to use these instead of the SSLParameters? Also not those are mentioned indirectly when talking about  "SSLContext.getDefaultSSLParameters() and SSLContext.getSupportedSSLParameters()".  Only asking...

jmagne: I just noticed that there is a reference to the JSSParameters separate man page.

docs/usage/jssengine.md
```diff
@@ -0,0 +1,521 @@
...
+`JSSEngine` is an abstract class extending [`SSLEngine`][javax.ssl-engine].
+This class implements some of the boilerplate required for implementing a
+`SSLEngine`, including handling cipher and protocol version configuration.
+Individual implementations implement `wrap`, `unwrap`, and whatever specifics
+are necessary to initialize and release the SSL-backed `PRFileDesc`.
+
+We expect two primary implementations:
+
+ - `JSSEngineReferenceImpl`, a reference implementation with more logging
+   and debugging statements. This also includes port-based debugging, so
```
jmagne, line 239: Working on the following sentence seems a bit off to me maybe just a typo.

jmagne: "is without writing to the network"

docs/usage/jssengine.md
```diff
@@ -0,0 +1,521 @@
...
+implementation in `org/mozilla/jss/ssl/javax`. Each `BufferPRFD` is backed
+by two `j_buffer` instances (located adjacent). A `j_buffer` is a circular
+ring buffer optimized to allow efficient access to the underlying data. One
+`j_buffer` is dedicated to the read end of this `BufferPRFD` (the data that
+is returned when `recv(...)` is called); the other is dedicated to the write
+end of this `BufferPRFD` (the data that is waiting to get sent on the wire).
+Note that the `j_buffer` instances exist independently from the `BufferPRFD`:
+the `JSSEngine` itself creates the `j_buffer`s and hands them to the
+`BufferPRFD` to use, because it too needs to be able to read from them.
+
```
jmagne, line 265: I found it interesting when looking at the code that the NSPR sockets were being used to wrap or  unwrap locally and then handing the data off to wrap or unwrap. Would a simple statement to that at a high level make the whole thing a bit clearer to the reader? I know that all that gets explained below at the nitty gritty level.

docs/usage/jssengine.md
cipherboy: OK clarified. `JSSParameters` inherits from `SSLParameters`; the latter is in the standard library and should be used by most everyone, but if you need something only `JSSParameters` provides (or wish to re-use existing JSS classes such as the `SSLCipher` enums) it is better to use `JSSParameters`.

docs/usage/jssengine.md
cipherboy: Yeah, good catch. Thanks. :)

docs/usage/jssengine.md
cipherboy: OK, I think I've added a clarifying section at the beginning of all this nitty-gritty stuff. :)

cipherboy:
I'm going to go ahead and merge this now. Worst case we can always update the documentation later. :)

PR #430, opened 2020-03-17
Remove token key checks

cipherboy:
Previously we enforced strict token key matching: the primary key used
for the operation must strictly reside on the current PKCS#11 token,
otherwise JSS would bail. However, NSS has the ability to move the key
to whichever token best supports the given operation. This means that
we'd prematurely bail when the operation would succeed if it were
actually executed. By removing these checks, we still leave the ability
to generate keys on a specific token, we just allow them to be used on
whatever token supports the given operation (and the key is allowed to
be moved to).


cipherboy:
Because 58d68c8faf09b55414c1b6ba2a6df9b979bd8d85 introduced a `RuntimeException` in all cases where we previously would've ignored it (in production mode; debug mode would've thrown an exception), this is technically a regression from `v4.4.x`/RHEL 7 behavior.

PR #432, opened 2020-03-20
Enable SSLSession - Query GetChannelInfo / GetPreliminaryChannelInfo / InvalidateSession

cipherboy:
This PR adds most of the information for `SSLSession` stuff:

 - A bunch of new enums for types from NSS (`SSLKEAType`, `SSLNamedGroup`, `SSLAuthType`, `SSLCipherAlgorithm`, `SSLCompressionMethod`, `SSLMACAlgorithm`, and `SSLSignatureScheme`). 
 - Adds wrappers for the two return structs from NSS: `SSLChannelInfo` and `SSLPreliminaryChannelInfo`.
 - Adds the getters into `org.mozilla.jss.nss.SSL`: `SSL.GetChannelInfo` and `SSL.GetPreliminaryChannelInfo`.
 - Adds the `SSL.InvalidateSession` helper to allow the server to invalidate client sessions when necessary. 
 - Adds a basic functionality test to make sure everything is wired up correctly. 

When #150 and this PR are both merged, a new PR  (#434) will be opened with enables `SSLSession` in the existing `JSSEngine` implementations.

PR #434, opened 2020-03-20
Enable SSLSession

cipherboy:
After ~#150~ and ~#432~ are merged, this can be rebased, reviewed, and merged.

This enables `SSLSession` in three commits:
 - Moves the `SSLFDProxy` from the specific instances to the `JSSEngine` interface. Also exposes access to it in a coordinated way. (`ssl_fd` is nullable, so callers should always use `getSSLFDProxy()` instead of caching the value).
 - Updates `JSSSession` with new values, sets the values from `JSSEngineReferenceImpl`, and adds some NSS-specific tricks.
 - Adds some basic tests for SSLSession to the test suite. 

This `SSLSession` implementation is definitely advanced enough for both JSS-specific callers and also generic enough to support Tomcat.

### DIFF
--- a/org/mozilla/jss/ssl/javax/JSSEngine.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngine.java
@@ -145,6 +145,12 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
     protected JSSSession session;
 
     /**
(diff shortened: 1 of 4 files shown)


cipherboy:
Ok, @edewata this is updated.

PR #435, opened 2020-03-23
Move to proper feature checks

cipherboy:
`SSLCipher` was the first use of a feature which could appear in a later
version of NSS than we support. Rather than bumping the minimum NSS
version, we chose to use compile-time detection of the NSS version and
limit our code accordingly. However, this sets a precedence for ignoring
the features actually present in the NSS system. Certain downstream
distributions are fond of backporting features, which means our code
could've executed but didn't.

Switching to feature detection (via the `check_struct_has_member` macro)
allows us to be sure we execute this code on as many platforms as
possible.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build/
 
 # These files are automatically generated from their .in equivalents
 org/mozilla/jss/util/jssver.h
+org/mozilla/jss/jssconfig.h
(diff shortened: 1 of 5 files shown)


cipherboy:
@edewata I'm wanting to merge this before #432 and then use it for the fields there instead of version checks as well.

cipherboy:
Thanks @edewata!

PR #440, opened 2020-03-24
Add feature detection for SSLCipherInfo.peerDelegCred

cipherboy:
While trying to add CentOS 8-based CI images, I noticed this field wasn't compiling correctly. In particular, NSS v3.44 claimed not to know about this field, despite it supposing to exist in v3.34. 

Add feature detection to it in CMake, fixing up the docs &c. accordingly.


cipherboy:
@edewata I think I'm ready to discuss alternative ways of doing this now. Do you have any suggestions? :-)

cipherboy:
Yes, we can remove the `haveNSSXXX` fields or we could leave them (in the struct) and set them to true (and remove them from CMake / the constructor / the native code / ...)  --- this doesn't break others and we can add a `@Deprecated` warning.

cipherboy:
I'm merging this for now. Until we release v4.6.4, I think we have time to agree on a way of doing this. But it lets us move forward with other PRs in the meantime.

PR #441, opened 2020-03-24
Add CentOS-based CI

cipherboy:
~This will be rebased on top of #440 once that is reviewed and merged.~

----

This adds CentOS 7 and 8 CI to the master branch of JSS, under the optional section. This means it isn't required for merge and if necessary, we can break it. However, adding both CentOS 7 and CentOS 8 should give us a wider range of NSS versions to test on, which I think is a good thing for detecting when we truly need to bump the minimum NSS version in the spec file.

### DIFF
--- a/.github/workflows/optional.yml
+++ b/.github/workflows/optional.yml
@@ -51,3 +51,17 @@ jobs:
       uses: actions/checkout@v2
     - name: Build and Run the Docker Image
       run: bash tools/run_container.sh "fedora_sandbox" || echo "::warning ::Job exited with status $?"
+  centos7:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Clone the repository
+      uses: actions/checkout@v2
+    - name: Build and Run the Docker Image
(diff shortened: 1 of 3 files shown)


SilleBille:
This patch targets master (v4.6) which is delivered on RHEL8.x. Isn't it enough to test against CentOS8? IIUC, CentOS7 mimics RHEL7 and quite soon or later, the builds are gonna fail due to missing deps. I understand that it won't affect commits or PRs in anyway. But, just a thought on the maintenance (for instance, the `git -C`) :-)

cipherboy:
I could drop the `build.sh` patch if that's preferable. It isn't actually used for RHEL 7 because we have to add EPEL and pull in CMake3. The latter gets the executable name `cmake3`, which means we can't just build it from the `rpm.spec`.

cipherboy:
> the builds are gonna fail due to missing deps

I don't think so. There's actually very few deps that we have for JSS. And mostly we want to know when we need to bump the minimum NSS version. Also, EPEL is a wonderful thing we can use in CentOS and upstream CI but can't be used downstream in RHEL ;-)

PR #442, opened 2020-03-25
Add support for versioned features

cipherboy:
This is one approach to resolve some of the discussion in #432 and #440.

Using `haveNSS<version>` was undesirable for two reasons:

 - As we found out, sometimes these versions are wrong (see #440).
 - Checking a parallel property (versus a property of the value itself) is undesirable.

This allows us to be clear about what the minimum NSS version is to use this feature (+/- OS's backporting is), and should improve clarity in all code which uses this. 

In particular, our interface now wraps values, allowing them to be checked (`haveFeature(...)`) before being accessed (`getValue(...)`) -- with a sane default (`null`) in the case where this feature isn't present.

The one downside is that each value must be individually checked -- in `SSLPreliminaryChannelInfo` we have 3 values with the same NSS version guard that we now much wrap separately. I think this is a fine trade off.

### DIFF
--- a/org/mozilla/jss/nss/SSLChannelInfo.java
+++ b/org/mozilla/jss/nss/SSLChannelInfo.java
@@ -2,8 +2,9 @@ package org.mozilla.jss.nss;
 
 import java.lang.StringBuilder;
 
-import org.mozilla.jss.ssl.*;
 import org.mozilla.jss.netscape.security.util.Utils;
+import org.mozilla.jss.ssl.*;
+import org.mozilla.jss.util.VersionedFeature;
 
 /**
  * Class representing the SSLChannelInfo struct from NSS's sslt.h.
@@ -174,22 +175,6 @@ public class SSLChannelInfo {
     public SSLSignatureScheme signatureScheme;
 
     /**
-     * This field controls whether or not we have the following three fields:
-     *
-     *  - originalKeaGroup,
-     *  - resumed, and
-     *  - peerDelegCred.
-     *
-     * When this field is true, the values of these fields can be trusted.
-     * Otherwise, their values should be ignored.
-     *
-     * The corresponding fields are present when the NSS version used to
-     * compile JSS and the runtime version of NSS match, and both have these
-     * fields.
-     */
-    public boolean haveNSS334;
-
-    /**
      * This field holds the key exchange algorithm group during the initial
      * handshake.
      *
@@ -200,21 +185,21 @@ public class SSLChannelInfo {
      * When the session was resumed this holds the key exchange group of the
      * original handshake.
      */
-    public SSLNamedGroup originalKeaGroup;
+    public VersionedFeature<SSLNamedGroup> originalKeaGroup;
 
     /**
      * Whether or not this session was resumed.
      *
      * This field was added in NSS 3.34.
      */
-    public boolean resumed;
+    public VersionedFeature<Boolean> resumed;
 
     /**
      * Whether or not the peer used a delegated credential for authentication.
      *
      * This field was added in NSS 3.34.
      */
-    public boolean peerDelegCred;
+    public VersionedFeature<Boolean> peerDelegCred;
 
     /**
      * Constructor used by SSL.GetChannelInfo(...).
@@ -257,12 +242,15 @@ public class SSLChannelInfo {
         this.authType = SSLAuthType.valueOf(authType);
         this.signatureScheme = SSLSignatureScheme.valueOf(signatureScheme);
 
-        this.haveNSS334 = haveNSS334;
+        this.originalKeaGroup = new VersionedFeature<SSLNamedGroup>("3.34");
+        this.resumed = new VersionedFeature<Boolean>("3.34");
+        this.peerDelegCred = new VersionedFeature<Boolean>("3.34");
 
-        this.originalKeaGroup = SSLNamedGroup.valueOf(originalKeaGroup);
-        this.resumed = resumed;
-
-        this.peerDelegCred = peerDelegCred;
+        if (haveNSS334) {
+            this.originalKeaGroup.setValue(SSLNamedGroup.valueOf(originalKeaGroup));
+            this.resumed.setValue(resumed);
+            this.peerDelegCred.setValue(peerDelegCred);
+        }
     }
 
     /**
@@ -287,9 +275,13 @@ public class SSLChannelInfo {
         result.append("\n- macAlgorithm: " + macAlgorithm);
         result.append("\n- authType: " + authType);
         result.append("\n- signatureScheme: " + signatureScheme);
-        if (haveNSS334) {
+        if (originalKeaGroup.haveFeature()) {
             result.append("\n- originalKeaGroup: " + originalKeaGroup);
+        }
+        if (resumed.haveFeature()) {
             result.append("\n- resumed: " + resumed);
+        }
+        if (peerDelegCred.haveFeature()) {
             result.append("\n- peerDelegCred: " + peerDelegCred);
         }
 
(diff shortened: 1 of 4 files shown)


cipherboy:
@edewata If you like this approach, I'll update #440 to use it.

cipherboy:
Notably, this removes the explicit fields called "version", but make it harder for us to drop version checks later.

edewata:
I'm not sure we really need a `VersionedFeature` class for this. When do we use the `present` and `version` fields?

How about using `null` to indicate that the feature is not supported?
```
public class SSLChannelInfo {
    SSLNamedGroup originalKeaGroup;
    Boolean resumed;
    Boolean peerDelegCred;

    public String toString() {
        if (originalKeaGroup != null) {
            result.append("\n- originalKeaGroup: " + originalKeaGroup);
        }
    }
}
```
The caller will only set those fields if the feature is supported.
```
SSLChannelInfo channelInfo = new SSLChannelInfo(...);
if (haveNSS334) {
    channelInfo.setOriginalKeaGroup(originalKeaGroup);
    channelInfo.setResumed(resumed);
    channelInfo.setPeerDelegCred(peerDelegCred);
}
```

cipherboy:
We'd use the `present` field to indicate this version of NSS (at runtime and compile time) supports this operation.

Consider if we used this for say, KBKDF (its a bad example for other reasons, but let's roll with it):

```java
class KBKDFAlgorithm {
    VersionedFeature<KBKDF> feature = VersionedFeature<KBKDF>("v3.51");

    doKBKDF() {
        if (!feature.haveFeature()) {
                throw new RuntimeException("In order to use this feature, compile with NSS version " + feature.getMinimumVersion() + " and use it at runtime!");
        }

        feature.getValue().doKBKDF();
    }
}
```

The reason why we need something like this (whether an interface/storage class, like in the `VersionedFeature<T>` case or a `boolean` alongside the value as we currently do) is to represent the _dimensionality_ of this particular data/feature/...:


 - There's one dimension is "does this field/operation/... have a value/work/ &c" -- that's whether or not the value is `null`.
 - There's another dimension which is "does the current JSS support this current field/operation/...".

My premise is the two dimensions are orthogonal. If you conflate the two dimensions (and use `null` for both), you can't adequately detect "something went wrong reading this value" from "my current setup will never give me a working value -- because JSS needs to be rebuilt or NSS needs to be updated".

Ideally you'd like the former case (something went wrong) to raise an exception (though, where this is used, we're really close to the NSS/Java interconnect layer and so error handling could really be on either side here and we could raise an exception) -- but in the latter case, you'd like to be able to say "eh, that's fine" or "no, I really want that value".

In the KBKDF case, there's clear intent saying "I need this value" -- and so you want to error. In the SSLSupport case, where this is optional data likely never read by 90% of applications that we're still exposing from NSS because we need _some_ of this data anyways... we're fine handing this to the calling application to deal with as they see fit. Somebody might need this. But most people wont, so there's no point throwing an exception about it. And those that do, can conform to the interface.

My 2c. 

That said, if you have some better idea, I'm all ears :-)

cipherboy:
The KBKDF is a good example for the reason that we need a way of detecting _if_ NSS supports it, while at the same time, not requiring it. Because it isn't a widely used feature, we don't need to panic if NSS doesn't support it yet (Debian has already upgraded to a new enough JSS that supports KBKDF but hasn't upgraded NSS to v3.51 as that was just released at the start of this month). But if someone tries to use it and fails, we do need to tell them why.

edewata:
To my understanding we're adding feature detection since the NSS version number is not a reliable indicator, so I don't think we should provide a `feature.getMinimumVersion()` since it could be misleading.

If `null` is a valid value of the feature, then yes, we need to use another field (i.e. `present`) to keep track of the feature's availability. Since these are NSS features (i.e. C code), I was hoping all valid values will be non-null so we can use the `null` itself to indicate when the feature is unavailable. If something went wrong, we cannot really assume that the caller will always check the return value and take the appropriate action. Instead, I think we should generate an exception that interrupts the execution to prevent further errors. The exception itself should include a user-friendly error message to help troubleshooting, and also include the return value/error code so the error can be handled programmatically if needed.

My preference is to raise the minimum NSS minor version as high as possible, and designate a JSS minor version for that. For example, JSS 4.6 will require NSS 3.31 which should cover all platforms that we currently support. In JSS 4.6 the KBKDF will be an optional feature, so it needs feature detection. Later we'll have JSS 4.7 which requires NSS 3.51. In that case KBKDF is guaranteed to exist, so we don't need feature detection anymore. I think this would make it easier for users to determine which JSS and NSS versions that they need instead of only finding out that they don't have the required features at runtime. If we could use `null` to indicate unavailability of a feature, it would provide a smoother transition between JSS versions.

cipherboy:
> To my understanding we're adding feature detection since the NSS version number is not a reliable indicator

Its an expected lower bound. It is reliable in that, that is the minimum NSS version that guarantees this feature exists. Consider KBKDF: no version before 3.51 has (a working) KBKDF. But _all_ NSS 3.51 and after have this feature. People don't care what the minimum NSS version is when they have it -- they care when they don't have it. And having a separate boolean `present` versus a string `version` field lets them use the former for detection and the latter only for displaying an error message.

And you (as a user) know whether you fall into the "more likely to backport" or "more likely to upgrade" -- and the small subset of users that tend to backport know where to find commits from NSS and grab them individually. That's the state of KBKDF in downstream actually: we took n-1's NSS release (3.44? -- released before we even started CMAC much less KBKDF) and added the CMAC and KBKDF patches. To me, this shows why version detection (at build/release time) isn't sufficient and but also why this needs to be worded to be a "guaranteed" version to have it.

---

> Since these are NSS features (i.e. C code), I was hoping all valid values will be non-null so we can use the null itself to indicate when the feature is unavailable.

I don't think this is true in general. In particular, if you call `SSL.GetChannelInfo()` too early, it succeeds (as in, returns `SECSuccess`), but none of the fields have any meaning, except where they're initialized as such early on:

```java
SSLChannelInfo:
- protocolVersion: null                         <-- HERE, not version locked.
- cipherSuite: TLS_NULL_WITH_NULL_NULL
- authKeyBits: 0
- keaKeyBits: 0
- creationTime: 0
- lastAccessTime: 0
- expirationTime: 0
- sessionID:                                                  <-- HERE, not version locked.
- compressionMethod: NULL
- extendedMasterSecretUsed: false
- earlyDataAccepted: false
- keaType: ssl_kea_null
- keaGroup: null                                             <-- HERE, not version locked.
- symCipher: ssl_calg_null
- macAlgorithm: ssl_mac_null
- authType: ssl_auth_null
- signatureScheme: ssl_sig_none
- originalKeaGroup: null                         <-- HERE, version locked, but NSS > 3.34 provided.
- resumed: false
- peerDelegCred: false
```

---

> My preference is to raise the minimum NSS minor version as high as possible, and designate a JSS minor version for that.

Youchie! I think that's fine for an end application, but not fine for a library. When would we ship JSSv4.7? At this rate of NSS updates, probably not until most of the way through the RHEL 8 cycle! While we have a need for KBKDF (eventually -- but we've satisfied that on RHEL / Fedora), not everyone needs KBKDF. And that's the only major version-locked feature.

These are fields in a struct that grows over time. We can either ignore the new fields (and not report them) or report them for those that want them. They don't warrant bumping the minimum NSS version. 

Additionally, it'd be nice to maintain as much compatibility as we can. We do sometimes break APIs (v4.4 vs v4.5 especially), but for the most part, some of the changes we've made in v4.6 are nice for, e.g., debugging -- when possible, having those available on older platforms (even if not by default) would be helpful. :-)

My 2c tho.

cipherboy:
To be clear, this doesn't provide runtime checking of features (nor does #440 or anything else) -- this is strictly compile time checking.

In particular if `sizeof(SSLContextInfo)` at runtime differs from `sizeof(SSLContextInfo)` we used at compile time, all of our existing code will fail anyways. This will do the same.

Ideally, I'm thinking we'll move the Compile/Runtime check to sometime late in `CryptoManager` initialization and have it fail if this is broken.

And to be clear, this is a problem with this particular NSS interface. Other NSS interfaces don't have these complex versioning problems. Mostly they're opaque structs.

cipherboy:
I have an alternative idea. Let me try that out.

PR #443, opened 2020-03-25
Add documentation on the Mozilla-JSS Provider

cipherboy:

This begins to document our changes w.r.t. `JSSProviderLoader` &c. 

You can see the rendered documentation [here](https://github.com/cipherboy/jss/blob/add-provider-loader-docs/docs/usage/jssprovider.md).

### DIFF
--- a/docs/usage/jssprovider.md
+++ b/docs/usage/jssprovider.md
@@ -1,0 +1,166 @@
+# `Mozilla-JSS` Provider
+
+The `Mozilla-JSS` JCA-compatible Provider exposes most of the functionality
+of JSS to external packages. This interface is the recommend interface most
+developers should build against. However, once the dependencies are satisfied
+and JSS's native component is available to the JVM, we still have to load
+and initialize JSS.
+
+There are two routes to do this:
+
+ 1. Via the `CryptoManager` interface, _and_
+ 2. Via `java.security`, directly loading the `JSSProvider`.
+
+
+## Loading JSS via `CryptoManager`
+
+To load JSS from a `CryptoManager`, it is necessary to decide what level of
+configuration is necessary. If you're happy with the defaults, it is
+sufficient to only specify a NSS DB:
+
+```java
+import org.mozilla.jss.CryptoManager;
+
+CryptoManager.initialize("/path/to/nss-db");
+```
+
+At this point, JSS will be initialized and can be used. A password might be
+required, so see the section below for providing password callback handlers.
+
+Certain default values might not work in all situations. For instance,
+Candlepin expects other providers to be default, with the Mozilla-JSS
+provider being the least-preferred provider. To do this, they'd construct an
+`InitializationValues` instance and pass that to
+`CryptoManager.initialize(...)`:
+
+```java
+import org.mozilla.jss.InitializationValues;
+
+InitializationValues ivs = new InitializationValues("/path/to/nss-db");
+ivs.installJSSProviderFirst = false;
+
+CryptoManager.initialize(ivs);
+```
+
+See the section below on other options available to configure.
+
+
+### `InitializationValues` options
+
+There are two constructors for `InitializationValues`:
+
+ - Taking only a NSS DB directory. This was utilized above.
+ - Taking a NSS DB directory, a prefix for the certificate database, a
+   prefix for the key database, and the name of the secmod configuration.
+
+The latter is a more advanced use case and few individuals likely need to use
+it. For more information, see the corresponding NSS documentation on these
+values.
+
+Refer to the javadoc for `InitializationValues` for the supported parameters
+and their default values.
+
+### `PasswordCallback` handlers
+
+In order to authenticate against a PKCS#11 token or to the internal
+certificate store, it is necessary to select a `PasswordCallback` handler.
+By default this is a console-based `PasswordCallback` handler. This prompts
+the user for the password via the Console. However, this is not appropriate
+in all scenarios.
+
+Developers are expected to extend and implement this as desired by their
+application.
+
+For example, to set a static `PasswordCallback` handler:
+
+```java
+CryptoManager cm = CryptoManager.getInstance();
+cm.setPasswordCallback(new Password("password".toCharArray()));
+```
+
+## Loading JSS via `java.security`
+
+You can directly add the `JSSProvider` by adding it to the `java.security`
+file:
+
+```properties
+security.provider.<n> = org.mozilla.jss.JSSProvider /path/to/jss.cfg
+```
+
+There are two ways use this `java.security` file: by directly installing
+it to the system or by using `-Djava.security.properties=/path/to/file`.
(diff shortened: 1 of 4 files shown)


cipherboy:
Discussion from https://github.com/dogtagpki/jss/pull/454 should be added here.

docs/usage/jssprovider.md
```diff
@@ -0,0 +1,205 @@
...
+import org.mozilla.jss.CryptoManager;
+
+CryptoManager.initialize("/path/to/nss-db");
+```
+
+At this point, JSS will be initialized and can be used. However, a password
+might still be required, so see the section below for providing password
+callback handlers.
+
+However, certain default values might not work in all situations. For
```
edewata, line 31: How about removing the "however" here since the previous paragraph also contains a "however"?

docs/usage/jssprovider.md
```diff
@@ -0,0 +1,205 @@
...
+
+`JSSProvider` behaves like the `SunPKCS11-NSS` provider, requiring a
+configuration file to initialize JSS (and the `CryptoManager` object).
+
+`jss.cfg` takes the same `InitializationValues` parameters, except in a
+properties file format.
+
+### JSS Config
+
+| property                           | `InitializationValues` member name  |
```
edewata, line 145: Some properties are mapped to `CryptoManager` method, so "InitializationValues member name" is a bit inaccurate. How about "Mapped to"?

docs/usage/jssprovider.md
```diff
@@ -0,0 +1,205 @@
...
+|------------------------------------|-------------------------------------|
+| `jss.fips`                         | `ivs.fipsMode`                      |
```
edewata, line 147: If we remove InitializationValues from the colum header, this should be `InitializationValues.fipsMode`. Same thing for other ivs fields below.

docs/usage/jssprovider.md
```diff
@@ -0,0 +1,205 @@
...
+Note that two equals signs may be used, in which case the system configuration
+is ignored and fully overridden by this file.
+
+`JSSProvider` behaves like the `SunPKCS11-NSS` provider, requiring a
+configuration file to initialize JSS (and the `CryptoManager` object).
+
+`jss.cfg` takes the same `InitializationValues` parameters, except in a
+properties file format.
+
+### JSS Config
```
edewata, line 143: I don't think we actually need the `jss.` or `nss.` prefixes, but it's up to you.

docs/usage/jssprovider.md
```diff
@@ -0,0 +1,205 @@
...
+`installJSSProviderFirst` are ignored, as they can be controlled by directly
+manipulating the `java.security` file.
+
+## Upgrading Old Code
+
+There are two paths to upgrade an older code base to a newer JSS version,
+using the JSSProvider interface:
+
+ 1. Continue using `CryptoManager.initialize(...)` as before. This gives the
+    local application the most control over the NSS DB path.
```
edewata, line 180: Should we clarify that there is no changes required?

docs/usage/jssprovider.md
```diff
@@ -0,0 +1,205 @@
...
+    would be sufficient to check the value of `CryptoManager.getInstance(...)`
+    before configuration:
+
+    ```java
+    try {
+        cm = CryptoManager.getInstance();
+    } catch (NotInitializedException nie) {
+        CryptoManager.initialize(...);
+        cm = CryptoManager.getInstance();
+    }
```
edewata, line 196: Not sure it's necessary to tell people to change their code this way.
Theoretically people can continue to use the old API which will work
with the old and the new JSS:
```
CryptoManager.initialize(...);
CryptoManager cm = CryptoManager.getInstance();
```
Once they no longer need to support the old JSS they can simply drop
`CryptoManager.initialize()` and switch to `java.security`.

docs/usage/jssprovider.md
```diff
@@ -0,0 +1,205 @@
...
+    ```
+
+    This gives the user control over NSS DB path via modifying either of
+    those two configuration files (or by providing a local override).
+
+Note that, between `java.security` and `CryptoManager.initialize()`, the
+latter takes precedence unless `CryptoManager` has already been initialize.
```
edewata, line 203: initialized

docs/usage/jssprovider.md
cipherboy: There's other subtrees (`ocsp` being the deeper one); I think it makes sense then to have a separate `jss` and `nss` subtree. For reference, the `nss.cfg` used by OpenJDK is flat:

```
name = NSS
nssLibraryDirectory = /usr/lib64
nssDbMode = noDb
attributes = compatibility
handleStartupErrors = ignoreMultipleInitialisation
```

But they also use camelCase and prefix NSS-specific attributes with `nss`.

docs/usage/jssprovider.md
```diff
@@ -0,0 +1,205 @@
...
+
+ - Taking only a NSS DB directory. This was utilized above.
+ - Taking a NSS DB directory, a prefix for the certificate database, a
+   prefix for the key database, and the name of the secmod configuration.
+
+The latter is a more advanced use case and few individuals likely need to use
+it. For more information, see the corresponding NSS documentation on these
+values.
+
+The following initialization values are supported:
```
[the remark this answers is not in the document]
cipherboy: Sure, I think that'd be better.

docs/usage/jssprovider.md
cipherboy: This is the section under `java.security`. Presumably, if they're using this code snippet, they'd prefer a `java.security` JSS instance over a `CryptoManager` instance. This describe a fallback when either JSS doesn't support `java.security` or JSS wasn't initialized that way. 

I've added a comment saying they could throw this exception instead of catching it.

I don't think its reasonable to assume everyone is going to switch to `java.security` right away, and those that will, won't necessarily support it consistently.

cipherboy:
@edewata I believe this is updated. Thanks for your feedback!

Thoughts?

docs/usage/jssprovider.md
```diff
@@ -0,0 +1,166 @@
+# `Mozilla-JSS` Provider
+
+The `Mozilla-JSS` JCA-compatible Provider exposes most of the functionality
+of JSS to external packages. This interface is the recommend interface most
+developers should build against. However, once the dependencies are satisfied
+and JSS's native component are available to the JVM, we still have to load
```
cipherboy, line 6: are->is

docs/usage/jssprovider.md
```diff
@@ -0,0 +1,166 @@
...
+and initialize JSS.
+
+There are two routes to do this:
+
+ 1. Via the `CryptoManager` interface, _and_
+ 2. Via `java.security`, directly loading the `JSSProvider`.
```
cipherboy, line 12: `JSSProvider` -> JSS Provider.

cipherboy:
This can always be updated in future PRs. Merging!

PR #444, opened 2020-03-25
Prevent post-handshake configuration changes

cipherboy:
Certain parameters cannot be modified after the handshake begins. These
are which mode we're using (client/server), cipher suites, and protocol
versions. Throw an exception when these are changed after the handshake
has begun.
    
----

Discussed in a [comment](https://github.com/dogtagpki/jss/pull/150#discussion_r394689897) on #150. Depends on #434, hence separate PR.

### DIFF
--- a/org/mozilla/jss/ssl/javax/JSSEngine.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngine.java
@@ -398,7 +398,13 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
      * Sets the list of enabled cipher suites from a a list of SSLCipher enum
      * instances.
      */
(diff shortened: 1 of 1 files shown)


org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -603,6 +616,12 @@ public void setKeyMaterials(PK11Cert our_cert, PK11PrivKey our_key) throws Illeg
             throw new IllegalArgumentException("JSSEngine.setKeyMaterials(): Either both cert and key must be null or both must be not-null");
         }
 
+        if (ssl_fd != null) {
```
cipherboy, line 619: I don't think this should be quite correct. We might rehandshake and need client authentication.

PR #445, opened 2020-03-25
JSS RSA-PSS support first cut.

jmagne:
Provide support for the various SHAxxxwithRSAPSS algorithms.

Supprt for 256, 384, and 512 variants included.
Included test case for SHA256withRSA/PSS.

This fix also requires a corresponding fix to the pki server in
order to exercise this functionality in the context of a pki
ca server.

### DIFF
--- a/org/mozilla/jss/JSSProvider.java
+++ b/org/mozilla/jss/JSSProvider.java
@@ -59,6 +59,28 @@ public final class JSSProvider extends java.security.Provider {
         put("Alg.Alias.Signature.SHA256/RSA", "SHA-256/RSA");
         put("Alg.Alias.Signature.SHA256withRSA", "SHA-256/RSA");
 
+        put("Signature.RSASSA-PSS",
+            "org.mozilla.jss.provider.java.security.JSSSignatureSpi$RSAPSSSignature");
+
+        put("Alg.Alias.Signature.1.2.840.113549.1.1.10",     "RSASSA-PSS");
+        put("Alg.Alias.Signature.OID.1.2.840.113549.1.1.10", "RSASSA-PSS");
+
+        put("Signature.SHA-256/RSA/PSS",
+            "org.mozilla.jss.provider.java.security.JSSSignatureSpi$SHA256RSAPSS");
+
+        put("Alg.Alias.Signature.SHA256withRSA/PSS","SHA-256/RSA/PSS");
+
+        put("Signature.SHA-384/RSA/PSS",
+            "org.mozilla.jss.provider.java.security.JSSSignatureSpi$SHA384RSAPSS");
+
+        put("Alg.Alias.Signature.SHA384withRSA/PSS","SHA-384/RSA/PSS");
+
+        put("Signature.SHA-512/RSA/PSS",
+            "org.mozilla.jss.provider.java.security.JSSSignatureSpi$SHA512RSAPSS");
+
+        put("Alg.Alias.Signature.SHA512withRSA/PSS","SHA-512/RSA/PSS");
+
+
         put("Signature.SHA-384/RSA",
             "org.mozilla.jss.provider.java.security.JSSSignatureSpi$SHA384RSA");
         put("Alg.Alias.Signature.SHA384/RSA", "SHA-384/RSA");
@@ -155,6 +177,9 @@ public final class JSSProvider extends java.security.Provider {
             "org.mozilla.jss.provider.java.security.IvAlgorithmParameters");
         put("AlgorithmParameters.RC2AlgorithmParameters",
             "org.mozilla.jss.provider.java.security.RC2AlgorithmParameters");
+
+        put("AlgorithmParameters.RSAPSSAlgorithmParameters",
+            "org.mozilla.jss.provider.java.security.RSAPSSAlgorithmParameters");
 
         /////////////////////////////////////////////////////////////
         // Cipher
--- a/org/mozilla/jss/netscape/security/x509/AlgorithmId.java
+++ b/org/mozilla/jss/netscape/security/x509/AlgorithmId.java
@@ -22,6 +22,10 @@ import java.io.OutputStream;
 import java.io.Serializable;
 import java.security.AlgorithmParameters;
 import java.security.NoSuchAlgorithmException;
+import java.security.NoSuchProviderException;
+import java.security.spec.InvalidParameterSpecException;
+import java.security.spec.MGF1ParameterSpec;
+import java.security.spec.PSSParameterSpec;
 
 import org.mozilla.jss.netscape.security.util.DerEncoder;
 import org.mozilla.jss.netscape.security.util.DerInputStream;
@@ -63,7 +67,9 @@ public class AlgorithmId implements Serializable, DerEncoder {
     private ObjectIdentifier algid = null;
 
     // The (parsed) parameters
-    private AlgorithmParameters algParams;
+    private AlgorithmParameters algParams = null;
+    // Use this for the various flavors of the RSA PSS alg.
+    private String cachedAlgName = null;
 
     /**
      * Parameters for this algorithm. These are stored in unparsed
@@ -97,11 +103,13 @@ public class AlgorithmId implements Serializable, DerEncoder {
     public static AlgorithmId get(String algname)
             throws NoSuchAlgorithmException {
         ObjectIdentifier oid = algOID(algname);
-
         if (oid == null)
             throw new NoSuchAlgorithmException("unrecognized algorithm name: " + algname);
-
-        return new AlgorithmId(oid);
+        try {
+            return new AlgorithmId(oid,algname);
+        } catch (Exception e) {
+            throw new NoSuchAlgorithmException(e);
+        }
     }
 
     /**
@@ -150,7 +158,11 @@ public class AlgorithmId implements Serializable, DerEncoder {
                 !algid.equals(sha512WithEC_oid)) {
             alg = new AlgorithmId(algid, params);
         } else {
-            alg = new AlgorithmId(algid);
+            try {
+                alg = new AlgorithmId(algid);
+            } catch (Exception e) {
+                throw new IOException(e);
+            }
         }
         if (params != null)
             alg.decodeParams();
@@ -177,12 +189,83 @@ public class AlgorithmId implements Serializable, DerEncoder {
     }
 
     /**
-     * Constructs a parameterless algorithm ID.
+     * Constructs an algorithm ID with a fully encoded params object
+     * @ param oid the identifier for the algorithm
+     * @ param params the fully encoded AlgorithmIdentifier Object
+     * @throws NoSuchAlgorithmException
+     * @throws IOException
+     */
+    public AlgorithmId(ObjectIdentifier oid, AlgorithmParameters params)
+        throws IOException, NoSuchAlgorithmException {
+        algid = oid;
+        algParams = params;
+
+        if(algParams == null) {
+            throw new NoSuchAlgorithmException("AlgorithmId: unsupported Alg Params.");
+        }
+        this.params = new DerValue(algParams.getEncoded());
+    }
+
+    /**
+     * Constructor that takes the oid and name, so the name can be cachedf or laster use.
+     * @throws NoSuchAlgorithmException
+     * @throws IOException
+     *
+     */
+    public AlgorithmId(ObjectIdentifier oid, String algName) 
+        throws  IOException, NoSuchAlgorithmException {
+        algid = oid;
+        cachedAlgName = algName;
+        /* Create the params if our algorithm is RSA PSS related */
+        if (algName != null && algName.contains("PSS")) {
+            this.algParams = getPSSParams(algName);
+            this.params = new DerValue(this.algParams.getEncoded());
+        }
+    }
+
+    /**
+     * Constructs an  algorithm ID with possible RSAPSS params.
      *
      * @param oid the identifier for the algorithm
      */
     public AlgorithmId(ObjectIdentifier oid) {
         algid = oid;
+        String algName = algName();
+
+        /* Create the params if our algorithm is RSA PSS related */
+        if (algName != null && algName.contains("PSS")) {
+            try {
+                this.algParams = getPSSParams(algName);
+
+                if (this.algParams != null) {
+                    try {
+                        this.params = new DerValue(this.algParams.getEncoded());
+                    } catch (IOException e) {
+                        throw new IOException(e);
+                    }
+                }
+            } catch (Exception e) {
+                //Preserve original signature...
+                System.out.println("Unable to create pssPrams in Algorithmid(ObjectIdentifier oid");
+                this.algParams = null;
+            }
+        }
+    }
+
+    private AlgorithmParameters getPSSParams(String algName)
+            throws NoSuchAlgorithmException, IOException {
+        cachedAlgName = algName;
+
+        AlgorithmParameters ret = null;
+        /* Create the params if our algorithm is RSA PSS related */
+        try {
+            ret = createPSSAlgorithmParameters(algName);
+        } catch (Exception e) {
+            throw new NoSuchAlgorithmException(e);
+        }
+
+        return ret;
+
     }
 
     private AlgorithmId(ObjectIdentifier oid, DerValue params)
@@ -204,8 +287,17 @@ public class AlgorithmId implements Serializable, DerEncoder {
 
     protected void decodeParams() throws IOException {
         try {
-            this.algParams = AlgorithmParameters.getInstance
-                    (this.algid.toString());
+
+            if (algid.equals(AlgorithmId.rsaPSS_oid)) {
+                try {
+                    this.algParams = createPSSAlgorithmParametersFromData(this.params.toByteArray());
+                    return;
+                } catch (Exception e) {
+                    throw new IOException(e);
+                }
+            } else {
+                this.algParams = AlgorithmParameters.getInstance(this.algid.toString());
+            }
         } catch (NoSuchAlgorithmException e) {
             /*
              * This algorithm parameter type is not supported, so we cannot
@@ -229,7 +321,6 @@ public class AlgorithmId implements Serializable, DerEncoder {
     /**
      * DER encode this object onto an output stream.
      * Implements the <code>DerEncoder</code> interface.
-     *
      * @param out
      *            the output stream on which to write the DER encoding.
      *
@@ -255,6 +346,47 @@ public class AlgorithmId implements Serializable, DerEncoder {
             out.write(tmp.toByteArray());
         }
     }
+
+    /**
+     * DER encode this object onto an output stream.
+     * Implements the <code>DerEncoder</code> interface.
+     *
+     * @param out
+     *            the output stream on which to write the DER encoding params,using context value.
+     *
+     * @exception IOException on encoding error.
+     */
+    public void derEncodeWithContext(OutputStream out,int contextVal) throws IOException {
+        try (DerOutputStream tmp = new DerOutputStream()) {
+            DerOutputStream bytes = new DerOutputStream();
+            bytes.putOID(algid);
+
+            byte val = (byte) contextVal;
+            // omit parameter field for ECDSA
+            if (!algid.equals(sha224WithEC_oid) &&
+                    !algid.equals(sha256WithEC_oid) &&
+                    !algid.equals(sha384WithEC_oid) &&
+                    !algid.equals(sha512WithEC_oid)) {
+                if (params == null) {
+                    bytes.putNull();
+                } else
+                    bytes.putDerValue(params);
+            }
+
+            DerOutputStream seq = new DerOutputStream();
+
+            seq.write(DerValue.tag_Sequence, bytes);
+
+            tmp.write(DerValue.createTag(DerValue.TAG_CONTEXT,
+                                             true, val), seq);
+
+
+
+            out.write(tmp.toByteArray());
+        }
+    }
+
+
 
     // XXXX cleaning required
     /**
@@ -369,6 +501,8 @@ public class AlgorithmId implements Serializable, DerEncoder {
         if (name.equals("SHAwithDSA") || name.equals("SHA1withDSA")
                 || name.equals("SHA/DSA") || name.equals("SHA1/DSA"))
             return AlgorithmId.sha1WithDSA_oid;
+        if (name.equals("SHA256withRSA/PSS") || name.equals("SHA384withRSA/PSS") || name.equals("SHA512withRSA/PSS"))
+            return AlgorithmId.rsaPSS_oid;
 
         return null;
     }
@@ -385,7 +519,6 @@ public class AlgorithmId implements Serializable, DerEncoder {
      */
     private String algName() {
         // Common message digest algorithms
-
         if (algid.equals(AlgorithmId.MD5_oid))
             return "MD5"; // RFC 1423
         if (algid.equals(AlgorithmId.MD2_oid))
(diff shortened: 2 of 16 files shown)


org/mozilla/jss/JSSProvider.java
```diff
@@ -59,6 +59,21 @@ public JSSProvider() {
         put("Alg.Alias.Signature.SHA256/RSA", "SHA-256/RSA");
         put("Alg.Alias.Signature.SHA256withRSA", "SHA-256/RSA");
 
+        put("Signature.SHA-256/RSA/PSS",
```
cipherboy: I'm fine with these names for clarity and specifying which hash algorithm to use, but please add the supported name too, and allow for detection of the signature parameters:

(from `SunRsaSignEntries.java` in JDK11)
```java
        map.put("Signature.RSASSA-PSS",
                "sun.security.rsa.RSAPSSSignature");
```

And also the other aliases they expect:

```
src/java.base/share/classes/sun/security/rsa/SunRsaSignEntries.java:71:        map.put("Signature.RSASSA-PSS",
src/java.base/share/classes/sun/security/rsa/SunRsaSignEntries.java:129:        map.put("Alg.Alias.Signature.1.2.840.113549.1.1.10",     "RSASSA-PSS");
src/java.base/share/classes/sun/security/rsa/SunRsaSignEntries.java:130:        map.put("Alg.Alias.Signature.OID.1.2.840.113549.1.1.10", "RSASSA-PSS");

```

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
```diff
@@ -22,7 +22,10 @@
 import java.io.Serializable;
 import java.security.AlgorithmParameters;
 import java.security.NoSuchAlgorithmException;
-
+import java.security.NoSuchProviderException;
+import java.security.spec.InvalidParameterSpecException;
+import java.security.spec.MGF1ParameterSpec;
+import java.security.spec.PSSParameterSpec;
```
cipherboy: I'd keep a newline I think.

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
```diff
@@ -204,8 +291,19 @@ public AlgorithmId() {
 
     protected void decodeParams() throws IOException {
         try {
-            this.algParams = AlgorithmParameters.getInstance
-                    (this.algid.toString());
+
+            if (algid.equals(AlgorithmId.rsaPSS_oid)) {
+                try {
+                    //dumpBytes(this.params.toByteArray());
```
cipherboy: I'd remove this for review ;-)

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
```diff
@@ -552,7 +714,64 @@ public final boolean equals(ObjectIdentifier id) {
...
+     * this can be extended to support SHA384 and 512 PSS digests.
+    */
+    public static AlgorithmParameters createPSSAlgorithmParameters(String algName) throws Exception {
+        if (algName == null) {
+            throw new Exception("Invalid Algorithm name input.");
+        }
+        AlgorithmParameters pssParams = null;
+        PSSParameterSpec pssSpec = null;
+        // Make sure we are in the RSA PSS family
+        if (algName.contains("PSS")) {
```
cipherboy: This'd be simpler if we wrote:

`if (!algName.contains("PSS")) { throw Exception(...) }` and then we wouldn't need the body indented.

cipherboy:
@jmagne Can you rebase this on top of the latest `v4.4.x`? We needed to update CI on `v4.4.x`. :-)

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -31,12 +32,18 @@ getSomeKey(JNIEnv *env, jobject sig, void **key, short type);
...
+getDigestAlgorithm(JNIEnv *env, jobject sig);
+
+
 static void
 setSigContext(JNIEnv *env, jobject sig, jobject context);
 
 static PRStatus
 getSigContext(JNIEnv *env, jobject sig, void**pContext, SigContextType* pType);
 
+static SECAlgorithmID *getRSAPSSParamsAndSigningAlg(PRArenaPool *arena,JNIEnv *env, jobject this,SECKEYPrivateKey *privk);
```
cipherboy: nit: I'd prefer `env, this, arena, privk` rather than `arena, env, this, privk` -- `env, this` pattern is typically used from the `native` stubs JNI produces and is always first. I think grouping Java and then NSS arguments makes the most sense to me. 

And one other thing to consider: what do the failure cases in `getRSAPSSParamsAndSigningAlg` look like? Would

```c
static SECStatus getRSAPSSParamsAndSigningAlg(JNIEnv *env, jobject this, PRArenaPool *arena, SECAlgorithmId *alg, SECKEYPrivateKey privk);
```

be a better signature? Then you can do (below):

```c
SECAlgorithmId signAlg;
if (getRSAPSSParamsAndSigningAlg(env, this, arena, &signAlg, privk) != PR_SUCCESS) {
    goto finish;
}

ctx = SGN_NewContextWithAlgorithmID(&signAlg, privk);
```

and the intent is more clear IMO.

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -48,39 +55,79 @@ Java_org_mozilla_jss_pkcs11_PK11Signature_initSigContext
...
+        SGN_CONTEXT,arena);
+
     if(contextProxy == NULL) {
         PR_ASSERT( (*env)->ExceptionOccurred(env) != NULL);
         goto finish;
     }
+
     setSigContext(env, this, contextProxy);
 
 finish:
```
cipherboy: FWIW, I think you leak `signAlg` here. If you accept my [other suggestion](https://github.com/dogtagpki/jss/pull/445/files#r398823299), you wouldn't.

cipherboy: Ah I see. You're doing an arena-based allocation for the `signAlg` too. That definitely warrants a comment.

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -48,39 +55,79 @@ Java_org_mozilla_jss_pkcs11_PK11Signature_initSigContext
...
+    int keyTypeOKForPSS = 0;
 
     /* Extract the private key from the PK11Signature */
     if( getPrivateKey(env, this, &privk) != PR_SUCCESS) {
         PR_ASSERT( (*env)->ExceptionOccurred(env) != NULL);
         goto finish;
     }
+    
+    // See if the key is a candidate for PSS signature 
+    if (privk->keyType == rsaKey || privk->keyType == rsaPssKey) {
```
cipherboy: Why not pull this into a macro instead of a variable? Seems weird, since you only use it once.

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -48,39 +55,79 @@ Java_org_mozilla_jss_pkcs11_PK11Signature_initSigContext
...
 
-    /* Start the signing operation */
-    ctxt = SGN_NewContext(getAlgorithm(env, this), privk);
-	if(ctxt == NULL) {
-		JSS_throwMsg(env, TOKEN_EXCEPTION, "Unable to create signing context");
-		goto finish;
-	}
+    signingAlg = getAlgorithm(env,this);
+
+    if(signingAlg == SEC_OID_PKCS1_RSA_PSS_SIGNATURE && keyTypeOKForPSS) {
```
cipherboy: What I'm missing is why are we continuing if the key isn't OK for PSS?

In particular, when `signingAlg == SEC_OID_PKCS1_RSA_PSS_SIGNATURE` and `!keyTypeOKForPSS`, shouldn't we fail?

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: I will look at it, I did do a valgrind run at one point and it looked ok, thx...

cipherboy:
Nah I convinced myself it was good :-)

org/mozilla/jss/JSSProvider.java
jmagne: Done.

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
jmagne: Done, I think :)

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
jmagne: Done.

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
jmagne: Done.

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: I took your advice mostly.. Done. Thx.

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: Added comment somewhere :)

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: Got rid of the variable. Upon further review this is a redundant check, we only care if the alg is correct.

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: The else branch is just for everything else not PSS. Note that the context is created differently in those two cases.

cipherboy: Right, I'm happier with the new check.

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -90,20 +133,67 @@ Java_org_mozilla_jss_pkcs11_PK11Signature_initVfyContext
 {
 	VFYContext *ctxt=NULL;
 	jobject contextProxy=NULL;
-	SECKEYPublicKey *pubk;
+        SECKEYPublicKey *pubk;
```
cipherboy: We should probably keep the same indentation as the rest of the file.

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -355,6 +454,81 @@ getAlgorithm(JNIEnv *env, jobject sig)
...
+    }
+
+    retval = JSS_getOidTagFromAlg(env, alg);
+
+finish:
+    return retval;
+}
+
+static
+
```
cipherboy: Spare newline.

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -355,6 +454,81 @@ getAlgorithm(JNIEnv *env, jobject sig)
...
+SECStatus getRSAPSSParamsAndSigningAlg(JNIEnv *env, jobject this, PRArenaPool *arena, SECAlgorithmID **alg, SECKEYPrivateKey *privk)
+{
+    SECItem *sigAlgParams;
+    SECAlgorithmID *signAlg=NULL;
+    SECOidTag digestAlg = SEC_OID_UNKNOWN;
+    SECStatus rv = SECSuccess ;
+
+    if(alg == NULL) {
```
cipherboy: Space please? :-) 

```c
if (alg == NULL) {
```

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -355,6 +454,81 @@ getAlgorithm(JNIEnv *env, jobject sig)
...
+}
+
+static
+
+SECStatus getRSAPSSParamsAndSigningAlg(JNIEnv *env, jobject this, PRArenaPool *arena, SECAlgorithmID **alg, SECKEYPrivateKey *privk)
+{
+    SECItem *sigAlgParams;
+    SECAlgorithmID *signAlg=NULL;
+    SECOidTag digestAlg = SEC_OID_UNKNOWN;
+    SECStatus rv = SECSuccess ;
```
cipherboy: Oh here's where that space went. ;-) I'd remove this one. 

```c
SECStatus rv = SECSuccess;
```

Also, do we want this to be `SECFailure`?

I'm confused:

```c
if (alg == NULL) {
    return SECSuccess;
}
```

Why would we want a success exit code here? 

```c
    if (signAlg == NULL) {
        return SECSuccess;
    }
```

Also seems like the wrong return if `PORT_AreanZAlloc` fails.

```c

    if (!sigAlgParams) {
        rv = SECFailure;
        goto finish;
    }
```

That seems right. :-) But we can just initialize `rv` to `SECSuccess` and remove the set here. :)

cipherboy: Actually, going a step further, I don't think the `finish` label is really required. The `arena` handles freeing the memory. So we can replace `goto finish` with `return rv` directly.

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -644,11 +823,17 @@ Java_org_mozilla_jss_pkcs11_SigContextProxy_releaseNativeResources
 
     /* Free the context and the proxy */
     if(proxy->type == SGN_CONTEXT) {
-        SGN_DestroyContext( (SGNContext*)proxy->ctxt, PR_TRUE /*freeit*/);
-    } else {
+        SGN_DestroyContext( (SGNContext*)proxy->ctxt, PR_TRUE /*freeit*/); } else {
```
cipherboy: Why did this get collapsed?

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -644,11 +823,17 @@ Java_org_mozilla_jss_pkcs11_SigContextProxy_releaseNativeResources
...
         PR_ASSERT(proxy->type == VFY_CONTEXT);
         VFY_DestroyContext( (VFYContext*)proxy->ctxt, PR_TRUE /*freeit*/);
     }
+
+    if(proxy->arena != NULL) {
+        PR_fprintf(PR_STDERR,"About to remove arena!\n");
```
cipherboy: Debug statement should probably be removed.

cipherboy: And space on the line above too :-)

org/mozilla/jss/pkcs11/PK11Signature.java
```diff
@@ -317,12 +334,47 @@ native protected boolean engineVerifyNative(byte[] sigBytes)
...
+         String hashAlgName = params.getDigestAlgorithm();
+         Algorithm hashAlg = null;
+        if(hashAlgName.equals("SHA-256")) {
+            hashAlg =  DigestAlgorithm.SHA256;
+        }  else if(hashAlgName.equals("SHA-512")) {
+            hashAlg =  DigestAlgorithm.SHA512;
+        }  else if(hashAlgName.equals("SHA-384")) {
+            hashAlg =  DigestAlgorithm.SHA384;
+        } else {
+            hashAlg =  DigestAlgorithm.SHA256;
```
cipherboy: Not sure I understand this one. Shouldn't we fail here if no algorithm was specified?

org/mozilla/jss/pkcs11/PK11Signature.java
```diff
@@ -317,12 +334,47 @@ native protected boolean engineVerifyNative(byte[] sigBytes)
     public void engineSetParameter(AlgorithmParameterSpec params)
         throws InvalidAlgorithmParameterException, TokenException
     {
-        Assert.notYetImplemented("PK11Signature.engineSetParameter");
+        //For now we only care about RSAPSS
+        if( isRSAPSSAlgorithm((SignatureAlgorithm) algorithm) ) {
+            digestAlgorithm = getRSAPSSDigestAlgFromSpec((PSSParameterSpec) params);
+        }
```
cipherboy: Do we want to leave an else block and say it isn't implemented / supported to pass parameters to other signature schemes?

org/mozilla/jss/pkcs11/PK11Signature.java
```diff
@@ -317,12 +334,47 @@ native protected boolean engineVerifyNative(byte[] sigBytes)
...
+
+            return true;
+        }
+        return false;
     }
 
 	protected PK11Token token;
 	protected TokenProxy tokenProxy;
 	protected Algorithm algorithm;
+        protected Algorithm digestAlgorithm;
```
cipherboy: Could we keep the indentation the same? Thanks!

org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
```diff
@@ -35,12 +36,16 @@ public void engineInitSign(java.security.PrivateKey privateKey)
     {
         try {
             sig = getSigContext(privateKey);
+            if(paramSpec != null) sig.setParameter(paramSpec);
```
cipherboy: At least a separate line? :-)

org/mozilla/jss/provider/java/security/RSAPSSAlgorithmParameters.java
```diff
@@ -0,0 +1,283 @@
...
+            return "SHA-256"; 
+        else if("SHA384".equals(algName))
+            return "SHA-384";
+        else if("SHA512".equals(algName))
+            return "SHA-512";
+        else //default
+            return "SHA-1";
+
+    }
+    private static void dumpBytes(byte[] data) {
```
cipherboy: Could we remove this too? And the comments above?

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: Apologies, was still working on this latish last night :)

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
```diff
@@ -177,12 +189,88 @@ public static AlgorithmId parse(byte[] val)
     }
 
     /**
-     * Constructs a parameterless algorithm ID.
+     * Constructs an algorithm ID with a fully encoded params object
+     *
+     * @ param oid the identifier for the algorithm
```
cipherboy: I think if you remove the space here, it'll make javadocs happy.

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: I don't understand this one. If you look at the file, it's all aligned..

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
```diff
@@ -177,12 +189,88 @@ public static AlgorithmId parse(byte[] val)
...
+    }
+
+    /**
+     * Constructor that takes the oid and name, so the name can be cachedf or laster use.
+     * @throws NoSuchAlgorithmException
+     * @throws IOException
+     *
+     */
+
+    public AlgorithmId(ObjectIdentifier oid, String algName) throws  IOException {
```
cipherboy: Remove space above, and should this also throw NSAE?

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
```diff
@@ -177,12 +189,88 @@ public static AlgorithmId parse(byte[] val)
...
+    }
+
+    /**
+     * Constructs an  algorithm ID with possible RSAPSS params.
      *
      * @param oid the identifier for the algorithm
+     * @throws NoSuchAlgorithmException
+     * @throws IOException
      */
     public AlgorithmId(ObjectIdentifier oid) {
```
cipherboy: Same as above, seems like this should throw the new exceptions if at all possible.

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
```diff
@@ -177,12 +189,88 @@ public static AlgorithmId parse(byte[] val)
     }
 
     /**
-     * Constructs a parameterless algorithm ID.
+     * Constructs an algorithm ID with a fully encoded params object
+     *
+     * @ param oid the identifier for the algorithm
+     * @ param params the fully encoded AlgorithmIdentifier Object
+     */
+
```
cipherboy: Same as below; NSAE and IO Exception should be in the constructor and remove space.

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
```diff
@@ -256,6 +353,47 @@ public void derEncode(OutputStream out) throws IOException {
         }
     }
 
+    /**
+     * DER encode this object onto an output stream.
+     * Implements the <code>DerEncoder</code> interface.
+     *
+     * @param out
```
cipherboy: Spare newline. :-)

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
```diff
@@ -566,7 +784,8 @@ public final boolean equals(ObjectIdentifier id) {
     private static final int SHA256_data[] = { 2, 16, 840, 1, 101, 3, 4, 2, 1 };
     private static final int SHA384_data[] = { 2, 16, 840, 1, 101, 3, 4, 2, 2 };
     private static final int SHA512_data[] = { 2, 16, 840, 1, 101, 3, 4, 2, 3 };
-
+    private static final int SHA224_data[] = { 2, 16, 840, 1, 101, 3, 4, 2, 4 };
```
cipherboy: Just askin', but why was SHA224 added here? We don't really support SHA224 anywhere else in JSS.

org/mozilla/jss/pkcs11/PK11Signature.c
cipherboy: The issue is tabs versus spaces. The rest of this function uses tabs I think, but you've inserted spaces.

Try `:set list` in vim to show tabs/spaces. Not sure about other editors. :)

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
jmagne: I was doing this way back when for completeness. I think I will just get rid of it.

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: Done. I think. If I missed any of these, it was an over sight... :)

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: Done.

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: Fixed this to the spirit of what you were suggesting...

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: I do no know how that happened, but I fixed it.

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: Done.

org/mozilla/jss/pkcs11/PK11Signature.java
jmagne: Done, throws exception.

org/mozilla/jss/pkcs11/PK11Signature.java
jmagne: Done, throw exception.

org/mozilla/jss/pkcs11/PK11Signature.java
jmagne: There was a mixup of this and existing code. I just fixed them all in that section.

org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
jmagne: Done.

org/mozilla/jss/provider/java/security/RSAPSSAlgorithmParameters.java
jmagne: Done.

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
jmagne: Done.

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
jmagne: Done.

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
jmagne: Done, I think :)

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
jmagne: Done.

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
jmagne: Got rid of it.

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: I missed this one. Soon I will have to move on to pki :)

org/mozilla/jss/netscape/security/x509/AlgorithmId.java
jmagne: OH , now I remember, I wanted to keep this the same signature, since it is heavily used, could be breaking code all over the place.

PR #446, opened 2020-03-25
Document known NSS incompatibilities

cipherboy:

### DIFF
--- a/docs/nss.md
+++ b/docs/nss.md
@@ -1,0 +1,22 @@
+# NSS
+
+## Incompatible NSS/JSS Pairings
+
+The minimum NSS version required by JSS is v3.44. Up to v3.44, NSS didn't
+correctly support SECKEY_ConvertToPublicKey, returning NULL in some cases
+with an EC private key. This was fixed in v3.44.
+
+### JSS v4.6.3
+
+JSS v4.6.3 doesn't support NSS versions between v3.47 and v3.50, inclusive.
+This is because [moz-bz#1570501][moz-bz-1570501] introduced a bug that wasn't
+caught and fixed until [moz-bz#1611209][moz-bz-1611209]. NSS versions v3.46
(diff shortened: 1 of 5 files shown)


docs/nss.md
```diff
@@ -0,0 +1,21 @@
+# NSS
+
+## Incompatible NSS/JSS Pairings
+
+The minimum NSS version required by JSS is v3.30.
+
+### JSS v4.6.3
+
+JSS v4.6.3 doesn't support NSS versions between v3.47 and v3.50, inclusive.
```
edewata: Should we specify the `Conflicts` with those NSS versions in `jss.spec`?

cipherboy: I guess we could've for that version. :-) But for this version, we've fixed the incompatibilities. And it only was incompatible if you tried building (tests would've built) or if you tried using CMAC or KBKDF. They're not popular -- we're the only ones that use them (and these versions of NSS are really new) -- so "meh"? :-) I could push an update to Fedora if you'd like though.

edewata: If it's only a build issue it's probably not necessary since we control Fedora builds (unless someone accidentally builds with the wrong NSS version). If it's a potential runtime issue (i.e. user installs the wrong JSS/NSS combinantion) we might want to add the `Conflicts`.

cipherboy:
FWIW, I just marked this as WIP as I _do_ want to try and find the Debian versions I recall having issues with, along with validating the minimum version still stands. It might now be at 3.31, but I need to check it and I'm having issues building old NSS versions :-)

cipherboy:
@edewata OK, you win. I bumped to NSS v3.44.

cipherboy:
FWIW the commit was merged back in [JSS v4.6.2](https://github.com/dogtagpki/jss/commit/cfd0ff7493af712ba42f853e5f212770f3f97b55#diff-39131a065bda79b8a2896391e485cb0e)

cipherboy:
Branching done, merging.

PR #449, opened 2020-03-26
Add GitHub Actions - v4.4.x Backport

cipherboy:
This is a backport of our move to GitHub Actions to the v4.4.x branch.

---

This branch supports older Fedora and CentOS versions. It doesn't rely on CMake.

I've left F29 in place (just like F26 is left in place) to keep testing a wider range of older Fedora versions.


cipherboy:
CI passed, I think that's all that can be expected. :-)

PR #454, opened 2020-03-30
CryptoManager - Remove initialization from test suite

cipherboy:
In an unrelated effort to add another test to JSS, the following bug was discovered:

`CryptoManager.getInstance()` throws a `NotInitializedException` when the `Mozilla-JSS` provider hasn't yet loaded. This means that the following basic test case will fail:

```
instance = CryptoManager.getInstance()
```

Because the provider isn't yet loaded from the `java.security` file. There are four commits here:

 1. Add the above check, and update `CryptoManager.initialize(...)` to be aware of `java.security`.  
 2. Migrate two tests to using this new form of initialize, as they need to be configured specially. (One creates new NSS DBs).
 3. Ensure these tests don't get passed a `java.security` file anyways.
 4. Remove `CryptoManager.initialize(...)` from all other existing tests, to prove the `JSSLoader` really does work. :-)

Some justification and discussion for these changes is given [here](https://gist.github.com/cipherboy/308229616364946df94b382cc93bb5a3).

### DIFF
--- a/cmake/JSSTests.cmake
+++ b/cmake/JSSTests.cmake
@@ -167,7 +167,7 @@ macro(jss_tests)
     )
     jss_test_java(
         NAME "List_CA_certs"
-        COMMAND "org.mozilla.jss.tests.ListCACerts" "${RESULTS_NSSDB_OUTPUT_DIR}"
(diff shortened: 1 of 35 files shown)


cipherboy:
`fedora_sandbox` is broken due to Fedora Rawhide being broken. This is unrelated to these changes (we don't control DNF packaging of perl scripts), so I'm going ahead and merging this.

Thanks @edewata!

PR #463, opened 2020-04-01
Add optional CI to check if PKI build is broken

cipherboy:
This adds an optional CI check that builds JSS and then builds PKI,
installing both sets of RPMs.

----

@SilleBille -- is it possible to execute a PKI job inside this? I'd like to at least spawn a few subsystems, but that requires an init system. Got any thoughts for me?

### DIFF
--- a/tools/Dockerfiles/pki_build
+++ b/tools/Dockerfiles/pki_build
@@ -1,0 +1,33 @@
+FROM fedora:latest
+
+# Install generic dependencies to build jss and pki
+RUN true \
+        && dnf update -y --refresh \
+        && dnf install -y dnf-plugins-core gcc make rpm-build \
+        && dnf copr -y enable ${JSS_4_6_REPO:-@pki/master} \
+        && dnf build-dep -y jss pki-core \
+        && mkdir -p /home/sandbox \
+        && git clone https://github.com/dogtagpki/pki /home/sandbox/pki \
+        && dnf clean -y all \
+        && true
+
+# Link in the current version of jss from the git repository
+WORKDIR /home/sandbox
+COPY . /home/sandbox/jss
+
+# Install dependencies from the spec file in case they've changed
+# since the last release on this platform.
+RUN true \
+        && dnf build-dep -y --spec /home/sandbox/jss/jss.spec \
+        && dnf build-dep -y --spec /home/sandbox/pki/pki.spec \
+        && true
+
+# Perform the actual RPM build
+WORKDIR /home/sandbox/jss
+CMD true \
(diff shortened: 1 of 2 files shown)


cipherboy:
@edewata Without executing PKI Travis (Reverse CI?), this won't help our particular issues here. But it should help know if we make a (build) breaking change in the future.

tools/Dockerfiles/pki_build
```diff
@@ -0,0 +1,33 @@
+FROM fedora:latest
+
+# Install generic dependencies to build jss and pki
+RUN true \
+        && dnf update -y --refresh \
+        && dnf install -y dnf-plugins-core gcc make rpm-build \
+        && dnf copr -y enable ${JSS_4_6_REPO:-@pki/master} \
+        && dnf build-dep -y jss pki \
```
edewata, line 8: I think it should be `pki-core` instead of `pki`.

tools/Dockerfiles/pki_build
```diff
@@ -0,0 +1,33 @@
+FROM fedora:latest
+
+# Install generic dependencies to build jss and pki
+RUN true \
+        && dnf update -y --refresh \
+        && dnf install -y dnf-plugins-core gcc make rpm-build \
+        && dnf copr -y enable ${JSS_4_6_REPO:-@pki/master} \
+        && dnf build-dep -y jss pki-core \
```
SilleBille, line 8: builddep is covered again in line 21 and 22, which will install any missing/updated deps. Do we want this line?

cipherboy: This pulls in 99% of the deps in the first layer. The layer is cleaned (see `dnf clean -y all` -- though I had to keep `tzinfo` and the other directories unlike the other layers because PKI build fails without it for some reason).

This results in a smaller first layer with all the deps, and then only what is new (in these two versions of JSS/PKI) is pulled in during a separate layer.

cipherboy:
@SilleBille -- I'm confused though, why would I want to trigger PKI's CI via COPR? I want a `pkispawn` to happen on every pull request against JSS, as an optional check. I don't want to execute PKI's entire CI pipeline, and I don't want to merge the PR first -- only to later find out it breaks PKI. You know?

SilleBille:
This was an idea I came up with when you mentioned "Reverse CI". We can't really trigger the chain of dependencies that JSS depends on. But, we can do a reverse CI on the major consumer of jss -- PKI.

This was a very rough idea and was not very well laid out. But, in general, when we move PKI to GH actions, we have the option to upload artifacts to GH repo itself. :-)

PR #467, opened 2020-04-02
Support RSA-PSS Signature scheme

cipherboy:
Provide support for the various `SHAxxxwithRSAPSS` algorithms, including
SHA-256, SHA-384, and SHA-512 variants.

Authored by Jack Magne; revised patch forwarded ported from [`v4.4.x`](https://github.com/dogtagpki/jss/pull/445) by
Alexander Scheel.


org/mozilla/jss/provider/java/security/RSAPSSAlgorithmParameters.java
```diff
@@ -0,0 +1,244 @@
...
+        // Sequence has 3 members, trailer field ignored
+        DerValue seq[] = in.getSequence(3);
+        if (seq.length != 3) {
+            throw new IOException("Invalid data! Expected a sequence of at least 3 members.");
+        }
+
+        if (seq[0].isContextSpecific((byte)0)) {
+            seq[0] = seq[0].data.getDerValue();
+        } else {
+             throw new IOException("Invalid encoded data!");
```
cipherboy, line 134: @jmagne While this is still fresh, do you have better descriptions here? I'm not sure I've fully grasped this DER decoding business here and its relevance.

This applies to lines 148 also.

org/mozilla/jss/provider/java/security/RSAPSSAlgorithmParameters.java
```diff
@@ -0,0 +1,244 @@
...
+
+    private void decode(DerInputStream in , byte[] encoded) throws IOException {
+        if (in == null) {
+            throw new IOException("Invalid input: got null DerInputStream");
+        }
+
+        // Sequence has 3 members, trailer field ignored
+        DerValue seq[] = in.getSequence(3);
+        if (seq.length != 3) {
+            throw new IOException("Invalid data! Expected a sequence of at least 3 members.");
```
cipherboy, line 128: **of exactly 3 members**.

org/mozilla/jss/provider/java/security/RSAPSSAlgorithmParameters.java
```diff
@@ -0,0 +1,244 @@
...
+
+        // Create the hash alg and mask gen func objects
+        if (hashAlgName.equals("SHA-256")) {
+           hashAlg = new AlgorithmId(AlgorithmId.SHA256_oid);
+        }  else if(hashAlgName.equals("SHA-512")) {
+           hashAlg = new AlgorithmId(AlgorithmId.SHA512_oid);
+        }  else if(hashAlgName.equals("SHA-384")) {
+           hashAlg = new AlgorithmId(AlgorithmId.SHA384_oid);
+        } else {
+           hashAlg = new AlgorithmId(AlgorithmId.SHA_oid);
```
cipherboy, line 228: @jmagne Why do we fallback to SHA1 here? Is there a relevant portion of the spec somewhere?

org/mozilla/jss/provider/java/security/RSAPSSAlgorithmParameters.java
```diff
@@ -0,0 +1,244 @@
...
+    private String getSpecAlgName(String algName) {
+        if ("SHA256".equals(algName)) {
+            return "SHA-256";
+        } else if("SHA384".equals(algName)) {
+            return "SHA-384";
+        } else if("SHA512".equals(algName)) {
+            return "SHA-512";
+        } else {
+            // Default to SHA-1
+            return "SHA-1";
```
cipherboy, line 241: @jmagne Same here?

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -45,87 +53,168 @@ JNIEXPORT void JNICALL
...
-	}
+    if (contextProxy == NULL && ctxt != NULL) {
+        /* we created a context but not the Java wrapper, so we need to
+         * delete the context here */
+        VFY_DestroyContext(ctxt, PR_TRUE /*freeit*/);
+    }
+
+    if (tempPubKey != NULL) {
+        SECKEY_DestroyPublicKey(tempPubKey);
+        tempPubKey = NULL;
```
cipherboy, line 211: remove

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -332,6 +421,73 @@ getAlgorithm(JNIEnv *env, jobject sig)
...
+    PR_ASSERT(env != NULL && sig != NULL);
+
+    sigClass = (*env)->GetObjectClass(env, sig);
+    PR_ASSERT(sigClass != NULL);
+
+    algField = (*env)->GetFieldID(env, sigClass, SIG_DIGEST_ALGORITHM_FIELD,
+                                  SIG_ALGORITHM_SIG);
+    if (algField == NULL) {
+        ASSERT_OUTOFMEM(env);
+        goto finish;
```
cipherboy, line 441: return, don't `goto`.

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -45,87 +53,168 @@ JNIEXPORT void JNICALL
...
     setSigContext(env, this, contextProxy);
 
 finish:
-    if(contextProxy==NULL && ctxt!=NULL) {
+    if (contextProxy == NULL && ctxt != NULL) {
         /* we created a context but not the Java wrapper, so we need to
          * delete the context here. */
         SGN_DestroyContext(ctxt, PR_TRUE /*freeit*/);
+        if (arena != NULL) {
+            PORT_FreeArena(arena, PR_TRUE /* zero */);
```
cipherboy, line 122: When `ctx == NULL` and we're in an error case, we leak `arena`, meaning `signAlg` gets leaked.

org/mozilla/jss/provider/java/security/RSAPSSAlgorithmParameters.java
jmagne: OK this context stuff has to do with the sequence of this thing, here is the blob again:

RSASSA-PSS-params ::= SEQUENCE {
   hashAlgorithm      [0] OAEP-PSSDigestAlgorithms  DEFAULT sha1,
   maskGenAlgorithm   [1] PKCS1MGFAlgorithms  DEFAULT mgf1SHA1,
   saltLength         [2] INTEGER  DEFAULT 20,
   trailerField       [3] INTEGER  DEFAULT 1


See, the 0,1,2, and 3 ?? This stuff goes into the encoding and is read out when decoding.. It took me many failures of constructing a legal blob before I looked into the context tags like here...

org/mozilla/jss/provider/java/security/RSAPSSAlgorithmParameters.java
jmagne: I just put that there cause other implementations of this thing I observed that being the default value. I can see throwing an exception could be appropriate here..

org/mozilla/jss/provider/java/security/RSAPSSAlgorithmParameters.java
jmagne: Same as above, but of course this should never happen. but ...

org/mozilla/jss/pkcs11/PK11Signature.c
jmagne: Yep, looks to be true.

org/mozilla/jss/provider/java/security/RSAPSSAlgorithmParameters.java
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    private void decode(DerInputStream in , byte[] encoded) throws IOException {
+        if (in == null) {
+            throw new IOException("Invalid input: got null DerInputStream");
+        }
+
+        // Sequence has 3 members, trailer field ignored
+        DerValue seq[] = in.getSequence(3);
+        if (seq.length != 3) {
```
cipherboy, line 127: @jmagne So I'm confused here. The comment above says "trailer field ignored". I'm not entirely clear on how `getSequence` works, but if there was a trailer field, wouldn't it return an array of 4 `DerValue` elements? 

In particular `readVector(...)` uses `startLen` as the initialize size of the vector, but it'll grow if there are four fields.

Shouldn't this read `if (seq.length < 3 || seq.length > 4)`?

org/mozilla/jss/provider/java/security/RSAPSSAlgorithmParameters.java
cipherboy: Pending resolution of above.

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -45,87 +53,160 @@ JNIEXPORT void JNICALL
...
-			"Unable to begin verification context");
-		goto finish;
-	}
+    signingAlg = getAlgorithm(env,this);
+    if (signingAlg == SEC_OID_PKCS1_RSA_PSS_SIGNATURE) {
+        /* Create place holder private key, just to create the PSS Params. */
+        unsigned key_bits = SECKEY_PublicKeyStrengthInBits(pubk);
+        privk = SECKEY_CreateRSAPrivateKey(key_bits, &tempPubKey, NULL);
+        if (privk == NULL) {
+            PR_ASSERT((*env)->ExceptionOccurred(env) != NULL);
```
cipherboy, line 151: This will never assert.

org/mozilla/jss/pkcs11/PK11Signature.c
```diff
@@ -304,32 +385,95 @@ getAlgorithm(JNIEnv *env, jobject sig)
...
+    SECOidTag digestAlg = SEC_OID_UNKNOWN;
+    SECStatus rv = SECFailure;
+
+    if (alg == NULL) {
+        return rv;
+    }
+
+    signAlg = (SECAlgorithmID *)PORT_ArenaZAlloc(arena, sizeof(SECAlgorithmID));
+    if (signAlg == NULL) {
+        return rv;
```
cipherboy, line 461: Throw OUTOFMEM

PR #468, opened 2020-04-03
Enable Shutdown of NSS and CryptoManager

cipherboy:
I'm still trying to decide the semantics I want for this and how far we want to go with it.

This pull request enables shutdown of a `CryptoManager` instance, shutting down NSS in the process. This first frees all remaining `NativeProxy` references, since they're (implicitly?) references to NSS values, which won't work after `NSS_Shutdown` is called. The result is that any attempt to use them will likely throw a `NullPointerException` _somewhere_. 

I did this because in #467, I wanted cleaner memory checking. In particular, I wanted to remove all `NativeProxy` references and clean up after NSS, so the only things that remain are true C level leaks.

Thoughts?

### DIFF
--- a/org/mozilla/jss/util/NativeProxy.java
+++ b/org/mozilla/jss/util/NativeProxy.java
@@ -127,7 +127,7 @@ public abstract class NativeProxy implements AutoCloseable
      */
     public final void close() throws Exception {
         try {
-            if (registry.remove(this)) {
+            if (registry.remove(this) && mPointer != null) {
                 releaseNativeResources();
             }
         } finally {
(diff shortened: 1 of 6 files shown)


org/mozilla/jss/util/NativeProxy.java
```diff
@@ -205,4 +205,32 @@ public synchronized static void assertRegistryEmpty() {
...
+    public synchronized static void unsafePurgeRegistry() throws Exception {
+        Exception first = null;
+        HashSet<NativeProxy> registryClone = new HashSet<NativeProxy>(registry.size());
+        registryClone.addAll(registry);
+
+        for (NativeProxy proxy : registryClone) {
+            try {
+                proxy.close();
+            } catch (Exception e) {
+                if (first == null) {
```
cipherboy, line 226: Needs a comment about what we're doing here.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -205,4 +205,32 @@ public synchronized static void assertRegistryEmpty() {
...
+
+    /**
+     * Unsafe: Purges all NativeProxies from memory.
+     *
+     * In the rare instances where we wish to shutdown an existing
+     * CryptoManager, all native proxies need to be cleared and freed.
+     * This will result in any lingering references to stop working, but
+     * should ensure that an application can recover from this scenario.
+     */
+    public synchronized static void unsafePurgeRegistry() throws Exception {
```
edewata, line 217: Instead of "unsafe" should we call it forceClose() or something like that?

cipherboy: Sure, works for me.

cipherboy: I've changed this to `purgeAllInRegistry`.

cipherboy:
Error in Fedora Sandbox unrelated, will be addressed in a separate PR.

PR #472, opened 2020-04-06
Add test for GlobalRefProxy

cipherboy:

This just adds a test case for `GlobalRefProxy`, since I've been hitting issues with it while trying to address some memory issues in `NativeProxy`. I don't think there is anything wrong with this class, but since I wrote the test, we might as well keep it.

### DIFF
--- a/org/mozilla/jss/tests/TestGlobalReference.java
+++ b/org/mozilla/jss/tests/TestGlobalReference.java
@@ -1,0 +1,35 @@
+package org.mozilla.jss.tests;
+
+import org.mozilla.jss.util.*;
+
+public class TestGlobalReference {
+    public static void main(String[] args) throws Exception {
+        String arg = "Something";
+
+        for (int i = 0; i < 100; i++) {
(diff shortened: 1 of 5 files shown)


org/mozilla/jss/tests/TestGlobalReference.java
```diff
@@ -0,0 +1,23 @@
...
+
+public class TestGlobalReference {
+    public static void main(String[] args) throws Exception {
+        String arg = "Something";
+
+        for (int i = 0; i < 100; i++) {
+            GlobalRefProxy proxy = new GlobalRefProxy(arg);
+            proxy.close();
+            proxy.close();
+            proxy.clear();
```
edewata, line 13: Could you add a note describing what's being tested and the expected behavior? E.g. test calling `GlobalRefProxy.close()` multiple times, it should complete successfully.

cipherboy: Added.

PR #473, opened 2020-04-06
Improve Memory Management

cipherboy:
This is built on top of #472.

There are a few new fixes:
 - `PK11Cert`, `PK11Cipher`, `PK11Key`, `PK11MessageDigest`, and `PK11Signature` all now have `AutoCloseable` for their internal CertProxies.
 - Fix `NativeProxy` registry tracking, candidate for backport.
 - Improve `SSLFDProxy`'s `globalRef` access.
 - Fix handling of `NULL` pointers in various places.

### DIFF
--- a/org/mozilla/jss/nss/PR.c
+++ b/org/mozilla/jss/nss/PR.c
@@ -90,7 +90,7 @@ Java_org_mozilla_jss_nss_PR_NewBufferPRFD(JNIEnv *env, jclass clazz,
 }
 
 JNIEXPORT int JNICALL
-Java_org_mozilla_jss_nss_PR_Close(JNIEnv *env, jclass clazz, jobject fd)
+Java_org_mozilla_jss_nss_PR_Close(JNIEnv *env, jclass clazz, jobject fd, jboolean clear)
 {
     PRFileDesc *real_fd = NULL;
 
@@ -106,7 +106,7 @@ Java_org_mozilla_jss_nss_PR_Close(JNIEnv *env, jclass clazz, jobject fd)
     }
 
     PRStatus ret = PR_Close(real_fd);
-    if (ret == PR_SUCCESS) {
+    if (ret == PR_SUCCESS && clear == JNI_TRUE) {
         JSS_clearPtrFromProxy(env, fd);
     }
 
--- a/org/mozilla/jss/nss/PR.java
+++ b/org/mozilla/jss/nss/PR.java
@@ -66,20 +66,38 @@ public class PR {
                                                  BufferProxy write_buf,
                                                  byte[] peer_info);
     /**
-     * Close an existing PRFDProxy.
+     * Close an existing PRFDProxy, clearing the pointer if successful.
      *
      * See also: PR_Close in /usr/include/nspr4/prio.h
      */
-    public static native int Close(PRFDProxy fd);
+    public static int Close(PRFDProxy fd) {
+        return Close(fd, true);
+    }
+
+    /**
+     * Close an existing PRFDProxy with an option to clear the pointer.
+     *
+     * See also: PR_Close in /usr/include/nspr4/prio.h
+     */
+    public static native int Close(PRFDProxy fd, boolean clear);
 
     /**
      * Close an existing SSLFDProxy.
      *
      * See also: org.mozilla.jss.nss.PR.Close
-     *           org.mozilla.jss.nss.SSLFD.releaseNativeResources
+     *           org.mozilla.jss.nss.SSLFDProxy.releaseNativeResources
      */
-    public static int Close(SSLFDProxy fd) throws Exception {
-        int ret = PR.Close((PRFDProxy) fd);
+    public synchronized static int Close(SSLFDProxy fd) throws Exception {
+        if (fd.isNull()) {
+            return SUCCESS;
+        }
+
+        SSL.RemoveCallbacks(fd);
+
+        // Because a SSLFDProxy instance needs to free other native resources,
+        // we can't release the pointer here. Instead, let NativeProxy.close()
+        // handle clearing the PRFileDesc pointer.
+        int ret = PR.Close((PRFDProxy) fd, false);
         fd.close();
         return ret;
     }
--- a/org/mozilla/jss/util/NativeProxy.java
+++ b/org/mozilla/jss/util/NativeProxy.java
@@ -9,8 +9,11 @@ import java.util.HashSet;
 import java.lang.AutoCloseable;
 import java.lang.Thread;
 import java.util.Arrays;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.mozilla.jss.CryptoManager;
+import org.mozilla.jss.netscape.security.util.Utils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -51,10 +54,13 @@ public abstract class NativeProxy implements AutoCloseable
      */
     protected NativeProxy(byte[] pointer, boolean track) {
         mPointer = pointer;
+        mHashCode = registryIndex.getAndIncrement();
+        if (mPointer != null) {
+            mHashCode += Arrays.hashCode(mPointer);
+        }
 
         if (track) {
             assert(pointer != null);
-
             registry.add(this);
 
             if (saveStacktraces) {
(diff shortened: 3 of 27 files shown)


org/mozilla/jss/util/NativeProxy.java
```diff
@@ -125,11 +140,13 @@ protected void finalize() throws Throwable {
...
      */
-    public final void close() throws Exception {
+    public synchronized final void close() throws Exception {
         try {
-            if (registry.remove(this)) {
+            registry.remove(this);
+            if (mPointer != null) {
                 releaseNativeResources();
             }
+            registry.remove(this);
```
cipherboy, line 149: Duplicate line.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -144,15 +161,16 @@ public final void close() throws Exception {
      *
      * See also: JSS_clearPtrFromProxy(...) in jssutil.h
      */
-    public final void clear() {
-        this.mPointer = null;
+    public synchronized final void clear() {
         registry.remove(this);
+        this.mPointer = null;
```
cipherboy, line 166: Noisy diff.

org/mozilla/jss/nss/PR.java
```diff
@@ -66,20 +66,38 @@ public static native PRFDProxy NewBufferPRFD(BufferProxy read_buf,
...
     /**
      * Close an existing SSLFDProxy.
      *
      * See also: org.mozilla.jss.nss.PR.Close
-     *           org.mozilla.jss.nss.SSLFD.releaseNativeResources
+     *           org.mozilla.jss.nss.SSLFDProxy.releaseNativeResources
      */
-    public static int Close(SSLFDProxy fd) throws Exception {
-        int ret = PR.Close((PRFDProxy) fd);
+    public synchronized static int Close(SSLFDProxy fd) throws Exception {
```
edewata, line 90: How about we convert this method into SSLFDProxy.close()?

```
class PRFDProxy extends NativeProxy {
    public void close() {
        PR.Close(this); // don't call JSS_clearPtrFromProxy()
        super.close(); // release native resources
    }
}

class SSLFDProxy extends PRFDProxy {
    public void close() {
        if (!fd.isNull()) {
            SSL.RemoveCallbacks(fd); 
        }
        super.close();
    }
}
```

org/mozilla/jss/nss/PR.c
```diff
@@ -106,7 +106,7 @@ Java_org_mozilla_jss_nss_PR_Close(JNIEnv *env, jclass clazz, jobject fd)
     }
 
     PRStatus ret = PR_Close(real_fd);
-    if (ret == PR_SUCCESS) {
+    if (ret == PR_SUCCESS && clear == JNI_TRUE) {
```
edewata, line 109: It looks like this condition only happens in JSS tests. If there is no real usage where `clear == JNI_TRUE` we might as well simplify the code to no longer call `JSS_clearPtrFromProxy()`. See also my other comment about not calling `JSS_clearPtrFromProxy()`.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -173,6 +197,21 @@ public final void clear() {
      * releaseNativeResources() gets called.
      */
     static HashSet<NativeProxy> registry = new HashSet<NativeProxy>();
+    static int registryIndex;
```
edewata, line 200: We probably can use `AtomicInteger`:
```
static AtomicInteger counter = new AtomicInteger();
int index;

protected NativeProxy(byte[] pointer, boolean track) {
    index = counter.getAndIncrement();
}
public int hashCode() {
    return index;
}
```

org/mozilla/jss/nss/PR.java
cipherboy: `PR` is static class and contains wrappers for NSPR functions beginning with the prefix `PR_`, of which `PR_Close()` is one. Since `SSLFDProxy` has the same native pointer type as `PRFDProxy` (a `PRFileDesc *` -- the difference is the former has been wrapped by a call to `SSL.ImportFD(...)`) there is no difference. All `PR_` functions work, because NSS creates another `PRFileDesc` layer for SSL, on top of the existing one. We need to do a little more work in `PR_Close` on a `SSLFDProxy` instance, because we need to invoke the `SSLFDProxy.close(...)` method as well, which we don't necessarily want to do if it is `NULL`. 

On the other hand `PRProxy.close()` and `SSLFDProxy.close()` take no arguments and are part of the `NativeProxy`+`AutoCloseable` interface, which free the proxy and clear the `mPointer` instance.

----

Our use of `PR_Close(PRFileDesc *)` is limited mostly to `SSLFDProxy` instances, but we do test it on files as well, in the test suite.

cipherboy: Namely, `PRFDProxy` doesn't invoke anything actually -- but we explicitly call `PR.Close(...)` in SSLEngine's cleanup() handler. The issue is we can't call this strictly a "free" method -- closing a `PRFDProxy` via `PR_Close(...)` might send data over the network if the TLS connection hasn't properly terminated. We don't necessarily want to do that (nor do we necessary have space in our buffers, ... &c), hence why `PRFDProxy.close()` doesn't do anything and why `SSLFDProxy.close()` only (tries to) free the `GlobalRefProxy` self-pointer.

org/mozilla/jss/util/NativeProxy.java
cipherboy: Yeah, `AtomicInteger` is probably the best bet. The [old `NativeProxy`](https://github.com/dogtagpki/jss/blob/v4.5.0/org/mozilla/jss/util/NativeProxy.java#L115-L137) used a random integer. :unicorn:

org/mozilla/jss/nss/PR.java
cipherboy: Keep in mind, `nss.PR` is meant to be a thin wrapper over NSPR functions. Just like `nss.SSL` is a thin wrapper over `SSL_` functions. We're likely going to move as much of this behind a single, large JNI call in an optimized SSLEngine implementation. Some of these details will thus go away.

edewata: I'm not really familiar with all the details, so I'll let you decide :) And feel free to merge as is since it's already approved. Anything we discussed here is for a possible future enhancement.

To my understanding NSPR only has a `PR_Close(PRFileDesc*)` funcation which is already wrapped by `PR.Close(PRFDProxy)`. The `PR.Close(SSLFDProxy)` on the other hand does not have a corresponding native code in NSPR. Also, that method has a dependency on `SSL` class which is a wrapper for NSS. From design perspective I don't think a wrapper for NSPR should depend on a wrapper for NSS.

I think if we want to do something similar to `PR.Close(PRFDProxy)`, we probably should add an `SSL.Close(SSLFDProxy)` instead. Or we can just add an `SSLFDProxy.close()` like I described earlier. If we don't want to add `PRFDProxy.close()`, we probably can do something like this:

```
class SSLFDProxy extends PRFDProxy {
    public void close() {
        if (fd.isNull()) {
            return;
        }
        SSL.RemoveCallbacks(fd); 
        PR.Close(this, false);
        releaseNativeResources();
    }
}
```
It should be functionally identical to your code, but it will not create an NSPR -> NSS dependency issue.

org/mozilla/jss/util/NativeProxy.java
```diff
@@ -51,10 +53,13 @@ public NativeProxy(byte[] pointer) {
      */
     protected NativeProxy(byte[] pointer, boolean track) {
         mPointer = pointer;
+        mHashCode = registryIndex.getAndIncrement();
+        if (mPointer != null) {
+            mHashCode = Objects.hash(mHashCode, Arrays.hashCode(mPointer));
```
cipherboy, line 58: @edewata  -- I think this might be better and provide more random distribution of `mHashCode`.

org/mozilla/jss/nss/PR.c
cipherboy: Yeah I think we could update this in a separate PR and simplify how these are handled.

cipherboy:
@edewata What do you think of the new `mHashCode` implementation? We have time now that we could ship our existing solution and then test `master` and make sure this is sufficient for the future. Thoughts?

cipherboy:
@edewata This now includes the patch from my other repo: https://github.com/cipherboy/jss/pull/5

The last patch is a little excessive (and not necessary in the 0day) -- nothing else frees those resources and we guarantee `releaseNativeResources` only gets called once. `SSLSocket` is the only place where this gets (ab)used under the new resource releasing model.

edewata:
LGTM

cipherboy:
@edewata Thanks!

PR #474, opened 2020-04-06
Enable BadSSL Tests

cipherboy:
Enables a new CMake variable, `WITH_INTERNET`, to enable tests which require an internet connection. This currently includes only the BadSSL test, but could include other tests in the future (such as some under `org.mozilla.jss.ssl` currently).


cipherboy:
[Output](https://github.com/dogtagpki/jss/pull/474/checks?check_run_id=565920737#step:3:2000) is as expected. It doesn't invoke [during RPM build](https://github.com/dogtagpki/jss/runs/565920580?check_suite_focus=true#step:3:3609). So, merging. :-)

PR #476, opened 2020-04-07
Move from _NETSCAPE_ to _NSS_ PKCS#11 constants

cipherboy:
In NSS v3.52, support is coming for PKCS#11 v3.0. This deprecates the
`_NETSCAPE_` namespace for PKCS#11 constants in favor of `_NSS_`. The few
remaining `_NETSCAPE_` constants will be moved to `_NSS_`. We only use one,
`CKM_NETSCAPE_PBE_SHA1_DES_CBC`. Add an `#ifdef` for compatibility with the
new preferred name.

See also: [`moz-bz#1603628`](https://bugzilla.mozilla.org/show_bug.cgi?id=1603628)

### DIFF
--- a/org/mozilla/jss/pkcs11/PK11SymKey.c
+++ b/org/mozilla/jss/pkcs11/PK11SymKey.c
@@ -14,6 +14,11 @@
 #include <jssutil.h>
 #include "pk11util.h"
 
+/* For PKCS#11 v3.0 compatibility */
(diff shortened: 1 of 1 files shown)


cipherboy:
This probably also warrants an update to `PKCS11Constants.java` and its generation to do both `_NETSCAPE_` and `_NSS_` constants in the future.

cipherboy:
`PKCS11Constants.java` generation can be updated in a later PR.

PR #489, opened 2020-04-14
SSLEngine - PostHandshake Authentication / Renegotation

cipherboy:
This should address some of the certificate issues we've seen in PKI CI. This is in several parts:

 - Expose a bunch of new SSL options
 - Expose `SSL_ReHandshake` (Necessary for renegotiation in TLS < 1.3)
 - Expose `SSL_SendCertificateRequest` (necessary for PHA in TLS >= 1.3). Currently experimental, added in NSS v3.44.
 - Expose the "real" handshake completion status in `SSLFDProxy`. Note that `SecurityStatusResult.on` reports `on` before the finish message is sent/received.
 - Add support for NSS configuration, PHA in `JSSEngine`/`JSSEngineReferenceImpl`.
 - Test everything! :D

### DIFF
--- a/org/mozilla/jss/nss/SSL.c
+++ b/org/mozilla/jss/nss/SSL.c
@@ -2,6 +2,7 @@
 #include <nss.h>
 #include <ssl.h>
 #include <sslerr.h>
+#include <sslexp.h>
 #include <limits.h>
 #include <stdint.h>
 #include <jni.h>
@@ -765,6 +766,22 @@ Java_org_mozilla_jss_nss_SSL_PeerCertificateChain(JNIEnv *env, jclass clazz,
     }
 
     return JSS_PK11_wrapCertChain(env, &chain);
+}
+
+JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_nss_SSL_SendCertificateRequest(JNIEnv *env, jclass clazz,
+    jobject fd)
+{
+    PRFileDesc *real_fd = NULL;
+
+    PR_ASSERT(env != NULL && fd != NULL);
+    PR_SetError(0, 0);
+
+    if (JSS_PR_getPRFileDesc(env, fd, &real_fd) != PR_SUCCESS) {
+        return SECFailure;
+    }
+
+    return SSL_SendCertificateRequest(real_fd);
 }
 
 JNIEXPORT jint JNICALL
--- a/org/mozilla/jss/nss/SSL.c
+++ b/org/mozilla/jss/nss/SSL.c
@@ -871,3 +871,9 @@ Java_org_mozilla_jss_nss_SSL_getSSLSECWouldBlock(JNIEnv *env, jclass clazz)
 {
     return SECWouldBlock;
 }
+
+JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_nss_SSL_getSSLEnablePostHandshakeAuth(JNIEnv *env, jclass clazz)
+{
+    return SSL_ENABLE_POST_HANDSHAKE_AUTH;
+}
--- a/org/mozilla/jss/nss/SSL.c
+++ b/org/mozilla/jss/nss/SSL.c
@@ -973,3 +973,27 @@ Java_org_mozilla_jss_nss_SSL_getSSLEnableFallbackSCSV(JNIEnv *env, jclass clazz)
 {
     return SSL_ENABLE_FALLBACK_SCSV;
 }
+
+JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_nss_SSL_getSSLRequireNever(JNIEnv *env, jclass clazz)
+{
+    return SSL_REQUIRE_NEVER;
+}
+
+JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_nss_SSL_getSSLRequireAlways(JNIEnv *env, jclass clazz)
+{
+    return SSL_REQUIRE_ALWAYS;
+}
+
+JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_nss_SSL_getSSLRequireFirstHandshake(JNIEnv *env, jclass clazz)
+{
+    return SSL_REQUIRE_FIRST_HANDSHAKE;
+}
+
+JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_nss_SSL_getSSLRequireNoError(JNIEnv *env, jclass clazz)
+{
+    return SSL_REQUIRE_NO_ERROR;
+}
--- a/org/mozilla/jss/nss/SSL.c
+++ b/org/mozilla/jss/nss/SSL.c
@@ -859,6 +859,27 @@ Java_org_mozilla_jss_nss_SSL_RemoveCallbacks(JNIEnv *env, jclass clazz,
 }
 
 JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_nss_SSL_EnableHandshakeCallback(JNIEnv *env, jclass clazz,
+    jobject fd)
+{
+    PRFileDesc *real_fd = NULL;
+    jobject fd_ref = NULL;
+
+    PR_ASSERT(env != NULL && fd != NULL);
+    PR_SetError(0, 0);
+
+    if (JSS_PR_getPRFileDesc(env, fd, &real_fd) != PR_SUCCESS) {
+        return SECFailure;
+    }
+
+    if (JSS_NSS_getGlobalRef(env, fd, &fd_ref) != PR_SUCCESS) {
+        return SECFailure;
+    }
+
+    return SSL_HandshakeCallback(real_fd, JSSL_SSLFDHandshakeComplete, fd_ref);
+}
+
+JNIEXPORT jint JNICALL
 Java_org_mozilla_jss_nss_SSL_getSSLRequestCertificate(JNIEnv *env, jclass clazz)
 {
     return SSL_REQUEST_CERTIFICATE;
--- a/org/mozilla/jss/nss/SSL.c
+++ b/org/mozilla/jss/nss/SSL.c
@@ -785,6 +785,22 @@ Java_org_mozilla_jss_nss_SSL_SendCertificateRequest(JNIEnv *env, jclass clazz,
 }
 
 JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_nss_SSL_KeyUpdate(JNIEnv *env, jclass clazz,
+    jobject fd, jboolean requestUpdate)
+{
+    PRFileDesc *real_fd = NULL;
+
+    PR_ASSERT(env != NULL && fd != NULL);
+    PR_SetError(0, 0);
+
+    if (JSS_PR_getPRFileDesc(env, fd, &real_fd) != PR_SUCCESS) {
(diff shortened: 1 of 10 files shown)


org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -804,6 +825,31 @@ public boolean isOutboundDone() {
...
     public abstract SecurityStatusResult getStatus();
 
+    /**
+     * Gets the default configuration.
+     */
+    public HashMap<Integer, Integer> getDefaultConfiguration() {
+        HashMap<Integer, Integer> result = new HashMap<Integer, Integer>();
+        result.put(SSL.ENABLE_RENEGOTIATION, SSL.RENEGOTIATE_REQUIRES_XTN);
+        result.put(SSL.REQUIRE_SAFE_NEGOTIATION, 1);
+        result.put(28, 1);
```
cipherboy, line 835: Need to add the SSL option here.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -267,11 +276,17 @@ private void initServer() {
             throw new RuntimeException("Unable to configure SSL_REQUEST_CERTIFICATE option: " + errorText(PR.GetError()));
         }
 
-        if (SSL.OptionSet(ssl_fd, SSL.REQUIRE_CERTIFICATE, need_client_auth ? 1 : 0) == SSL.SECFailure) {
+        if (SSL.OptionSet(ssl_fd, SSL.REQUIRE_CERTIFICATE, need_client_auth ? 3 : 0) == SSL.SECFailure) {
```
cipherboy, line 279: Need to include these values as well in `nss.SSL`.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -395,15 +422,48 @@ public void beginHandshake() {
...
+
+            // Reset the handshake status, using the new socket and
+            // configuration which was just created.
+            if (SSL.ResetHandshake(ssl_fd, as_server) == SSL.SECFailure) {
+                throw new RuntimeException("Unable to begin handshake: " + errorText(PR.GetError()));
+            }
+        } else {
+            unwrap = !unwrap;
+
+            if (session.getSSLVersion() == SSLVersion.TLS_1_3) {
```
cipherboy, line 438: Needs a comment about what we're doing here.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -395,15 +422,48 @@ public void beginHandshake() {
...
+
+            SSL.ForceHandshake(ssl_fd);
         }
 
-        // Always, reset the handshake status, using the existing
-        // socket and configuration (which might've been just created).
-        if (SSL.ResetHandshake(ssl_fd, as_server) == SSL.SECFailure) {
-            throw new RuntimeException("Unable to begin handshake: " + errorText(PR.GetError()));
+        if (unwrap) {
+            // Update handshake status; client initiates connection, so wait
```
cipherboy, line 454: These comments need to be updated.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -619,8 +679,7 @@ private void updateHandshakeState() {
         // currently on, update our handshake status. This happens even if
         // we later exit before calling SSL.ForceHandshake() so that we can
         // see what the session data contains.
-        SecurityStatusResult preHandshakeStatus = getStatus();
```
cipherboy, line 622: Could probably drop these calls.

org/mozilla/jss/ssl/javax/JSSSession.java
```diff
@@ -50,8 +51,9 @@ public JSSEngine getEngine() {
     }
 
     public SSLChannelInfo getChannelInfo() {
-        if (parent.getSSLFDProxy() != null) {
-            return SSL.GetChannelInfo(parent.getSSLFDProxy());
+        SSLFDProxy ssl_fd = parent.getSSLFDProxy();
+        if (ssl_fd.handshakeComplete) {
```
cipherboy, line 55: Needs `null` check as well.

org/mozilla/jss/ssl/javax/JSSSession.java
```diff
@@ -184,6 +193,8 @@ protected void setPeerCertificates(Certificate[] new_certs) {
     }
 
     public X509Certificate[] getPeerCertificateChain() throws SSLPeerUnverifiedException {
+        refreshData();
```
cipherboy, line 196: Remove this since they're unused currently.

org/mozilla/jss/nss/SSL.c
```diff
@@ -598,6 +599,23 @@ Java_org_mozilla_jss_nss_SSL_ResetHandshake(JNIEnv *env, jclass clazz,
...
+    PR_ASSERT(env != NULL && fd != NULL);
+    PR_SetError(0, 0);
+
+    if (JSS_PR_getPRFileDesc(env, fd, &real_fd) != PR_SUCCESS) {
+        return SECFailure;
+    }
+
+    return SSL_ReHandshake(real_fd, flushCache);
+}
+
```
cipherboy, line 617: Spare space.

org/mozilla/jss/nss/SSL.java
```diff
@@ -52,6 +52,107 @@
...
+    public static final int REQUIRE_FIRST_HANDSHAKE = getSSLRequireFirstHandshake();
+
+    /**
+     * Value for requiring a certificate but not erring if the peer doesn't
+     * provide one. Value for use with SSL_REQUIRE_CERTIFICATE with OptionGet
+     * and OptionSet.
+     *
+     * See also: SSL_REQUIRE_NO_ERROR in /usr/include/nss3/ssl.h
+     */
+    public static final int REQUIRE_NO_ERROR = getSSLRequireNever();
```
cipherboy, line 154: `getSSLRequireNoError()`

org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -804,6 +825,31 @@ public boolean isOutboundDone() {
...
+        result.put(SSL.ENABLE_RENEGOTIATION, SSL.RENEGOTIATE_REQUIRES_XTN);
+        result.put(SSL.REQUIRE_SAFE_NEGOTIATION, 1);
+        result.put(SSL.ENABLE_FALLBACK_SCSV, 1);
+        return result;
+    }
+
+    /**
+     * Updates the configuration with the given value.
+     */
+    public void addConfiguration(int key, int value) {
```
edewata, line 842: Should we call it `setConfiguration()` instead?

cipherboy: Ah I was going to add a `setConfiguration(HashMap<...> config)` function, thanks!

org/mozilla/jss/nss/SSL.c
```diff
@@ -842,6 +891,27 @@ Java_org_mozilla_jss_nss_SSL_RemoveCallbacks(JNIEnv *env, jclass clazz,
...
 }
 
+JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_nss_SSL_EnableHandshakeCallback(JNIEnv *env, jclass clazz,
+    jobject fd)
+{
+    PRFileDesc *real_fd = NULL;
+    jobject fd_ref = NULL;
+
+    PR_ASSERT(env != NULL && fd != NULL);
```
jmagne, line 901: Question, since this block of code seems to happen in a bunch of methods, would having a convenience method be of use?

org/mozilla/jss/nss/SSLFDProxy.c
```diff
@@ -240,3 +240,33 @@ JSSL_SSLFDCertSelectionCallback(void *arg,
...
+    }
+
+    handshakeCompleteField = (*env)->GetFieldID(env, sslfdProxyClass,
+                                                "handshakeComplete", "Z");
+    if (handshakeCompleteField == NULL) {
+        return;
+    }
+
+    (*env)->SetBooleanField(env, sslfd_proxy, handshakeCompleteField, JNI_TRUE);
+}
```
jmagne, line 272: Just a question here. I see this callback just sets some boolean when the handshake it complete. Would there be any case to have a java method receive this callback as well?

cipherboy: For one, there's no callback in the SSLEngine specification, so adding it would be surprising. I don't think there's value in adding it because... 

SSLEngine is non-blocking. At each step of the initial handshake, there's a handshake status. It flips back and forth between Wrap and Unwrap as necessary. When the handshake is finished (i.e., this callback fires), it returns Finished. After that, it returns Not Handshaking. So the caller who would be expecting a handshake complete callback execution could alternatively wait for Finished and execute it then. That's the beauty of non-blocking SSLEngine over a SSLengine: it needs fewer callbacks. :-)

org/mozilla/jss/nss/SSL.c
cipherboy: A lot of code uses this pattern. I don't think you can use a convenience method because in debug mode, it aborts. It'd be a candidate for a macro in the future (because we need to return from this method), but a lot of existing code also needs to be replaced. I think we need to replace it with a macro that checks these for `NULL` and returns `NullPointerException`.

cipherboy: They could also be removed, as the JNI contract should guarantee `env` should be non-`NULL`. If `env` is `NULL`, we have no safe way of handling it other than aborting. So `env != NULL` should probably be removed from all functions and we can accept the segfault that occurs in the unlikely scenario that `env` is ever `NULL`. We're essentially doing that now as `PR_ASSERT` only executes in debug mode. 

That leaves `fd != NULL`. This (and a bunch of other cases where it is used) should be replaced by a macro that fails with NPE if/when this is `NULL`. It is up to the caller to ensure non-NULL of `fd` in this scenario.

org/mozilla/jss/nss/SSLFDProxy.c
cipherboy: And, `SSL_SecurityStatus()` reports `on > 0` even when the handshake itself hasn't completed. This is most commonly when the `FINISHED` message hasn't yet been sent but the server has approved a particular protocol version / cipher suite pair. The `HandshakeCallback` delays firing until the `FINISHED` message is on the wire, making it better than `SecurityStatus.on > 0` (which is what we've replaced in this PR).

Since our call graph looks like:

 - `SSLEngine.wrap` or `SSLEngine.unwrap`
 - `JSSEngine.updateHandshakeState()`
 - `SSL_ForceHandshake()`
 - ... NSS stuff ...
 - This handshake complete handler

We'd prefer to set a boolean and unwind back to `JSSEngine.UpdateHandshakeState` rather than nesting further; the next thing we need to do is in `UpdateHandshakeState` once `SSL_ForceHandshake` returns.

cipherboy:
Thanks all! Merging and then #501 will be rebased.

PR #501, opened 2020-04-21
SSLEngine - Handle Large wrap/unwrap Calls

cipherboy:
One of the issues I identified when testing JSSEngine in IPA and PKI is handling of calls to wrap/unwrap with large wrap data. Here, large is defined as "exceeding the size of the internal buffers".

This PR is in three parts:

 - Add a test case for calls to wrap/unwrap with large buffers. 
 - Allow dummy `PR.Write` calls.
 - Fix wrapping/unwrapping with large buffers.
 - Fix calls to `PR.Write` when entire buffer isn't consumed. 

In particular, this is [consistent](https://github.com/apache/tomcat/blob/master/java/org/apache/tomcat/util/net/openssl/OpenSSLEngine.java#L458-L486) with Tomcat's OpenSSL native SSLEngine.

Note that dummy `PR.Write` calls are consistent with what [NSS does internally](https://github.com/nss-dev/nss/blob/master/lib/ssl/sslcon.c#L138). Sometimes we need to force NSS to push data out but we have no more data to send -- it is already in NSS's internal buffers and needs to get written to our wrapped buffer-backed `PRFileDesc *`.

### DIFF
--- a/org/mozilla/jss/nss/PR.c
+++ b/org/mozilla/jss/nss/PR.c
@@ -4,6 +4,7 @@
 #include <jni.h>
 
 #include "jssutil.h"
+#include "jss_exceptions.h"
 #include "PRFDProxy.h"
 #include "BufferProxy.h"
 #include "BufferPRFD.h"
@@ -208,12 +209,12 @@ Java_org_mozilla_jss_nss_PR_Write(JNIEnv *env, jclass clazz, jobject fd,
     jbyteArray buf)
 {
     PRFileDesc *real_fd = NULL;
-    unsigned int real_length = 0;
     int max_length = 0;
+    uint8_t dummy_buffer = 0;
     uint8_t *buffer = NULL;
     int result = 0;
 
-    PR_ASSERT(env != NULL && fd != NULL && buf != NULL);
+    PR_ASSERT(env != NULL && fd != NULL);
     PR_SetError(0, 0);
 
     if (JSS_PR_getPRFileDesc(env, fd, &real_fd) != PR_SUCCESS) {
@@ -222,20 +223,24 @@ Java_org_mozilla_jss_nss_PR_Write(JNIEnv *env, jclass clazz, jobject fd,
 
     PR_ASSERT(real_fd != NULL);
 
-    real_length = (*env)->GetArrayLength(env, buf);
-    if (real_length > INT_MAX) {
-        max_length = INT_MAX;
-    } else {
-        max_length = (int)(real_length % INT_MAX);
-    }
+    if (buf != NULL) {
+        max_length = (*env)->GetArrayLength(env, buf);
 
-    buffer = (uint8_t*)((*env)->GetByteArrayElements(env, buf, NULL));
-    if (buffer == NULL) {
-        return 0;
+        buffer = (uint8_t*)((*env)->GetByteArrayElements(env, buf, NULL));
+        if (buffer == NULL) {
+            ASSERT_OUTOFMEM(env);
+            return 0;
+        }
+    } else {
+        buffer = &dummy_buffer;
+        max_length = 0;
     }
 
     result = PR_Write(real_fd, buffer, max_length);
-    (*env)->ReleaseByteArrayElements(env, buf, (jbyte *)buffer, JNI_ABORT);
+
+    if (buf != NULL && buffer != &dummy_buffer) {
+        (*env)->ReleaseByteArrayElements(env, buf, (jbyte *)buffer, JNI_ABORT);
+    }
 
     return result;
 }
@@ -280,7 +285,6 @@ Java_org_mozilla_jss_nss_PR_Send(JNIEnv *env, jclass clazz, jobject fd,
     jbyteArray buf, jint flags, jlong timeout)
 {
     PRFileDesc *real_fd = NULL;
-    unsigned int real_length = 0;
     int max_length = 0;
     uint8_t *buffer = NULL;
     PRIntervalTime timeout_interval = (PRIntervalTime)(timeout % UINT32_MAX);
@@ -296,12 +300,7 @@ Java_org_mozilla_jss_nss_PR_Send(JNIEnv *env, jclass clazz, jobject fd,
 
     PR_ASSERT(real_fd != NULL);
 
-    real_length = (*env)->GetArrayLength(env, buf);
-    if (real_length > INT_MAX) {
-        max_length = INT_MAX;
-    } else {
-        max_length = (int)(real_length % INT_MAX);
-    }
+    max_length = (*env)->GetArrayLength(env, buf);
 
     buffer = (uint8_t*)((*env)->GetByteArrayElements(env, buf, NULL));
     if (buffer == NULL) {
--- a/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -50,6 +50,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
 
     private int unknown_state_count;
     private boolean step_handshake;
+    private boolean returned_finished;
 
     private SSLException ssl_exception;
     private boolean seen_exception;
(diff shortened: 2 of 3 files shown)


org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -790,58 +777,63 @@ public SSLEngineResult unwrap(ByteBuffer src, ByteBuffer[] dsts, int offset, int
...
+        // do-while loop.
 
-        // When we have data from src, write it to read_buf.
-        if (wire_data > 0) {
-            byte[] wire_buffer = new byte[wire_data];
-            src.get(wire_buffer);
-            int written = (int) Buffer.Write(read_buf, wire_buffer);
+        // wire_data is the number of bytes from src we've written into
+        // read_buf. This is bounded above by src.capcity but also the
+        // free space left in read_buf to write to. Allows us to size the
```
cipherboy, line 786: This isn't actually correct.

cipherboy:
@edewata OK, this is ready for review.

org/mozilla/jss/nss/PR.c
```diff
@@ -222,20 +224,31 @@ Java_org_mozilla_jss_nss_PR_Write(JNIEnv *env, jclass clazz, jobject fd,
...
-    } else {
-        max_length = (int)(real_length % INT_MAX);
-    }
+    if (buf != NULL) {
+        real_length = (*env)->GetArrayLength(env, buf);
+        if (real_length > INT_MAX) {
+            max_length = INT_MAX;
+        } else {
+            max_length = (int)(real_length % INT_MAX);
+        }
```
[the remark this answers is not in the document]
cipherboy: Ahh I was thinking `jsize` was `#define ssize_t` instead of `jint`.

cipherboy:
Thanks @edewata!

cipherboy:
Things to do based on #489 and this PR merge order:

 - [ ] Make sure `debug` wrapping in test suite
 - [ ] Update `returned_finished` to be reset during PHA.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -50,6 +50,7 @@
 
     private int unknown_state_count;
     private boolean step_handshake;
+    private boolean returned_finished;
```
cipherboy, line 53: @edewata Can you think of anything more elegant for this? Or is adding a boolean flag and tracking it fine? Maybe I should split it into a `updateHandshakePre()` and a `updateHandshakeState()`? -- `Pre` only getting called once and `State` getting called only when actually handshaking? Hmm...

edewata: I don't really have an opinion on this. If it works then let's keep it :)
We can always refactor later if we find a better way.

I've ACKed this PR, feel free to merge, but I'd prefer @jmagne to take a look at this as well, either before or after merging.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -811,58 +817,68 @@ public SSLEngineResult unwrap(ByteBuffer src, ByteBuffer[] dsts, int offset, int
...
+                //     to write any more data.
+                //  2. SOCKET_SHUTDOWN_ERRORs are safe, because if the
+                //     underling cause was fatal, we'd catch it after exiting
+                //     the do-while loop, in checkSSLAlerts().
+                if (error != 0 && error != PRErrors.WOULD_BLOCK_ERROR && error != PRErrors.SOCKET_SHUTDOWN_ERROR) {
                     ssl_exception = new SSLException("Unexpected return from PR.Read(): " + errorText(error));
                     seen_exception = true;
                 }
             }
+        } while (this_src_write != 0 || this_dst_write != 0);
```
jmagne, line 877: I"m assuming that this_src_write and ths_dst_write will eventually become exactly zero?

cipherboy: Yeah, I need to update the docs pr (#428) with this. But the theory is this: Let's take `wrap` as an example. We need to take data from `src` and put it into `dst` if the handshake has finished. Otherwise if we're still handshaking we need to take whatever handshake data is made and puts it into `dst`. 

So let's take the handshaking scenario:

 - We'll be ignoring `src` because `PR.Write(...)` will fail until the handshake is complete.
 - So `UpdateHandshakeState()` will write data into the write buffer.
 - We'll drain data from the read buffer and write it into `dst`.

During this:

 - `this_src_write` will be 0 because `PR.Write(...)` returns 0 / `PR_WOULD_BLOCK` until handshake completes.
 - While we're still handshaking in _this_ step, `UpdateHandshakeState()` will write into the read buffer, which then goes into `dst`, incrementing `this_dst_write`.
 - When we run out of handshake data to write, `UpdateHandshakeState()` will quit writing data (and return `PR_WOULD_BLOCK`) and `this_dst_write` will go to data as the write buffer will be empty.

---

A similar thing applies when `src` is non-empty and we're not handshaking:

 - We'll be writing from `src` into the SSL `PRFileDesc *`. `PR.Write(...)` returns number of bytes written which ends up into `this_src_write`. As long as we're still writing data (and not getting an error / `PR_WOULD_BLOCK`), this'll be non-zero. It'll also continue being non-zero while there's space in the write buffer. However, this write buffer gets cleared out by the last step...
 - `UpdateHandshakeState()` will return, a no-op because we're not handshaking.
 - When there's src data and `PR.Write()` succeeds, there'll be data in the write buffer so we'll be able to put it in `dst` and `this_dst_write` will be non-zero. Otherwise, if `src` fails, `this_src_write` becomes 0 and there'll be nothing in the write buffer so `this_dst_write` will go to 0 as well.

---

In all helper methods, we guarantee that on error, 0 is returned instead of a less than zero value.

---

In both cases, when we run out of handshake and/or `src` data, we'll quit writing to the write buffer and `this_dst_write` will eventually go to 0 as well.

cipherboy:
Why are large buffers different? And what is large?

PoV of an app using SSLEngine:

Let's take `wrap` as an example. `SSLEngine.wrap(src, dst)` (*) takes data we want to encrypt (`src`) and encrypts it in the current TLS session to the peer. It is our job to actually _send_ to the peer, but it is the SSLEngine's job to encrypt it (into `dst`).

The contract here is, after the handshake completes, _everything_ from `src` should get written into `dst`.

Now, from the PoV of the SSLEngine... "large" is when the size of `src` _exceeds_ [`JSSEngine.BUFFER_SIZE`](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/ssl/javax/JSSEngine.html#BUFFER_SIZE). In particular, if we write a message of size `BUFFER_SIZE+1` via [`PR.Write(...)`](https://github.com/cipherboy/jss/blob/ssl-engine/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java#L901), it'll return `PR_WOULD_BLOCK` as the status, after having written approximately `BUFFER_SIZE` bytes (+/- padding of the underlying crypto system). It will never write all `BUFFER_SIZE` bytes. 

So what happens to these bytes we successfully sent? NSS encrypts them and sends them to its underlying `PRFileDesc` instance. In that case, its our [buffer-backed](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/ssl/javax/BufferPRFD.c) `PRFileDesc` -- which is a buffer of size `BUFFER_SIZE` bytes that's accessible from [`JSSEngine`](https://github.com/cipherboy/jss/blob/ssl-engine/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java#L48). The `JSSEngine` then [reads them from the `write_buf`](https://github.com/cipherboy/jss/blob/ssl-engine/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java#L901) and puts them into [`dst`](https://github.com/cipherboy/jss/blob/ssl-engine/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java#L1057).

This last step frees up space in `write_buf` because we're consuming those bytes -- NSS is now free to write into the buffer again.

What does this mean? In the code currently in JSS, we'd now exit. We'd have left bytes in `src` (that the application wasn't expecting!) and they'd be surprised to have to call `SSLEngine.wrap` again.

On the other hand, because we just freed up enough space to continue writing (via `PR.Write()`), we could handle that call ourselves. Hence, the do-while loop.

---

(*): technically its _`srcs`_ plural, but for our purposes it could be considered a single `src`. There's a single-`src` form as well.

PR #503, opened 2020-04-22
Minor improvement to GH actions

SilleBille:
Use matrix strategy to avoid redundant code. Easier to read and maintain CI code.

I wasn't sure if "test" was the right job name. But, let me know if you come up with a better name :wink: 

`Signed-off-by: Dinesh Prasanth M K <dmoluguw@redhat.com>`

### DIFF
--- a/.github/workflows/optional.yml
+++ b/.github/workflows/optional.yml
@@ -9,66 +9,24 @@ on:
     - master
 
 jobs:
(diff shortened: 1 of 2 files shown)


cipherboy:
Eh, test works as well as anything else. Thanks! I don't know why I didn't find matrix jobs earlier.

PR #515, opened 2020-04-27
Refactored CertificateChain

edewata:
The `CertificateChain` class has been modified to use
a `List` instead of a fixed array of `X509Certificates`.

Additional methods will be added later to manipulate
the certificates in the chain.

This allows the `CertificateChain` to be used to compose
new certificate chains based on the issuer's chain.


edewata:
Agreed. Thanks!

PR #520, opened 2020-04-28
Added PKCS7 constructor

edewata:
A new constructor has been added to create a `PKCS7` from
an array of certificates. This can be used to simplify
`CertificateChain.encode()` and some other code in PKI.


edewata:
Thanks!

PR #522, opened 2020-04-28
Added CertificateChain.sort()

edewata:
The CertificateChain.sort() has been added to sort the
certificates in the certificate chain.


edewata:
Thanks!

PR #526, opened 2020-04-29
JSSEngine - Initial ALPN support

cipherboy:
This pull requests introduces support for ALPN negotiation via a fixed list of next protocol identifiers.

Technically this support was introduced in the native SSLEngine in JDK 9, but support in Tomcat exists for older versions when advertised.

ALPN is a method of negotiating the next protocol after TLS. Most of the time this will be `http/1.1`, but could be `h2` or `spdy/2` depending on browser and server support. Without this, Tomcat is limited to `http/1.1` and cannot speak e.g., `http/2`.

--

Also, having a ALPN-capable implementation means we can go implement [TLS-ALPN-01 / RFC 8737](https://tools.ietf.org/html/rfc8737) if we wanted to.

### DIFF
--- a/org/mozilla/jss/nss/NextProtoResult.java
+++ b/org/mozilla/jss/nss/NextProtoResult.java
@@ -24,7 +24,7 @@ public class NextProtoResult {
     }
 
     public String getProtocol() {
-        if (protocol == null || protocol.length == 0) {
+        if (protocol == null) {
             return null;
         }
 
--- a/org/mozilla/jss/nss/NextProtoResult.java
+++ b/org/mozilla/jss/nss/NextProtoResult.java
@@ -1,0 +1,41 @@
+package org.mozilla.jss.nss;
+
+import java.lang.StringBuilder;
+import java.util.Arrays;
+
+import org.mozilla.jss.ssl.SSLNextProtoState;
+
+/**
+ * The fields in the NextProtoResult indicate whether a given SSL-enabled
+ * PRFileDesc has negotiated a next protocol (via ALPN) and if so, what it
+ * is.
+ *
+ * These object is returned by org.mozilla.jss.nss.SSL.GetNextProto(fd).
+ * This is a native method; note that updating the constructor will require
+ * modifying util/java_ids.h and nss/SSL.c
+ */
+public class NextProtoResult {
+    public SSLNextProtoState state;
+    public byte[] protocol;
+
+    public NextProtoResult(int state_value, byte[] protocol) {
+        state = SSLNextProtoState.valueOf(state_value);
+        this.protocol = protocol;
+    }
+
+    public String getProtocol() {
+        if (protocol == null || protocol.length == 0) {
+            return null;
+        }
+
+        return new String(protocol);
+    }
+
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("State: " + state + "\n");
+        sb.append("Protocol: " + getProtocol() + " ");
+        sb.append(Arrays.toString(protocol) + "\n");
+        return sb.toString();
+    }
+}
--- a/org/mozilla/jss/nss/SSL.c
+++ b/org/mozilla/jss/nss/SSL.c
@@ -817,6 +817,32 @@ Java_org_mozilla_jss_nss_SSL_KeyUpdate(JNIEnv *env, jclass clazz,
 }
 
 JNIEXPORT jint JNICALL
+Java_org_mozilla_jss_nss_SSL_SetNextProtoNeg(JNIEnv *env, jclass clazz,
+    jobject fd, jbyteArray wire_data)
+{
+    PRFileDesc *real_fd = NULL;
+    uint8_t *data = NULL;
+    size_t data_length = 0;
+    SECStatus ret = SECFailure;
+
+    PR_ASSERT(env != NULL && fd != NULL && wire_data != NULL);
+    PR_SetError(0, 0);
+
+    if (JSS_PR_getPRFileDesc(env, fd, &real_fd) != PR_SUCCESS) {
+        return ret;
+    }
+
+    if (!JSS_FromByteArray(env, wire_data, &data, &data_length)) {
+        return ret;
+    }
+
+    ret = SSL_SetNextProtoNego(real_fd, data, data_length);
+    free(data);
+
+    return ret;
+}
+
+JNIEXPORT jint JNICALL
 Java_org_mozilla_jss_nss_SSL_AttachClientCertCallback(JNIEnv *env, jclass clazz,
     jobject fd)
 {
--- a/org/mozilla/jss/nss/SSL.c
+++ b/org/mozilla/jss/nss/SSL.c
@@ -140,6 +140,40 @@ finish:
     return result;
 }
 
+jobject JSS_NewNextProtoResult(JNIEnv *env, int state, uint8_t *proto,
+    unsigned int proto_len)
+{
+    jclass resultClass;
+    jmethodID constructor;
+    jobject result = NULL;
+    jbyteArray protocol = NULL;
+
+    PR_ASSERT(env != NULL);
+
+    resultClass = (*env)->FindClass(env, NEXT_PROTO_CLASS_NAME);
+    if (resultClass == NULL) {
+        ASSERT_OUTOFMEM(env);
+        goto finish;
+    }
+
+    constructor = (*env)->GetMethodID(env, resultClass, PLAIN_CONSTRUCTOR,
+        NEXT_PROTO_CONSTRUCTOR_SIG);
+    if (constructor == NULL) {
+        ASSERT_OUTOFMEM(env);
+        goto finish;
+    }
+
+    if (proto != NULL) {
+        protocol = JSS_ToByteArray(env, proto, proto_len);
+    }
+
+    result = (*env)->NewObject(env, resultClass, constructor, state,
+                               protocol);
+
+finish:
+    return result;
+}
+
 JNIEXPORT jobject JNICALL
 Java_org_mozilla_jss_nss_SSL_ImportFD(JNIEnv *env, jclass clazz, jobject model,
     jobject fd)
@@ -840,6 +874,31 @@ Java_org_mozilla_jss_nss_SSL_SetNextProtoNeg(JNIEnv *env, jclass clazz,
     free(data);
 
     return ret;
+}
+
+JNIEXPORT jobject JNICALL
+Java_org_mozilla_jss_nss_SSL_GetNextProto(JNIEnv *env, jclass clazz,
+    jobject fd)
+{
+    PRFileDesc *real_fd = NULL;
+    SSLNextProtoState state;
+    uint8_t proto[255] = {0};
+    unsigned int proto_len;
+    SECStatus ret;
+
+    PR_ASSERT(env != NULL && fd != NULL);
+    PR_SetError(0, 0);
+
+    if (JSS_PR_getPRFileDesc(env, fd, &real_fd) != PR_SUCCESS) {
+        return NULL;
+    }
+
+    ret = SSL_GetNextProto(real_fd, &state, proto, &proto_len,
+                           sizeof(proto));
+    if (ret != SECSuccess) {
+        return JSS_NewNextProtoResult(env, state, NULL, 0);
+    }
+    return JSS_NewNextProtoResult(env, state, proto, proto_len);
 }
 
 JNIEXPORT jint JNICALL
--- a/org/mozilla/jss/ssl/javax/JSSEngine.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngine.java
@@ -187,6 +187,11 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
     private final static AtomicBoolean sessionCacheInitialized = new AtomicBoolean();
 
     /**
+     * Set of possible application protocols to negotiate.
+     */
+    protected String[] alpn_protocols;
+
+    /**
      * Constructor for a JSSEngine, providing no hints for an internal
      * session reuse strategy and no key.
      *
@@ -382,6 +387,12 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
         // it.
         if (parsed.getHostname() != null) {
             setHostname(parsed.getHostname());
+        }
+
+        // When we have a non-zero number of ALPNs, use them in the
+        // negotiation.
+        if (parsed.getApplicationProtocols() != null) {
+            setApplicationProtocols(parsed.getApplicationProtocols());
         }
     }
 
@@ -903,6 +914,58 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
      */
     public boolean getWantClientAuth() {
         return want_client_auth;
+    }
+
+    /**
+     * Set a specific list of protocols to negotiate next for ALPN support.
+     */
+    public void setApplicationProtocols(String[] protocols) {
+        alpn_protocols = protocols;
+    }
+
+    /**
+     * Get the most recently negotiated application protocol.
+     *
+     * Note that NSS only allows selection on the initial handshake so
+     * this is implemented via a call to getHandshakeApplicationProtocol().
+     */
+    public String getApplicationProtocol() {
+        return getHandshakeApplicationProtocol();
+    }
+
+    /**
+     * Get the application protocol negotiated during the initial handshake.
+     */
+    public String getHandshakeApplicationProtocol() {
+        if (session == null) {
+            return null;
+        }
+
+        return session.getNextProtocol();
+    }
+
+    /**
+     * Helper method for implementations: encodes ALPN protocols into wire
+     * format (8-bit length prefixed byte encoding).
+     */
+    public byte[] getALPNWireData() {
+        int length = 0;
+        for (String protocol : alpn_protocols) {
+            length += 1 + protocol.getBytes().length;
+        }
+
+        byte[] result = new byte[length];
+        int offset = 0;
+
+        for (String protocol : alpn_protocols) {
+            byte[] p_bytes = protocol.getBytes();
+            result[offset] = (byte) p_bytes.length;
+            offset += 1;
+            System.arraycopy(p_bytes, 0, result, offset, p_bytes.length);
+            offset += p_bytes.length;
+        }
+
+        return result;
     }
 
     /**
--- a/org/mozilla/jss/ssl/javax/JSSEngine.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngine.java
@@ -190,7 +190,7 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
     /**
      * Set of possible application protocols to negotiate.
      */
-    protected String[] alpn_protocols;
+    protected byte[][] alpn_protocols;
 
     /**
      * Constructor for a JSSEngine, providing no hints for an internal
@@ -392,8 +392,8 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
 
         // When we have a non-zero number of ALPNs, use them in the
         // negotiation.
-        if (parsed.getApplicationProtocols() != null) {
-            setApplicationProtocols(parsed.getApplicationProtocols());
+        if (parsed.getRawApplicationProtocols() != null) {
+            setApplicationProtocols(parsed.getRawApplicationProtocols());
         }
     }
 
@@ -921,6 +921,15 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
      * Set a specific list of protocols to negotiate next for ALPN support.
      */
     public void setApplicationProtocols(String[] protocols) {
+        JSSParameters parser = new JSSParameters();
+        parser.setApplicationProtocols(protocols);
+        alpn_protocols = parser.getRawApplicationProtocols();
+    }
+
+    /**
+     * Set a specific list of protocols to negotiate next for ALPN support.
+     */
+    public void setApplicationProtocols(byte[][] protocols) {
         alpn_protocols = protocols;
     }
 
@@ -951,19 +960,31 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
      */
     public byte[] getALPNWireData() {
         int length = 0;
-        for (String protocol : alpn_protocols) {
-            length += 1 + protocol.getBytes(StandardCharsets.UTF_8).length;
+        for (byte[] protocol : alpn_protocols) {
+            length += 1 + protocol.length;
         }
 
         byte[] result = new byte[length];
         int offset = 0;
 
-        for (String protocol : alpn_protocols) {
-            byte[] p_bytes = protocol.getBytes(StandardCharsets.UTF_8);
-            result[offset] = (byte) p_bytes.length;
+        // XXX: Handle custom encoding left over from NPN draft: NSS's
+        // SSL_SetNextProtoNego takes the first protocol and helpfully puts
+        // it at the end of the list for us... So when we're validating using
+        // the default ALPN callback handler, switch the first to the last to
+        // ensure we're passing it in the caller's specified order.
+        byte[] last = alpn_protocols[alpn_protocols.length - 1];
+        result[offset] = (byte) last.length;
+        offset += 1;
+        System.arraycopy(last, 0, result, offset, last.length);
+        offset += last.length;
+
+        // Now copy the rest of the protocols.
+        for (int index = 0; index < alpn_protocols.length - 1; index++) {
+            byte[] protocol = alpn_protocols[index];
+            result[offset] = (byte) protocol.length;
             offset += 1;
-            System.arraycopy(p_bytes, 0, result, offset, p_bytes.length);
-            offset += p_bytes.length;
+            System.arraycopy(protocol, 0, result, offset, protocol.length);
+            offset += protocol.length;
         }
 
         return result;
--- a/org/mozilla/jss/ssl/javax/JSSEngine.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngine.java
@@ -1,5 +1,6 @@
 package org.mozilla.jss.ssl.javax;
 
+import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.concurrent.atomic.AtomicBoolean;
 import javax.net.ssl.*;
@@ -951,14 +952,14 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
     public byte[] getALPNWireData() {
         int length = 0;
         for (String protocol : alpn_protocols) {
-            length += 1 + protocol.getBytes().length;
+            length += 1 + protocol.getBytes(StandardCharsets.UTF_8).length;
         }
 
         byte[] result = new byte[length];
         int offset = 0;
 
         for (String protocol : alpn_protocols) {
-            byte[] p_bytes = protocol.getBytes();
+            byte[] p_bytes = protocol.getBytes(StandardCharsets.UTF_8);
             result[offset] = (byte) p_bytes.length;
             offset += 1;
             System.arraycopy(p_bytes, 0, result, offset, p_bytes.length);
--- a/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -1108,22 +1108,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
             handshake_state = SSLEngineResult.HandshakeStatus.FINISHED;
             unknown_state_count = 0;
 
-            // Only update peer certificate chain when we've finished
-            // handshaking.
-            try {
-                PK11Cert[] peer_chain = SSL.PeerCertificateChain(ssl_fd);
-                session.setPeerCertificates(peer_chain);
-            } catch (Exception e) {
-                String msg = "Unable to get peer's certificate chain: ";
-                msg += e.getMessage();
-
-                seen_exception = true;
-                ssl_exception = new SSLException(msg, e);
-            }
-
-            // Also update our session information here.
-            session.refreshData();
-
+            updateSession();
             return;
         }
 
--- a/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -349,10 +349,13 @@ public class JSSEngineReferenceImpl extends JSSEngine {
 
         if (alpn_protocols != null) {
             byte[] wire_data = getALPNWireData();
+            if (wire_data == null) {
+                throw new RuntimeException("JSSEngine.init(): ALPN wire data is NULL but alpn_protocols is non-NULL.");
+            }
 
             ret = SSL.SetNextProtoNeg(ssl_fd, wire_data);
-            if (ret == SSL.SECFailure) {
-                throw new RuntimeException("JSSEngine.init(): Unable to set ALPN protocol list.");
+            if (ret != SSL.SECSuccess) {
+                throw new RuntimeException("JSSEngine.init(): Unable to set ALPN protocol list: " + errorText(PR.GetError()) + " " + ret);
             }
         }
     }
--- a/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -970,6 +970,8 @@ public class JSSEngineReferenceImpl extends JSSEngine {
                     ret.state != SSLNextProtoState.SSL_NEXT_PROTO_NO_OVERLAP)
                 {
                     session.setNextProtocol(ret.getProtocol());
+                } else {
+                    session.setNextProtocol("");
                 }
             }
         } catch (Exception e) {
--- a/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -347,8 +347,14 @@ public class JSSEngineReferenceImpl extends JSSEngine {
             throw new SSLException("Unable to enable SSL Handshake Callback on this SSLFDProxy instance.");
         }
 
-        // Pass this ssl_fd to the session object so that we can use
-        // SSL methods to invalidate the session.
+        if (alpn_protocols != null) {
+            byte[] wire_data = getALPNWireData();
+
+            ret = SSL.SetNextProtoNeg(ssl_fd, wire_data);
+            if (ret == SSL.SECFailure) {
+                throw new RuntimeException("JSSEngine.init(): Unable to set ALPN protocol list.");
+            }
+        }
     }
 
     private void initClient() throws SSLException {
@@ -935,6 +941,37 @@ public class JSSEngineReferenceImpl extends JSSEngine {
         }
 
         return data_index;
+    }
+
+    private void updateSession() {
+        if (ssl_fd == null) {
+            return;
+        }
+
+        try {
+            PK11Cert[] peer_chain = SSL.PeerCertificateChain(ssl_fd);
+            session.setPeerCertificates(peer_chain);
+
+            SSLChannelInfo info = SSL.GetChannelInfo(ssl_fd);
+            if (info == null) {
+                return;
+            }
+
+            session.setId(info.getSessionID());
+            session.refreshData();
+        
+            NextProtoResult ret = SSL.GetNextProto(ssl_fd);
+            if (ret != null) {
+                // Only advertise the resulting protocol if we have one.
+                if (ret.state != SSLNextProtoState.SSL_NEXT_PROTO_NO_SUPPORT &&
+                    ret.state != SSLNextProtoState.SSL_NEXT_PROTO_NO_OVERLAP)
+                {
+                    session.setNextProtocol(ret.getProtocol());
+                }
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(e.getMessage(), e);
+        }
     }
 
     private SSLException checkSSLAlerts() {
--- a/org/mozilla/jss/ssl/javax/JSSParameters.java
+++ b/org/mozilla/jss/ssl/javax/JSSParameters.java
@@ -3,6 +3,7 @@ package org.mozilla.jss.ssl.javax;
 import javax.net.ssl.*;
 import java.util.*;
 
+import org.mozilla.jss.util.JDKCompat;
 import org.mozilla.jss.ssl.*;
 
 /**
@@ -56,7 +57,7 @@ public class JSSParameters extends SSLParameters {
             setNeedClientAuth(downcast.getNeedClientAuth());
         }
 
-        String[] alpn = downcast.getApplicationProtocols(downcast);
+        String[] alpn = JDKCompat.SSLParametersHelper.getApplicationProtocols(downcast);
         if (alpn != null) {
             setApplicationProtocols(alpn);
         }
--- a/org/mozilla/jss/ssl/javax/JSSParameters.java
+++ b/org/mozilla/jss/ssl/javax/JSSParameters.java
@@ -28,7 +28,7 @@ public class JSSParameters extends SSLParameters {
     private SSLVersionRange range;
     private String alias;
     private String hostname;
-    private String[] appProtocols;
+    private byte[][] appProtocols;
 
     public JSSParameters() {
         // Choose our default set of SSLParameters here; default to null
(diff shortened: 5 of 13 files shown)


org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -576,6 +585,11 @@ private void updateSession() {
 
             session.setId(info.getSessionID());
             session.refreshData();
+        
+            NextProtoResult ret = SSL.GetNextProto(ssl_fd);
+            if (ret != null) {
+                session.setNextProtocol(ret.getProtocol());
```
cipherboy, line 591: Needs to only be done when `SSL_NEXT_PROTO_NEGOTIATED` or `SSL_NEXT_PROTO_SELECTED`; other values should be `null`.

cipherboy:
@edewata Note that ALPN support also gives us the ability to do TLS-ALPN-01 challenges with ACME if we wanted to.

org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -54,6 +56,11 @@ public JSSParameters(SSLParameters downcast) {
         if (downcast.getNeedClientAuth()) {
             setNeedClientAuth(downcast.getNeedClientAuth());
         }
+
+        String[] alpn = JDKCompat.SSLParametersHelper.getApplicationProtocols(downcast);
```
edewata, line 60: Is this necessary? The `SSLParameters.getApplicationProtocols()` doesn't seem to change in different Java versions:
* https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLParameters.html#getApplicationProtocols--
* https://docs.oracle.com/javase/9/docs/api/javax/net/ssl/SSLParameters.html#getApplicationProtocols--
* https://docs.oracle.com/en/java/javase/11/docs/api/java.base/javax/net/ssl/SSLParameters.html#getApplicationProtocols()

org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -297,6 +302,12 @@ public void setSSLParameters(SSLParameters params) {
         if (parsed.getHostname() != null) {
             setHostname(parsed.getHostname());
         }
+
+        // When we have a non-zero number of ALPNs, use them in the
+        // negotiation.
+        if (parsed.getApplicationProtocols() != null) {
+            setApplicationProtocols(parsed.getApplicationProtocols());
+        }
```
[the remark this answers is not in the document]
cipherboy: If we take a JDK8 `SSLParameters` (`params` here) and downcast to `JSSParameters` (`parsed`), we have no `getApplicationProtocols()` function in `params` so `parsed` will return `null`. However, because `SSLEngine` has a `setApplicationProtocols` method the following code would work:

```java
// SSLEngine inst;
inst.setApplicationProtocols(...);
inst.setSSLParameters(params);
```

If we overwrite it with `null` here, we could be overwriting intentional changes. We could try and detect this case but I'd prefer to leave it with the non-overwriting case to match all the other logic in the function anyways.

org/mozilla/jss/util/JDKCompat.java
```diff
@@ -0,0 +1,21 @@
...
+public class JDKCompat {
+    public static class SSLParametersHelper {
+        public static String[] getApplicationProtocols(SSLParameters inst) {
+            try {
+                Method getter = inst.getClass().getMethod("getApplicationProtocols");
+                Object result = getter.invoke(inst);
+                if (result instanceof String[]) {
+                    return (String[]) result;
+                }
+            } catch (Throwable t) {}
```
edewata, line 16: If we want to make ALPN support optional, we should only catch `NoSuchMethodException` and explicitly ignore or a generate warning. All other exceptions should still be propagated to the callers.
```
} catch (NoSuchMethodException e) {
    // ALPN not supported
} catch (Exception e) {
    new RuntimeException(e);
}
```
Also, it's not necessary to check the type of `result` since it's always `String[]` in all Java versions (if the method is available).

org/mozilla/jss/ssl/javax/JSSEngine.java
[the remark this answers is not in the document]
cipherboy: In particular, if you take this source:

https://gist.github.com/cipherboy/f1144c36bb635257bd2b6bf2adb72513

If you compile and run with the default JDK (in my case 8), it'll fail to compile. That's fine since _this_ JDK8 doesn't have the method.

If you compile with JDK 11 and run with this older JDK8, you'll get a nice exception telling you that you can't do that because the method doesn't exist. 

If you compile with JDK11 and run with JDK11, it'll work fine. Presumably, this'd be the same behavior that we'd get when JDK8 updates. 

I think that's about as close to non-surprising behavior as you can get :-)

org/mozilla/jss/util/JDKCompat.java
cipherboy: I think we need to catch more than `NoSuchMethodException` -- we also need to handle what happens when reflection fails (e.g., `SecurityManager` fails to allow the operation). In this case, we should log a message saying why we can't proceed and tell them to use `JSSParameters` if it is really serious.

org/mozilla/jss/ssl/javax/JSSEngine.java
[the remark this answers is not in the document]
cipherboy: Also to summarize some discussions on IRC:

BouncyCastle provides this in its BCSSLEngine/BCSSLParameters/... classes as well, for all JDK versions. When they compiled with JDK9, they'll add the methods to read the SSLParameters's value and put it into their BCSSLParameters copy.

org/mozilla/jss/util/JDKCompat.java
edewata: I think since we're determining ALPN support based on the availability of the method, we should only check `NoSuchMethodException` and ignore that exception only. If throws a different exception, that would be a different issue that needs to be reported to the caller. If we ignore all exceptions, that means ALPN support is really arbitrary (i.e. you can use it if it works) and there's no way to guarantee it if you need it.

org/mozilla/jss/ssl/javax/JSSParameters.java
cipherboy: See above; yes, because support in JDK8 is really, really new. :)

org/mozilla/jss/util/JDKCompat.java
cipherboy: I'm worried about what happens [when a `SecurityException` is thrown](https://docs.oracle.com/javase/8/docs/api/java/lang/Class.html#getMethod-java.lang.String-java.lang.Class...-).

I don't think we can guarantee that the method _does_ exist when a `SecurityException` is thrown. So we don't have enough information as to whether or not we should treat this as fatal or not. I'd be fine limiting it to `NoSuchMethodException` and `SecurityException`.

org/mozilla/jss/ssl/javax/JSSEngine.java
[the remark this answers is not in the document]
edewata: This probably why in general it's easier to bump the dependency (to Java 8u) such that certain feature is consistently available, especially when everybody is expected to use the latest Java 8 or later anyway.

If `SSLParameters.getApplicationProtocols()` is guaranteed to be available, the code could be as simple as this:
```
setApplicationProtocols(parsed.getApplicationProtocols());
```
If `SSLParameters.getApplicationProtocols()` is not consistently available, we probably should do something like this:
```
try {
    setApplicationProtocols(parsed.getApplicationProtocols());
} catch (NoSuchMethodException e) {
    // SSLParameters.getApplicationProtocols() is not available
}
```
Here we probably can implement a `JSSParameters.getApplicationProtocols()` that uses reflection to call `SSLParameters.getApplicationProtocols()`. If the method is not available, we simply let the caller handle the `NoSuchMethodException`.

Here the behavior would be more accurate: call `JSSEngine.setApplicationProtocols()` only if `SSLParameters.getApplicationProtocols()` is available.

org/mozilla/jss/util/JDKCompat.java
edewata: We only have this problem only because we need to store a value in `appProtocols`, and a `null` could be interpreted as real `null` or unset. See my other comment about catching the `NoSuchMethodException` in `JSSEngine` so we will not have this issue.

cipherboy:
I've rebased this to drop the JDK compatibility interface. Hopefully a high enough version of the JDK has shipped everywhere we need it to for this feature.

cipherboy:
@edewata Looks like Fedora 32 still doesn't have ALPN support in JDK8. Thoughts?

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -347,8 +347,17 @@ private void createBufferFD() throws SSLException {
             throw new SSLException("Unable to enable SSL Handshake Callback on this SSLFDProxy instance.");
         }
 
-        // Pass this ssl_fd to the session object so that we can use
-        // SSL methods to invalidate the session.
+        if (alpn_protocols != null) {
+            byte[] wire_data = getALPNWireData();
+            if (wire_data == null) {
+                throw new RuntimeException("JSSEngine.init(): ALPN wire data is NULL but alpn_protocols is non-NULL.");
+            }
```
edewata, line 354: I think `getALPNWireData()` will never return a `null`, but it's fine.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -937,6 +946,37 @@ private int putData(byte[] data, ByteBuffer[] buffers, int offset, int length) {
         return data_index;
     }
 
+    private void updateSession() {
```
edewata, line 949: Is this used anywhere?

org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -905,6 +916,58 @@ public boolean getWantClientAuth() {
...
+    }
+
+    /**
+     * Helper method for implementations: encodes ALPN protocols into wire
+     * format (8-bit length prefixed byte encoding).
+     */
+    public byte[] getALPNWireData() {
+        int length = 0;
+        for (String protocol : alpn_protocols) {
+            length += 1 + protocol.getBytes().length;
```
edewata, line 954: Do we need to use UTF-8 specifically? Otherwise `getBytes()` will use the system's default encoding. Or if the protocol names are ASCII anyway then it doesn't matter.

org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -190,4 +196,27 @@ public String getHostname() {
...
+
+    public void setApplicationProtocols(String[] protocols) throws IllegalArgumentException {
+        if (protocols == null) {
+            appProtocols = null;
+            return;
+        }
+
+        int index = 0;
+        for (String protocol : protocols) {
+            if (protocol.length() > 255 || protocol.getBytes().length > 255) {
```
edewata, line 208: If the protocol names are ASCII we probably just need to check the `protocol.length()`.

edewata (approved):
Not familiar with ALPN, but it looks fine to me, just some minor comments. Feel free to merge.

org/mozilla/jss/ssl/javax/JSSEngine.java
frasertweedale: Officially an ALPN protocol ID is a sequence of bytes and they are registered with IANA as such; see https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids.  Therefore I recommend the JSS interface deal with protocol IDs as `byte[]` the whole way through, and defer conversion to/from other types (such as `String`) to the library user.  This is what I did in my other ALPN PR #640.

cipherboy: This isn't actually possible. This PR is against the JCA-standard `javax.net.ssl` implementation that I've added to JSS. It resides under the `org.mozilla.jss.ssl.javax` package namespace to avoid confusing it with the custom, legacy "SSLSocket" implementation under `org.mozilla.jss.ssl`. 

In particular, looking at the interface we're given:

 - We get the output as a `String`: [`SSLSocket.getApplicationProtocol()`](https://docs.oracle.com/javase/9/docs/api/javax/net/ssl/SSLSocket.html#getApplicationProtocol--), [`SSLEngine.getApplicationProtocol()`](https://docs.oracle.com/javase/9/docs/api/javax/net/ssl/SSLEngine.html#getApplicationProtocol--)
 - We get the "input"/protocol choices as a `String`: [`SSLEngine.setHandshakeApplicationProtocolSelector(...)`](https://docs.oracle.com/javase/9/docs/api/javax/net/ssl/SSLEngine.html#setHandshakeApplicationProtocolSelector-java.util.function.BiFunction-)

So we either choose our own encoding and force it throughout, or let the user control it via system default.

It looks like the JDK11 always forces UTF-8:
https://github.com/openjdk/jdk11u-dev/blob/master/src/java.base/share/classes/sun/security/ssl/AlpnExtension.java#L100

So perhaps this will be good enough for us.

org/mozilla/jss/tests/TestSSLEngine.java
```diff
@@ -978,6 +978,38 @@ public static void testPostHandshakeAuth(SSLContext ctx, String client_alias, St
...
+        server_eng.setUseClientMode(false);
+
+        if (server_eng instanceof JSSEngineReferenceImpl) {
+            ((JSSEngineReferenceImpl) server_eng).setName("JSS Server for ALPN");
+            ((JSSEngineReferenceImpl) server_eng).enableSafeDebugLogging(7377);
+        }
+
+        try {
+            testBasicHandshake(client_eng, server_eng, false);
+            assert(server_eng.getApplicationProtocol().equals("h2"));
```
cipherboy, line 1005: I uh... Don't actually understand how this test case passed. :-) As I read the NSS code base, it requires a callback handler to work but we never provided any!

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
cipherboy: Ugh, did my rebase lose this again? :(

org/mozilla/jss/tests/TestSSLEngine.java
cipherboy: This now fails correctly.

edewata (approved):
I don't have any objection, as long as it passes CI feel free to merge it.
Is there a real application that can use this now, or do we need to
implement tls-alpn-01 ACME challenge in order to truly test it?

cipherboy:
We have two options. I'm inclined to leave it open a little longer and implement server support. Or we could point a small client socket at something like `google.com` which supports HTTP/2 and see if it negotiates it correctly. :)

frasertweedale:
@cipherboy it was already open nearly 6 months.  Leave it a bit longer.  I want to play with it and will adapt my test program from #640.  But it is not top priority, expect I will get to it within the next few weeks.

org/mozilla/jss/ssl/javax/JSSEngine.java
frasertweedale: It will be good enough for us.  But it is seriously flawed w.r.t. how the ALPN extension and protocol ID is actually specified.  Java's fault, we just have to put up with it.

org/mozilla/jss/util/java_ids.h
```diff
@@ -444,6 +444,12 @@ PR_BEGIN_EXTERN_C
 #define SSL_PRELIMINARY_CHANNEL_INFO_CLASS_NAME "org/mozilla/jss/nss/SSLPreliminaryChannelInfo"
 #define SSL_PRELIMINARY_CHANNEL_INFO_CONSTRUCTOR_SIG "(JIIZJZIZZII)V"
 
+/*
+ * NextProtoResult classes
+ */
+#define NEXT_PROTO_CLASS_NAME "org/mozilla/jss/nss/NextProtoResult"
+#define NEXT_PROTO_CONSTRUCTOR_SIG "(I[B)V"
```
frasertweedale, line 451: We need to add a reference to this in a comment at the corresponding constructor in the Java class, so that if someone changes the constructor they know to also change this, and related C code.

frasertweedale (changes requested):
There are some nits and one significant issue identified in comments.

Overall I am at a bit of a loss about the `JSSEngine` class in general (which implements the abstract `SSLEngine` class).  It is not at all clear to me how to use that class to establish a TLS-secured network connection.  Surely we also want to enhance `JSSSocket` with ALPN capability?

org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -190,4 +198,27 @@ public String getHostname() {
     public void setHostname(String server_hostname) {
         hostname = server_hostname;
     }
+
+    public void setApplicationProtocols(String[] protocols) throws IllegalArgumentException {
```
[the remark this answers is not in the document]
cipherboy: If you want to shake your fist at Java some more... :-)

[OpenJDK 9 b96](https://bugs.openjdk.java.net/browse/JDK-8144093) introduced ALPN support. We can only use JDK9+ on Fedora 33+. So we need something to support JDK8.

[OpenJDK 8 b252](https://bugs.openjdk.java.net/browse/JDK-8230977) should include ALPN support. However, this was only released late last year. We either introduce a dependency on a really recent version of the JDK (which we've historically avoided), or we introduce this into `JSSParameters`. Since `JSSParameters` exists to provide a mechanism for JSS-specific parameters to work (such as Hostname... </not-salty>) -- introducing another parameter for compatibility seems like a good idea.

My 2c but I'd rather not require a specific version of the JDK other than "at least JDK 8".

org/mozilla/jss/tests/TestSSLEngine.java
```diff
@@ -1001,6 +1034,23 @@ public static void testNativeClientServer(String[] args) throws Exception {
         testAllHandshakes(ctx, client_alias, server_alias, true);
         testPostHandshakeAuth(ctx, client_alias, server_alias);
         testJSSEToJSSHandshakes(ctx, server_alias);
+
+        // TODO: Enable once server-side ALPN support is merged.
+        // testALPNHandshake(ctx, server_alias);
+    }
+
+    public static void testALPNEncoding() throws Exception {
```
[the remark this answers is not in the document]
cipherboy: I disagree, let me reply to your other comment though. This _only_ tests the byte encoding to make sure it is what NSS expects.

org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -905,6 +917,58 @@ public boolean getWantClientAuth() {
...
+        byte[] result = new byte[length];
+        int offset = 0;
+
+        for (String protocol : alpn_protocols) {
+            byte[] p_bytes = protocol.getBytes(StandardCharsets.UTF_8);
+            result[offset] = (byte) p_bytes.length;
+            offset += 1;
+            System.arraycopy(p_bytes, 0, result, offset, p_bytes.length);
+            offset += p_bytes.length;
+        }
```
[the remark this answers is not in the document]
cipherboy: I disagree per spec.

Per [RFC 7301 section 3.2](https://tools.ietf.org/html/rfc7301#section-3.2), ordering only matters on the server-side. 

>    It is expected that a server will have a list of protocols that it
>    supports, in preference order, and will only select a protocol if the
>    client supports it. 

IOW, client order doesn't matter, only server order.

This function (`SSL_SetNextProtoNego`, which `getALPNWireData` builds the arguments for) is _only_ for client-side code. `getALPNWireData` will not be used for server-side code. What order we encode stuff in, while relevant to the underlying ALPN wire encoding, doesn't really matter to protocol selection.

org/mozilla/jss/nss/SSL.c
```diff
@@ -140,6 +140,40 @@ jobject JSS_NewSSLPreliminaryChannelInfo(JNIEnv *env, jlong valuesSet,
...
+    }
+
+    constructor = (*env)->GetMethodID(env, resultClass, PLAIN_CONSTRUCTOR,
+        NEXT_PROTO_CONSTRUCTOR_SIG);
+    if (constructor == NULL) {
+        ASSERT_OUTOFMEM(env);
+        goto finish;
+    }
+
+    if (proto != NULL) {
```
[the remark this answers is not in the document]
cipherboy: `SSL_GetNextProto` returns a complex object: `NextProtoResult` which includes a `SSLNextProtoState` member and a byte array. The value of the byte array should NOT be used when the `SSLNextProtoState` isn't either `SSL_NEXT_PROTO_NEGOTIATED` or `SSL_NEXT_PROTO_SELECTED` (or arguably, `SSL_NEXT_PROTO_EARLY_VALUE`).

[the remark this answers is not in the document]
cipherboy: (IOW, the absence of a negotiated protocol ID is already overly-defined) :-)

org/mozilla/jss/nss/SSL.c
```diff
@@ -816,6 +850,57 @@ Java_org_mozilla_jss_nss_SSL_KeyUpdate(JNIEnv *env, jclass clazz,
...
+    return ret;
+}
+
+JNIEXPORT jobject JNICALL
+Java_org_mozilla_jss_nss_SSL_GetNextProto(JNIEnv *env, jclass clazz,
+    jobject fd)
+{
+    PRFileDesc *real_fd = NULL;
+    SSLNextProtoState state;
+    uint8_t proto[512] = {0};
```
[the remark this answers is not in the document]
cipherboy: Sure, this seems fine. Worst case, NSS will err out if they somehow end up with a longer ALPN data than we mark here.

org/mozilla/jss/nss/NextProtoResult.java
```diff
@@ -0,0 +1,39 @@
...
+        state = SSLNextProtoState.valueOf(state_value);
+        this.protocol = protocol;
+    }
+
+    public String getProtocol() {
+        if (protocol == null) {
+            return null;
+        }
+
+        return new String(protocol);
```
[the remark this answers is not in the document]
cipherboy: Ah missed this case.

org/mozilla/jss/nss/SSL.c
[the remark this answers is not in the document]
cipherboy: I've made `getProtocol()` return `null` string when a 0-length array is passed in.

org/mozilla/jss/ssl/javax/JSSEngine.java
cipherboy: I mean, I don't disagree:

```java
        byte[] a = new byte[]{ (byte) 0xFA, (byte) 0xFA };
        String b = new String(a, StandardCharsets.UTF_8);
        byte[] c = b.getBytes(StandardCharsets.UTF_8);
        System.out.println(a.length + " " + a[0] + " " +a[1]);
        System.out.println(b); 
        System.out.println(c.length + " " + c[0] + " " + c[1]);
```

But also.. If you and the server both negotiate GREASE'd values... What does that even mean? What protocol are you actually negotiating there? It really isn't well defined. So they just hope the rest of the (valid) IANA protocol values are otherwise valid.

cipherboy: (I say negotiate because NSS and the protocol doesn't give the client a way to see what the server has for ALPN data priorities, and for that matter, the String comparison all happens on the server-side anyways, so we should probably coerce byte -> String and use `String.equals(...)`, rather than trying to go from String -> Byte and compare byte arrays... That ensure we probably have the same encoded representation rather than dealing with differences there...)

frasertweedale: Well, there's nothing we can do about it but to descend slightly further into this rabbit hole:

1. You don't negotiate GREASE (servers MUST ignore GREASE values).  But if you want to *send* GREASE in your ClientHello, 15 of the 16 defined values are impossible to send.
2. And if any other ALPN values get defined that are not the UTF-8 encoding of some string, then you can't send those either.

On your second comment, no, the string comparison is essential on the client side too: the client will usually need to check what protocol was negotiated.

As for how to compare, given the interface imposed on us I have no objection to `String.equals()`.

org/mozilla/jss/ssl/javax/JSSParameters.java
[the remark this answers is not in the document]
frasertweedale: @cipherboy thanks for clarifying.

FWIW f32 has OpenJDK 8 b265 and I confirmed it has the ALPN bits.

I don't mind if we reproduce the behaviour but there should at least be a comment that when we are depending on OpenJDK >=9 or >= 8 b252 then we can remove these implementations.

org/mozilla/jss/ssl/javax/JSSEngine.java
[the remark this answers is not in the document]
frasertweedale: `SSL_SetNextProtoNego` can be used for setting server-side preferences as well.  Have a look at the NSS implementation in `lib/ssl/sslsock.c`.  It calls `SSL_SetNextProtoCallback` internally to set a callback function that will read the given protocol list.

Furthermore, the RFC does seem to have some conflicting statements about whose preferences are respected, e.g. in https://tools.ietf.org/html/rfc7301#section-3.1:
```
   "ProtocolNameList" contains the list of protocols advertised by the
   client, in descending order of preference.
```

I agree with your interpretation of section 3.2.  But the RFC is clear that the client advertisement is also to be in preference order.  The behaviour suggested in 3.2 is a SHOULD and an implementation might decide to respect client preferences instead.

frasertweedale:
> > Overall I am at a bit of a loss about the JSSEngine class in general (which implements the abstract SSLEngine class). It is not at all clear to me how to use that class to establish a TLS-secured network connection. Surely we also want to enhance JSSSocket with ALPN capability?
> 
> This is done now. Technically this wasn't necessary if you had provided a `JSSParameters` to the `JSSSocket`. But easy enough to add the other methods. :-)

Thanks @cipherboy.  It is important to have `JSSSocket.getApplicationProtocol()` so the client or server can find out what (if any) protocol was negotiated.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -937,6 +946,37 @@ private int putData(byte[] data, ByteBuffer[] buffers, int offset, int length) {
...
+            }
+
+            session.setId(info.getSessionID());
+            session.refreshData();
+        
+            NextProtoResult ret = SSL.GetNextProto(ssl_fd);
+            if (ret != null) {
+                // Only advertise the resulting protocol if we have one.
+                if (ret.state != SSLNextProtoState.SSL_NEXT_PROTO_NO_SUPPORT &&
+                    ret.state != SSLNextProtoState.SSL_NEXT_PROTO_NO_OVERLAP)
```
frasertweedale, line 970: We should be more defensive here; also check that `ret.getProtocol()` is both non-null and non-empty.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -347,8 +347,17 @@ private void createBufferFD() throws SSLException {
...
-        // SSL methods to invalidate the session.
+        if (alpn_protocols != null) {
+            byte[] wire_data = getALPNWireData();
+            if (wire_data == null) {
+                throw new RuntimeException("JSSEngine.init(): ALPN wire data is NULL but alpn_protocols is non-NULL.");
+            }
+
+            ret = SSL.SetNextProtoNeg(ssl_fd, wire_data);
+            if (ret != SSL.SECSuccess) {
+                throw new RuntimeException("JSSEngine.init(): Unable to set ALPN protocol list: " + errorText(PR.GetError()) + " " + ret);
```
frasertweedale, line 358: Per other comments, this appears to be calling `SSL_SetNextProtoNego` for both client and server.  And that is the correct behaviour.  But we need to handle the address the NSS protocol ordering quirks.

org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -190,4 +198,27 @@ public String getHostname() {
...
+        if (protocols == null) {
+            appProtocols = null;
+            return;
+        }
+
+        int index = 0;
+        for (String protocol : protocols) {
+            if (protocol != "" && protocol.getBytes(StandardCharsets.UTF_8).length > 255) {
```
frasertweedale, line 210: should ensure non-empty and length < 255, e.g.
```java
for (String protocol : protocols) {
    if (null == protocol)
        throw new IllegalArgumentException("null protocol");
    byte[] bs = protocol.getBytes(StandardCharsets.UTF_8);
    if (bs.length < 1 || bs.length > 255)
        throw new IllegalArgumentException("protocol length must be 1..255");
}
```

This is actually stricter than what's in the openjdk implementation (it has null/empty check but not < 255 check).
But that is important to check.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -347,8 +347,17 @@ private void createBufferFD() throws SSLException {
             throw new SSLException("Unable to enable SSL Handshake Callback on this SSLFDProxy instance.");
         }
 
-        // Pass this ssl_fd to the session object so that we can use
-        // SSL methods to invalidate the session.
+        if (alpn_protocols != null) {
```
frasertweedale, line 350: should also check `alpn_protocols.length > 0`.

org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -905,6 +917,58 @@ public boolean getWantClientAuth() {
...
+
+        return result;
```
frasertweedale, line 969: We should also check that `result.length < 65536`.  Not sure what exception we should raise if that were not met, but we should check it.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
frasertweedale: @cipherboy I just noticed that the return contract of `getApplicationProtocol` (is SSLEngine and SSLSocket and probably other places) is:

> null if it has not yet been determined if application protocols might be used for this connection, an empty String if application protocols values will not be used, or a non-empty application protocol String if a value was successfully negotiated.

So it seems that we should always `setNextProtocol`

- empty string for `NO_SUPPORT` and `NO_OVERLAP`
- `ret.getProtocol()` for other cases.

cipherboy:
> > > Overall I am at a bit of a loss about the JSSEngine class in general (which implements the abstract SSLEngine class). It is not at all clear to me how to use that class to establish a TLS-secured network connection. Surely we also want to enhance JSSSocket with ALPN capability?
> > 
> > 
> > This is done now. Technically this wasn't necessary if you had provided a `JSSParameters` to the `JSSSocket`. But easy enough to add the other methods. :-)
> 
> Thanks @cipherboy. It is important to have `JSSSocket.getApplicationProtocol()` so the client or server can find out what (if any) protocol was negotiated.

`JSSSocket.getSession().getNextProtocol()` works too :-)

And note that this is again, only client-side checking. The server does not use `SSL_SetNextProtoNego()` and instead _only_ uses `SSL_SetNextProtoCallback`, which internally maintains its own list and/or is passed one on `arg` depending on how the user writes their callback handler. In a language more pythonic, the callback function signature is:

```python
def SSLNextProtoCallback(caller_arg: Any, fd: NSPR.PRFileDesc<NSS.SSL>, peer_protos: List[str], max_ret_length: int) -> str:
    pass
```

and there are no other helper functions which return the server-side configured list of protocols. So there is no way for the server to take that into account, if set. Additionally, it is in the wrong format: we have to encode the list of protocols ourselves, so we'd have to re-parse that into a useful value in order to make sense of them in the callback handler. However, NSS already does this for us with the client's list of protocols. 

Indeed, you can see that in the gtests here:

https://github.com/nss-dev/nss/blob/master/gtests/ssl_gtest/tls_connect.cc#L627

It does its own parsing of the server-provided "arg" value (which in turn is its own list of values) and uses that to determine which of the protos to return.

org/mozilla/jss/ssl/javax/JSSEngine.java
cipherboy: I suppose... This is why we have JSSEngine and JSSParameters after all... We could add a custom getter/setter which returns the raw byte values... Thoughts?

cipherboy: @frasertweedale -- I've brought this up for discussion on the OpenJDK `security-dev` mailing list and CC'd you. That was the approach suggested by `#java`.

frasertweedale:
> And note that this is again, only client-side checking. The server does not use `SSL_SetNextProtoNego()` and instead _only_ uses `SSL_SetNextProtoCallback`, which internally maintains its own list and/or is passed one on `arg` depending on how the user writes their callback handler. In a language more pythonic, the callback function signature is:
> 
> ```python
> def SSLNextProtoCallback(caller_arg: Any, fd: NSPR.PRFileDesc<NSS.SSL>, peer_protos: List[str], max_ret_length: int) -> str:
>     pass
> ```
> 
> and there are no other helper functions which return the server-side configured list of protocols. So there is no way for the server to take that into account, if set. Additionally, it is in the wrong format: we have to encode the list of protocols ourselves, so we'd have to re-parse that into a useful value in order to make sense of them in the callback handler. However, NSS already does this for us with the client's list of protocols.
> 
> Indeed, you can see that in the gtests here:
> 
> https://github.com/nss-dev/nss/blob/master/gtests/ssl_gtest/tls_connect.cc#L627
> 
> It does its own parsing of the server-provided "arg" value (which in turn is its own list of values) and uses that to determine which of the protos to return.

I admit I haven't tested it, but let me refer to the NSS code that leads me to the opposite conclusion:
https://github.com/nss-dev/nss/blob/master/lib/ssl/sslsock.c#L2224-L2254

As you can see, this sets the protocol list data `ss->opt.nextProtoNego.data` which a client will transmit to server.  It also calls `SSL_SetNextProtoCallback(fd, ssl_NextProtoNegoCallback, NULL)`.
The callback function `ssl_NextProtoNegoCallback` is a server-side callback that refers to the protocol list data just set (`ss->opt.nextProtoNego.data`).

In this way, `SSL_SetNextProtoNego` satisfies *both* the client side requirements, and the common case for the server side too.

org/mozilla/jss/ssl/javax/JSSEngine.java
frasertweedale: Yes, we could implement our own methods that deal with `byte[]`, if the requirement arises.

I don't think the need is there right now (I do not need to send grease or handle a non-UTF-8 protocol ID).

I'm just annoyed at such a mistake in a foundational API.  So easliy avoided, if they had just read the spec and implemented what is says.

cipherboy:
Aha, that's interesting. Then there must be some other bug in this code, I'll have to take a look more closely.

I was looking here:

https://github.com/nss-dev/nss/blob/master/lib/ssl/ssl3exthandle.c#L325-L334
https://github.com/nss-dev/nss/blob/master/lib/ssl/ssl3exthandle.c#L388-L396

I missed that `SSL_SetNextProtoNego` had its own custom handler. I gotcha now.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
cipherboy: What `SSLParameters` class? We explicitly downcast into `JSSParameters` [within `JSSEngine`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/ssl/javax/JSSEngine.java#L345-L355) so we can use a simple `null` check.

org/mozilla/jss/ssl/javax/JSSParameters.java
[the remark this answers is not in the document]
cipherboy: We need to keep these anyways, as we need to convert to byte encoding. There are no internal helpers (on SSLEngine) the would let us access the advertised list anyways.

What could theoretically go away is the SSLParameters stuff, but again, that's useful for other reasons (converting to byte-encoding).

So that just leaves the JDKCompat shim as the only thing that can and should actually be removed in the future. IMO.

org/mozilla/jss/tests/TestSSLEngine.java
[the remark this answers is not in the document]
cipherboy: Updated.

org/mozilla/jss/ssl/javax/JSSParameters.java
```diff
@@ -54,6 +57,11 @@ public JSSParameters(SSLParameters downcast) {
         if (downcast.getNeedClientAuth()) {
             setNeedClientAuth(downcast.getNeedClientAuth());
         }
+
+        String[] alpn = JDKCompat.SSLParametersHelper.getApplicationProtocols(downcast);
+        if (alpn != null) {
+            setApplicationProtocols(alpn);
+        }
```
frasertweedale, line 64: This is how can end up with *empty*, non-null list of protocols.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
frasertweedale: In the process of doing that downcast from `SSLParameters` to `JSSParameters`, we take the ALPN value as-is (see 
https://github.com/dogtagpki/jss/pull/526/files#r503618760).  So if we construct a `JSSEngine` with an "native" `SSLParameters`, and have not overridden the ALPN data, then we get an empty list. This is why the check is necessary.
This is not a theoretical issue - I hit it with my test program.

But even so, it is just a good defensive coding practice.  If the type admits an invalid datum (it does), we should code up checks against it - or change the type (not practical here because `String[]` is imposed).

frasertweedale: @cipherboy I've tested this PR and it's almost there.  We really do need this length check though.
Feel free to just apply this patch or modify the relevant commit: whichever you prefer:

```patch
commit 5e09241e923b426733c444ccb1ed9295726c3080
Author: Fraser Tweedale <ftweedal@redhat.com>
Date:   Tue Oct 13 12:45:46 2020 +1100

    ALPN: guard against empty array of protocol IDs
    
    javax.net.ssl.SSLParameters sets default ALPN protocol IDs to empty
    array.  When downcasting to JSSParameters, we take the value as-is,
    so we could also end up with an empty array.  Therefore when
    deciding whether to create the ALPN extension the non-null check is
    not sufficient.  Also check that the array is non-empty.

diff --git a/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java b/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
index 93fe703c..f84ae800 100644
--- a/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -347,7 +347,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
             throw new SSLException("Unable to enable SSL Handshake Callback on this SSLFDProxy instance.");
         }
 
-        if (alpn_protocols != null) {
+        if (alpn_protocols != null && alpn_protocols.length > 0) {
             byte[] wire_data = getALPNWireData();
             if (wire_data == null) {
                 throw new RuntimeException("JSSEngine.init(): ALPN wire data is NULL but alpn_protocols is non-NULL.");
```

frasertweedale (changes requested):

org/mozilla/jss/ssl/javax/JSSParameters.java
cipherboy: Sorry, how?

`appProtocols` by default is initialized to `null`.
If `alpn` is `null`, we leave the default.
If `alpn` is non-null, we continue to the string version of `setApplicationProtocols`.

`setApplicationProtocols(String[] protocols)` sets `appProtocols` to `null` if `protocols` is `null`. If not, we build a byte array (checking lengths and ensuring no interior protocol is `null`), and pass it to  `setApplicationProtocols(byte[][] protocols)`.

This last method sets `appProtocols` to `null` if `protocols` is `null` or has zero length. This prevents us from setting an empty array (0 length array) as a non-`null` `appProtocols` list.

Otherwise, we validate lengths, ensure no interior member is `null` and then construct the result as a new `byte[][] appProtocols`.

frasertweedale: > Sorry, how?

This `JSSParameters` constructor takes an `SSLParameters`.  The OpenJDK `SSLParameters` class initialises the application protocols to empty array, not null.  See
https://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/77b682e2d679/src/share/classes/javax/net/ssl/SSLParameters.java#l88 .

So aside from good hygiene, we really do need to handle the `(alpn != null && alpn.length < 1)` case.

cipherboy: Like I said, this is already handled by `setApplicationProtocols(byte[][] protocols)`. The following code passes on `9c3f195b3114a5c4a1c33653ada7768f1af67987` (You can see above when I pushed this 5 days ago).

What commit are you using?

I pushed an explicit test case on top of it in `a6c2f8ccf57f7fa1caec940944918ed7803fb5ce`. That's the only commit I introduced today.

This is already handled.

cipherboy:
@frasertweedale If you have cycles, do you mind looking at jdk8 failure? This started recently and didn't happen on the initial code, so I'm suspicious it was something simple. 


If you patch the Dockerfile:
```diff
diff --git a/tools/Dockerfiles/ubuntu_jdk8 b/tools/Dockerfiles/ubuntu_jdk8
index 48cc3720..64a561fe 100644
--- a/tools/Dockerfiles/ubuntu_jdk8
+++ b/tools/Dockerfiles/ubuntu_jdk8
@@ -24,11 +24,12 @@ COPY . /home/sandbox/jss
 
 # Perform the actual build
 WORKDIR /home/sandbox/jss
-CMD true \
+RUN true \
         && rm -rf build \
         && mkdir build \
         && cd build \
         && CFLAGS="-Wall -Wextra -Werror" cmake .. \
         && make all \
-        && ctest --output-on-failure \
         && true
+
+CMD /bin/bash
```

It will let you run but I don't know how to enable gdb inside a container. Perhaps I just need to build and run it as root.

org/mozilla/jss/ssl/javax/JSSParameters.java
frasertweedale: Thanks, you are right, I missed that the fix was introduced with the `byte[]`-based method commit.

frasertweedale:
> @frasertweedale If you have cycles, do you mind looking at jdk8 failure? This started recently and didn't happen on the initial code, so I'm suspicious it was something simple.

I don't have cycles to deeply investigate.  I created another PR to test this one alongside the PR #642.  Note the container is a based on a rolling release so maybe something changed in the base image?  We could also "bisect" the commits in this PR to narrow down the change that actually triggers the failure.

frasertweedale:
@cipherboy hm, yeah it sure is weird.  Only Ubuntu, only FIPS, only when ALPN is used.  I note that ubuntu "focal" is at NSS v3.49 - https://packages.ubuntu.com/source/focal/nss.  But Centos 7 is at 3.44 and f31 and f32 are at 3.53 and 3.55 respectively, so the version doesn't seem to be the issue.

Looking further, it is the server `unwrap` at NEED_UNWRAP that is failing.  But you would really have to get more logging to work out why it is failing.  Rather than trying to debug inside the container, maybe just make an Ubuntu VM and try and repro there?

I'd have a look at what is enabled/disabled in FIPS mode on Ubuntu; might give a clue.

frasertweedale:
Traceback:
```
INFO: [JSS Server for ALPN] JSSEngine.unwrap() - Got SSLException: javax.net.ssl.SSLException: Unexpected return from PR.Read(): SSL_ERROR_HANDSHAKE_FAILED (-12116)
Exception in thread "main" javax.net.ssl.SSLException: Unexpected return from PR.Read(): SSL_ERROR_HANDSHAKE_FAILED (-12116)
	at org.mozilla.jss.ssl.javax.JSSEngineReferenceImpl.unwrap(JSSEngineReferenceImpl.java:1257)
	at java.base/javax.net.ssl.SSLEngine.unwrap(SSLEngine.java:637)
	at org.mozilla.jss.tests.TestSSLEngine.testHandshake(TestSSLEngine.java:468)
	at org.mozilla.jss.tests.TestSSLEngine.testInitialHandshake(TestSSLEngine.java:786)
	at org.mozilla.jss.tests.TestSSLEngine.testALPNHandshake(TestSSLEngine.java:1005)
	at org.mozilla.jss.tests.TestSSLEngine.testNativeClientServer(TestSSLEngine.java:1039)
	at org.mozilla.jss.tests.TestSSLEngine.main(TestSSLEngine.java:1095)
```

frasertweedale:
Possibly relevant: Ubuntu patches nss to always return FALSE from NSS "is fips enabled?" subroutines: https://git.launchpad.net/ubuntu/+source/nss/tree/debian/patches/disable_fips_enabled_read.patch?id=af745d02d46dd21c52abe39093188ac6e0918df3

cipherboy:
Hmmm ok. I'll try the bisect method later. Curious. :-)

PR #531, opened 2020-04-30
Update the homepage for gh-pages

SilleBille:
- Add 4.6.x docs to the index page

PS: `*/docs/index.html` is missing


cipherboy:
I've been meaning to make `docs/index.html` auto-generate from a directory listing, but never got there on the backend. Feel free to work on that if you get time @SilleBille!

PR #532, opened 2020-04-30
SSLSocket from SSLEngine

cipherboy:
Now that we have an `SSLEngine` implementation in hand, various other applications and libaries (such as resteasy) expect `SSLSocket`  as well. The issue is that our existing `SSLSocket` (under `org.mozilla.jss.ssl.SSLSocket`) doesn't conform at all to the `javax.net.ssl.SSLSocket` abstract class. 

Even if we hacked a `javax`-conforming `SSLSocket` out of a `jss.ssl.SSLSocket`, we'd still need to figure out how to allocate sockets correctly and close the existing one -- but immediate reuse might not be possible!

So, implement `SSLSocket` using our new `SSLEngine`. 

TODO:

 - [x] Implement SSLSocket
     - [x] Figure out underlying socket construction
     - [x] Figure out consuming an existing `Socket`.
     - [x] Figure out reading/writing
     - [x] Figure out SocketChannel
 - [x] Implement SSLServerSocket (#539)
 - [x] Implement SSLSocketFactory (#540)
 - [x] Implement SSLServerSocketFactory (#540)
 - [x] Clean up
 - [x] Provide javadocs
 - [x] Verify JSSSocketChannel implementation
 - [ ] Test:
     - [ ] with old jss.ssl.SSLSocket
     - [ ] with javax SSLSocket
     - [x] with BadSSL (#540)

### DIFF
--- a/org/mozilla/jss/ssl/javax/JSSSocket.java
+++ b/org/mozilla/jss/ssl/javax/JSSSocket.java
@@ -1,0 +1,881 @@
+package org.mozilla.jss.ssl.javax;
+
+import java.io.*;
+import java.net.*;
+import java.nio.channels.*;
+import java.security.*;
+import java.util.*;
+import javax.net.ssl.*;
+
+import org.mozilla.jss.pkcs11.PK11Cert;
+import org.mozilla.jss.pkcs11.PK11PrivKey;
+import org.mozilla.jss.provider.javax.crypto.JSSKeyManager;
+import org.mozilla.jss.provider.javax.crypto.JSSTrustManager;
+
+/**
+ * SSL-enabled socket following the javax.net.ssl.SSLSocket interface.
+ *
+ * Most users will want to use the JSSSocketFactory provided by the Java
+ * Provider interface instead of using this class directly.
+ *
+ * This SSLSocket implementation is a wrapped implementation. In particular, we
+ * need to consume an existing Socket (via the consumeSocket(...) call) which
+ * we actually send data over. When called from a socket factory, this
+ * additional socket will be automatically created for the caller. This
+ * is necessary because SSLSocketFactory includes a mode which wraps an
+ * existing socket.
+ *
+ * All JSSSocket instances have a underlying SocketChannel, of type
+ * JSSSocketChannel. Notably lacking is a javax.net.ssl.SSLSocketChannel type,
+ * so JSSSocketChannel includes no additional SSL-specific options. However,
+ * the core of the SSLEngine wrapping logic exists there.
+ *
+ * In order to interoperate with JSSEngine, many of the adjacent methods have
+ * been included in this class as well.
+ *
+ * This socket can either be a client or a server, depending on how it was
+ * created. For more information, see the javax.net.ssl.SSLSocket
+ * documentation.
+ *
+ * To construct a (useful) new instance, the following calls must be made:
+ *
+ * - new JSSSocket();
+ * - consumeSocket(inst);
+ * - initSSLEngine(...);
+ * - setKeyManagers(...);
+ * - setTrustManagers(...);
+ *
+ * Optionally, setSSLContext(...) could be called to provide the SSLContext
+ * from which the SSLEngine should be constructed. This should be called prior
+ * to initSSLEngine(...) being called.
+ */
+public class JSSSocket extends SSLSocket {
+    /**
+     * Name of the SSLEngine protocol to use.
+     */
+    private String engineProviderProtocol = "TLS";
+
+    /**
+     * Name of the SSLEngine provider to use.
+     */
+    private String engineProvider = "Mozilla-JSS";
+
+    /**
+     * SSLContext to use to create the JSSEngine. Note that JSSSocket will fail
+     * if the context doesn't create JSSEngine instances.
+     */
+    private SSLContext jssContext;
+
+    /**
+     * JSSEngine instance to utilize for SSLEngine operations.
+     */
+    private JSSEngine engine;
+
+    /**
+     * All registered handshake callbacks.
+     */
+    private ArrayList<HandshakeCompletedListener> handshakeCallbacks = new ArrayList<HandshakeCompletedListener>();
+
+    /**
+     * The socket this JSSSocket was created over; all read/write operations
+     * go through this socket and all information exposed via Socket members
+     * go through here.
+     */
+    private Socket parent;
+
+    /**
+     * Previously consumed data, if any; utilized for certain SSLSocketFactory
+     * calls.
+     */
+    private InputStream consumedData;
+
+    /**
+     * Underlying SocketChannel for this socket; always exists.
+     */
+    private JSSSocketChannel channel;
+
+    /**
+     * Whether or not to automatically close the underlying Socket when this
+     * socket has been closed; defaults to true.
+     */
+    private boolean autoClose = true;
+
+    /**
+     * Start building a new JSSSocket.
+     *
+     * We specifically avoid creating any other constructors as we wish to
+     * consume an existing socket rather than creating a new one.
+     */
+    public JSSSocket() {}
+
+    /**
+     * Consume a parent socket, utilizing it for all read/write operations.
+     *
+     * This JSSSocket instance will inherit all information about the
+     * connection from this underlying socket. When utilized in a JSSSocket,
+     * callers should refrain from interacting with the underlying socket
+     * directly until the TLS connection is closed. Otherwise, messages may
+     * get dropped.
+     */
+    public void consumeSocket(Socket parent) throws IOException {
+        if (parent == null) {
+            String msg = "Unable to consume and utilize null parent socket!";
+            throw new IOException(msg);
+        }
+
+        this.parent = parent;
+    }
+
+    /**
+     * Get the SSLContext if one exists or create a new instance.
+     *
+     * This is used by initSSLEngine(..) to create the underlying SSLEngine.
+     */
+    protected SSLContext getSSLContext() throws IOException {
+        if (jssContext == null) {
+            try {
+                jssContext = SSLContext.getInstance(engineProviderProtocol, engineProvider);
+            } catch (Exception e) {
+                String msg = "Unable to create JSSSocket prior to Mozilla-JSS ";
+                msg += "initialization! " + e.getMessage();
+                throw new IOException(msg, e);
+            }
+        }
+
+        return jssContext;
+    }
+
+    /**
+     * Explicitly set the SSLContext utilized by this JSSSocket instance.
+     *
+     * This enables JSSServerSocket to copy its SSLContext over to the accepted
+     * JSSSocket.
+     */
+    public void setSSLContext(SSLContext ctx) throws IOException {
+        jssContext = ctx;
+    }
+
+    /**
+     * Initialize the underlying SocketChannel.
+     */
+    private void init() throws IOException {
+        if (engine == null) {
+            initEngine();
+        }
+
+        SocketChannel parentChannel = parent.getChannel();
+
+        if (parentChannel == null) {
+            ReadableByteChannel read = Channels.newChannel(parent.getInputStream());
+            WritableByteChannel write = Channels.newChannel(parent.getOutputStream());
+
+            channel = new JSSSocketChannel(this, parent, read, write, engine);
+        } else {
+            channel = new JSSSocketChannel(this, parentChannel, engine);
+        }
+
+        channel.setConsumedData(consumedData);
+        channel.setAutoClose(autoClose);
+    }
+
+    /**
+     * Explicitly initialize the SSLEngine with no session resumption
+     * information.
+     */
+    public void initEngine() throws IOException {
+        engine = (JSSEngine) getSSLContext().createSSLEngine();
+    }
+
+    /**
+     * Explicitly initialize the SSLEngine with information for session
+     * resumption, including peer's hostname and port.
+     */
+    public void initEngine(String host, int port) throws IOException {
+        engine = (JSSEngine) getSSLContext().createSSLEngine(host, port);
+    }
+
+    /**
+     * Get the underlying JSSEngine instance.
+     *
+     * Note that, just like accessing the underlying Socket instance while the
+     * JSSSocket instance is still open is dangerous, performing other TLS
+     * operations directly via JSSEngine is also dangerous. This is mostly
+     * exposed to enable advanced configuration of the JSSEngine that isn't
+     * otherwise allowed by JSSSocket, and to facilitate the accept() method
+     * on JSSServerSocket.
+     */
+    public JSSEngine getEngine() {
+        return engine;
+    }
+
+    /**
+     * Get the underlying SocketChannel for this Socket.
+     *
+     * @see java.net.Socket#getChannel()
+     */
+    public JSSSocketChannel getChannel() {
+        if (parent.getChannel() == null) {
+            return null;
+        }
+
+        return getInternalChannel();
+    }
+
+    /**
+     * Helper to always return the channel for this socket,
+     * initializing it if it isn't yet created.
+     */
+    protected JSSSocketChannel getInternalChannel() {
+        if (channel == null) {
+            try {
+                init();
+            } catch (IOException e) {
+                throw new RuntimeException("Unexpected error trying to construct channel: " + e.getMessage(), e);
+            }
+        }
+
+        return channel;
+    }
+
+    /**
+     * Get a copy of an input stream for this Socket.
+     *
+     * @see java.net.Socket#getInputStream()
+     */
+    public InputStream getInputStream() throws IOException {
+        if (channel == null) {
+            init();
+        }
+
+        return Channels.newInputStream(channel);
+    }
+
+    /**
+     * Get a copy of an output stream for this Socket.
+     *
+     * @see java.net.Socket#getOutputStream()
+     */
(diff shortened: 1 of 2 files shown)


org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,449 @@
...
+            // amount, but in the blocking case, we still need to bound our
+            // read.
+            return suggested;
+        }
+
+        // In both remaining case (no channel or channel is blocking), bound
+        // the read above by the available data in the socket's input stream.
+
+        int available = parentSocket.getInputStream().available();
+        return Math.max(suggested, available);
```
cipherboy, line 94: Should be `Math.min`.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,449 @@
...
+        return (int) read(new ByteBuffer[] { dst });
+    }
+
+    public long read(ByteBuffer[] dsts, int offset, int length) throws IOException {
+        if (!finishConnect()) {
+            throw new IOException("Unable to finish handshake for an unknown reason.");
+        }
+
+        long capacity = computeSize(dsts, offset, length);
+        int buffer_size = boundRead((int) Math.max(capacity, Integer.MAX_VALUE));
```
cipherboy, line 231: Again, `Math.min`

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,449 @@
...
+        return (int) write(new ByteBuffer[] { src });
+    }
+
+    public long write(ByteBuffer[] srcs, int offset, int length) throws IOException {
+        if (!finishConnect()) {
+            throw new IOException("Unable to finish handshake for an unknown reason.");
+        }
+
+        long capacity = computeSize(srcs, offset, length);
+        int buffer_size = boundRead((int) Math.max(capacity, Integer.MAX_VALUE));
```
cipherboy, line 281: `Math.min` here as well.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,462 @@
...
+
+    public JSSSocketChannel(JSSSocket sslSocket, SocketChannel parent, JSSEngine engine) {
+        super(null);
+
+        this.sslSocket = sslSocket;
+        this.parent = parent;
+        this.parentSocket = parent.socket();
+        readChannel = parent;
+        writeChannel = parent;
+        this.engine = engine;
```
cipherboy, line 43: need to call `configureBlocking()`.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,465 @@
...
+    private ReadableByteChannel readChannel;
+    private WritableByteChannel writeChannel;
+    private JSSEngine engine;
+
+    private InputStream consumed;
+    private boolean autoClose = true;
+
+    private boolean inboundClosed = false;
+    private boolean outboundClosed = false;
+
```
cipherboy, line 34: Move byte buffers here and make them sized based on `engine.getSession().getApplicationBufferSize()` rather than the internal constant value.

cipherboy: Could also move `empty` here.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,465 @@
...
+        return (int) read(new ByteBuffer[] { dst });
+    }
+
+    public long read(ByteBuffer[] dsts, int offset, int length) throws IOException {
+        if (inboundClosed) {
+            return -1;
+        }
+
+        int buffer_size = boundRead(JSSEngine.BUFFER_SIZE);
+        ByteBuffer src = ByteBuffer.allocate(buffer_size);
```
cipherboy, line 226: If we switch to static buffers, we can change this call to:

```
read_buffer.clear();
ByteBuffer.wrap(read_buffer.array(), 0, buffer_size);
```

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,465 @@
...
+
+        long remoteRead = readChannel.read(src);
```
cipherboy, line 228: @edewata Should we move this inside the loop below? In particular, if we read a non-zero amount and then unwrap it all, we could retry the read and potentially get more data. Thoughts? 

Note that we'd have to handle the case where the `consumed` has data left, but we don't currently handle that either way.

It'd make the `read()` code look more like `write()` though.

Thoughts?

cipherboy: In particular, to get `consumed` to work, we replace `readChannel` with  `Channels.newChannel(consumed)` and that'll give us something we can `.read(src)` into -- we just need to make sure to check `consumed.available` and limit the size of `buffer_size` above appropriately.

cipherboy: I've gotten `consumed` to work but I haven't moved the read inside the thing as it'd involve resizing the `src` buffer afterwards.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,495 @@
...
+                    msg += "calls to unwrap stalled, consuming and producing ";
+                    msg += "no data: unwrapped " + unwrapped + " bytes of ";
+                    msg += remoteRead + " bytes read from remote peer; got a ";
+                    msg += "decrypted size of " + decrypted + " bytes.";
+                    throw new IOException(msg);
+                }
+            } while (unwrapped < remoteRead);
+        } catch (SSLException ssle) {
+            String msg = "Error attempting to read from remote peer: ";
+            msg += "got unexpected exception during unwrap call: ";
```
edewata, line 288: The exception message probably can be shortened like "Unable to unwrap SSL data:".

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,495 @@
...
+
+                unwrapped += result.bytesConsumed();
+                decrypted += result.bytesProduced();
+
+                if (unwrapped < remoteRead && result.bytesConsumed() == 0 && result.bytesProduced() == 0) {
+                    String msg = "Error attempting to read from remote peer: ";
+                    msg += "calls to unwrap stalled, consuming and producing ";
+                    msg += "no data: unwrapped " + unwrapped + " bytes of ";
+                    msg += remoteRead + " bytes read from remote peer; got a ";
+                    msg += "decrypted size of " + decrypted + " bytes.";
```
edewata, line 282: This is a bit long as an exception message. I'm not sure the actual numbers would be useful for debugging. Should we just say something like "Incomplete SSL unwrap" instead?

org/mozilla/jss/ssl/javax/JSSSocket.java
```diff
@@ -0,0 +1,869 @@
...
+ * - consumeSocket(inst);
+ * - initSSLEngine(...);
+ * - setKeyManagers(...);
+ * - setTrustManagers(...);
+ *
+ * Optionally, setSSLContext(...) could be called to provide the SSLContext
+ * from which the SSLEngine should be constructed. This should be called prior
+ * to initSSLEngine(...) being called.
+ */
+public class JSSSocket extends SSLSocket {
```
edewata, line 52: In the future we probably could change `org.mozilla.jss.ssl.SSLSocket` to extend `JSSSocket`, and eventually merge it into `JSSSocket`.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,495 @@
...
+        SSLEngineResult.HandshakeStatus state = engine.getHandshakeStatus();
+        if (state == SSLEngineResult.HandshakeStatus.NOT_HANDSHAKING) {
+            return true;
+        }
+
+        int handshakeAttempts = 0;
+        int maxHandshakeAttempts = 100;
+
+        if (!isBlocking()) {
+            // When we're a non-blocking socket/channel, we'd farther rather
```
edewata, line 142: Is this phrase correct?

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,495 @@
...
+import java.security.*;
+import java.util.*;
+
+import javax.net.ssl.*;
+
+/**
+ * SSL-enabled SocketChannel following the javax.net.ssl.SSLSocket interface.
+ *
+ * This class should never be constructed directly and instead only accessed
+ * once a Socket is wrapped in a JSSSocket.
```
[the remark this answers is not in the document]
cipherboy: So we're not creating a `SocketChannel` here with `JSSSocketChannel`, just like we're not creating a `Socket` in the `JSSSocket` -- we're creating what _would_ be a `SSLSocketChannel` implementation. But such a thing doesn't (apparently) exist in the standard library. 

We're upgrading/"consuming" an existing `SocketChannel` and `Socket`, wrapping it in `JSSSocketChannel` and `JSSSocket`, and making the result do TLS over this wrapped `Socket`/`SocketChannel`. 

---

Part of the problem here is we need to blend two worlds: the blocking world of `SSLSocket` and the non-blocking world of `SSLEngine`. I chose to take the consumed `Socket` and convert the `InputStream`/`OutputStream`s into `Channel`s so that we can more easily utilize them with `SSLEngine`. Otherwise we'd have to put `byte[]` buffers directly inside `JSSSocket` and `.wrap` them to `ByteBuffer`s for use with `SSLEngine`.

We're not trying to directly mimic what the SunJSSE `SSLSocketImpl` does here either. It has very restrictive usages that the original `org.mozilla.jss.ssl.SSLSocket` doesn't have and we don't need either:

 1. If your socket isn't connected, SunJSSE's SocketFactory won't consume it via `createSocket(Socket, <host>, <port>, <auto close>)` -- it'll fail saying that `Socket` should've been connected first. We instead create an unconnected socket and leave it to the caller to call `connect` the new socket.
 2. If your socket is non-blocking, it'll fail to work. `startHandshake()` will fail with an invalid blocking mode exception. This is because they internally didn't implement any channel semantics.

Since we can use the JSSSocketChannel to provide both blocking and non-blocking semantics, I think we should. It lets people use a non-blocking channel without manually maintaining a SSLEngine instance. It just requires them to go through a SSLSocketFactory in order to do it, which I think is fine.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
cipherboy: farther rather :laughing: Nah, it isn't. I'll update this.

org/mozilla/jss/ssl/javax/JSSSocket.java
```diff
@@ -0,0 +1,869 @@
...
+        SocketChannel parentChannel = parent.getChannel();
+
+        if (parentChannel == null) {
+            ReadableByteChannel read = Channels.newChannel(parent.getInputStream());
+            WritableByteChannel write = Channels.newChannel(parent.getOutputStream());
+
+            channel = new JSSSocketChannel(this, parent, read, write, engine);
+        } else {
+            channel = new JSSSocketChannel(this, parentChannel, engine);
+        }
```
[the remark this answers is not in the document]
cipherboy: See above: we're using this for convenience of implementation. I'll change `getChannel()` to return `null` if `parent.getChannel()` returns `null`. But no, we're not opening a new socket channel here -- we're _wrapping_ an existing one.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -0,0 +1,449 @@
...
+     */
+    private int boundRead(int suggested) throws IOException {
+        return (int) boundRead((long) suggested);
+    }
+
+    /**
+     * Internal helper to bound the size of a blocking read to the maximum
+     * data available.
+     */
+    private long boundRead(long suggested) throws IOException {
```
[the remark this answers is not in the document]
cipherboy: Ah sorry. This is where I dislike `SSLEngine` vs `SSLSocket` vs `SocketChannel` semantics here: `SSLEngine` takes an `int` of ByteBuffers (each with a size in `int`) -- which means the total size would only fit in a `long`, not `int` -- but is only allowed to return an `int` worth of write data.

`SocketChannel`, on the other hand, recognizes this and allows the result to fit in a `long`. 

I was going back and forth over how this was implemented (and where all I'd place this call), and I think I can remove the `long` version.

org/mozilla/jss/ssl/javax/JSSSocket.java
```diff
@@ -0,0 +1,869 @@
...
+     *
+     * This is used by initSSLEngine(..) to create the underlying SSLEngine.
+     */
+    protected SSLContext getSSLContext() throws IOException {
+        if (jssContext == null) {
+            try {
+                jssContext = SSLContext.getInstance(engineProviderProtocol, engineProvider);
+            } catch (Exception e) {
+                String msg = "Unable to create JSSSocket prior to Mozilla-JSS ";
+                msg += "initialization! " + e.getMessage();
```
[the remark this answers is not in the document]
cipherboy: I ended up not really using this method. There's two issues with it:

 1. We have no real way of _initializing_ a JSSContext is one doesn't exist,
 2. We have no other way of creating `SSLEngine`'s. 

The problem is that, if we don't pass an explicit `SSLContext` instance (backed by our `JSSContextSpi`), we need to initialize it somehow. That means divining `KeyManager`s and `TrustManager`s to use.

Since we have a split `JSSEngine` interface/implementation (and `JSSContextSpi` currently controls which implementation gets selected -- where currently `JSSEngineReferenceImpl` is the only one available), we need _some_ context to use here.

When we construct a `JSSSocketFactory` from _within_ `JSSContextSpi` we additionally don't have a way of accessing the _parent_ `SSLContext` of this `JSSContextSpi` either! So we end up creating a new one! We could perhaps use the `SSLContext(SSLContextSpi, String, String)` constructor though... but we'd need to divine the `protocol` parameter still....

Thoughts for avoiding this problem definitely appreciated. :-)

org/mozilla/jss/ssl/javax/JSSSocket.java
```diff
@@ -0,0 +1,869 @@
...
+
+        boolean status = channel.finishConnect();
+        if (!channel.isBlocking()) {
+            // SSLSocket semantics explicitly say:
+            //     > This method is synchronous for the initial handshake on
+            //     > a connection and returns when the negotiated handshake is
+            //     > complete.
+            // so we have to block until the connection is complete. But use an
+            // exponential backoff so we have a chance of catching any peer
+            // data.
```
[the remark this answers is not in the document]
cipherboy: Sorry, linear was the intention.

org/mozilla/jss/ssl/javax/JSSSocket.java
cipherboy: `org.mozilla.jss.ssl.SSLSocket` doesn't respect many of the semantics around the `javax.net.ssl` suite:

 - `KeyManager` for key selection,
 - `TrustManager` for cert validation,
 - Converting between `SSLCipher` and JDK-standard names is easier.

We could fix the latter, but the former would be more invasive changes. Ideally we'd like our `javax.net.ssl.SSLSocket` implementation to behave as similarly to our `javax.net.ssl.SSLEngine` as possible. For people not expecting a `javax.net.ssl.SSLSocket` and who want a generic `Socket` (that happens to do TLS under the covers), we're fine keeping `org.mozilla.jss.ssl.SSLSocket`. I guess that's LDAP JDK and some Dogtag PKI helpers. But for any third-party library, I think `JSSSocket` is fine? I dunno. At any rate, not a problem to solve in this PR :-)

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
cipherboy: The biggest concern was if `unwrapped` was 0 and `remoteRead` was non-zero; it'd indicate that the SSLEngine implementation was flawed.

cipherboy: I've shortened this message though.

cipherboy:
@edewata This has been updated. Thoughts on my comments above?

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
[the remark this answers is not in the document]
cipherboy: @edewata  Do you think #2 makes sense? Should I also create a generic `SSLSocketChannel` that brings some of the methods from `SSLSocket` onto a `SSLSocketChannel`?

cipherboy:
Thanks!

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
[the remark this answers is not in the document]
edewata: These are tough questions! :)
I'd say go ahead and implement anything that we need to get it working,
and feel free to implement anything extra that will make JSS more usable
in general.

PR #535, opened 2020-05-01
JSSEngine: Add getSSLParameters

cipherboy:
This returns a new instance of `JSSParameters` which can be used to clone
the configuration of this `SSLEngine` into another. This is helpful for
implementing `SSLServerSocket`, which must `accept()`, creating a new
`SSLSocket` with the same initial configuration.

----

Only the top commit is new.

 - ~Blocked by #526 ; will be rebased once that merges~
 - Blocks #532 ; provides a method necessary for SSLServerSocket

### DIFF
--- a/org/mozilla/jss/ssl/javax/JSSEngine.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngine.java
@@ -51,6 +51,11 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
     protected String hostname;
 
     /**
+     * Certificate alias used by the JSSEngine instance.
+     */
+    protected String certAlias;
+
+    /**
      * Certificate used by this JSSEngine instance.
      *
      * Selected and inferred from the KeyManagers passed, when not passed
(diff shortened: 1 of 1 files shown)


org/mozilla/jss/util/JDKCompat.java
```diff
@@ -0,0 +1,21 @@
...
+public class JDKCompat {
+    public static class SSLParametersHelper {
+        public static String[] getApplicationProtocols(SSLParameters inst) {
+            try {
+                Method getter = inst.getClass().getMethod("getApplicationProtocols");
+                Object result = getter.invoke(inst);
+                if (result instanceof String[]) {
+                    return (String[]) result;
+                }
+            } catch (Throwable t) {}
```
edewata, line 16: If we want to make ALPN support optional, we should only catch `NoSuchMethodException` and explicitly ignore or a generate warning:
```
} catch (NoSuchMethodException e) {
    // ignore
}
```
All other exceptions should still be propagated to the callers.

edewata: Also, it's not necessary to check the type of `result` since it's always `String[]` in all Java versions (if the method is available).

cipherboy:
@edewata I separated this one from the ALPN PR (#526). I think this is simple enough that I'll just merge it.

PR #539, opened 2020-05-06
JSSServerSocket to JSSSocket

cipherboy:
Based off of #532. Will be rebased once this is merged.

---

There are three major functionalities in SSLContext:
 - `SSLEngine` -- merged in #150.
 - `SSLSocketFactory`
 - `SSLServerSocketFactory`

#532 and this PR introduce the underlying `JSSSocket` and `JSSServerSocket` implementations these factories will return. Note that while `SSLSocketFactory` doesn't explicitly return a `SSLSocket` -- but its javadocs say that it does. Many places (including in OpenJDK itself!) blindly cast the Socket resulting from a `SSLSocketFactory` to `SSLSocket` without checking if it is possible. This makes returning the existing `org.mozilla.jss.ssl.SSLSocket` impossible, unless it is extended to implement the `javax.net.ssl.SSLSocket` interface.

### DIFF
--- a/org/mozilla/jss/ssl/javax/JSSServerSocketChannel.java
+++ b/org/mozilla/jss/ssl/javax/JSSServerSocketChannel.java
@@ -1,0 +1,127 @@
+package org.mozilla.jss.ssl.javax;
+
+import java.io.*;
+import java.net.*;
+import java.nio.ByteBuffer;
+import java.nio.channels.*;
+import java.security.*;
+import java.util.*;
+
+import javax.net.ssl.*;
+
+public class JSSServerSocketChannel extends ServerSocketChannel {
+    private JSSServerSocket sslSocket;
+    private ServerSocketChannel parent;
+    private ServerSocket parentSocket;
+    private JSSEngine engine;
+
+    public JSSServerSocketChannel(JSSServerSocket sslSocket, ServerSocketChannel parent, JSSEngine engine) throws IOException {
+        super(null);
+
+        this.sslSocket = sslSocket;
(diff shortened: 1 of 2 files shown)


org/mozilla/jss/ssl/javax/JSSServerSocketChannel.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    public JSSSocketChannel accept() throws IOException {
+        if (parent == null) {
+            // Have to be in blocking mode. Call up to sslSocket to handle
+            // the accept. Note that JSSSocket always has a SocketChannel even
+            // when the underlying socket doesn't so the call to getChannel()
+            // will be non-null.
+            JSSSocket acceptedSocket = sslSocket.accept();
+            return acceptedSocket.getInternalChannel();
```
cipherboy, line 49: @edewata What do you think about this? It isn't technically right, but if we return `null` here, we could lose a client connection. Thoughts?

edewata: I'm not quite sure about the implication of each option, but if we don't want to deal with `parent == null` case we probably can throw an exception (either here or maybe in the constructor). That way people using JSS will know that it's not a supported configuration. As long as we provide a clear error message we don't have to worry about losing a client connection.

cipherboy: Okie dokie, I'll add the exception as a later PR. Thanks!

PR #546, opened 2020-05-08
SSLSocket: detect failure to recieve CLOSE_NOTIFY

cipherboy:
An important part of the TLS protocol is the ability to detect whether
or not the peer closed the connection or if an attacker terminated the
connection prematurely. In order to do this, both sides send a
`CLOSE_NOTIFY` event. When this event isn't detected, raise an `IOException`
to inform the application about it.

### DIFF
--- a/org/mozilla/jss/ssl/javax/JSSSocketChannel.java
+++ b/org/mozilla/jss/ssl/javax/JSSSocketChannel.java
@@ -344,35 +344,66 @@ public class JSSSocketChannel extends SocketChannel {
         // to confirm the peer got the message. Otherwise, only a single write
         // is necessary to send our acknowledgement of the peer's alert.
 
-        synchronized (this) {
-            // Bypass read check.
-            inboundClosed = false;
-            outboundClosed = false;
+        try {
+            synchronized (this) {
+                // unwrap() triggers a call to PR_Read(), which in turn will
+                // execute the received alert callback. However, PR_Read is
+                // effectively a no-op with an empty buffer, resulting in the
+                // callback never triggering. Use a single byte buffer instead,
+                // discarding any data because we're closing the channel. This
+                // should ensure we always get a callback.
+                ByteBuffer read_one = ByteBuffer.allocate(1);
 
-            read(empty);
-            engine.closeOutbound();
-            write(empty);
-            read(empty);
-            engine.closeInbound();
 
-            outboundClosed = true;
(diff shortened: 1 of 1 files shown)


org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -344,35 +344,64 @@ public void implCloseSelectableChannel() throws IOException {
...
-                parentSocket.shutdownInput();
-                parentSocket.shutdownOutput();
-                parentSocket.close();
+                    try {
+                        Thread.sleep(times * 10);
+                    } catch (Exception e) {}
+                }
+
+                if (!engine.isInboundDone()) {
+                    String msg = "Premature close of SSLSocket: peer didn't ";
```
cipherboy, line 381: @edewata What's your thoughts on this? `org.mozilla.jss.ssl.SSLSocket` doesn't implement this detection and will happily close. [TLSv1.1 made it optional](https://security.stackexchange.com/questions/82028/ssl-tls-is-a-server-always-required-to-respond-to-a-close-notify/82034) and Pornin seems to think its a non-issue. 

In [TLSv1.3 the new spec says](https://tools.ietf.org/html/rfc8446) on page 86:

>    Each party MUST send a "close_notify" alert before closing its write
>    side of the connection, unless it has already sent some error alert.
>    This does not have any effect on its read side of the connection.
>   Note that this is a change from versions of TLS prior to TLS 1.3 in
>   which implementations were required to react to a "close_notify" by
>   discarding pending writes and sending an immediate "close_notify"
>   alert of their own.  That previous requirement could cause truncation
>   in the read side.  Both parties need not wait to receive a
>   "close_notify" alert before closing their read side of the
>   connection, though doing so would introduce the possibility of
>   truncation.

Which seems to imply that we're safe to close the outbound side earlier, and that lack of a CLOSE_NOTIFY should be treated as fatal. Thoughts?

cipherboy: Looks like `sun.security.ssl.SSLSocketImpl` (SunJSSE's implementation) will wait for the `CLOSE_NOTIFY` event if it is the one initiating the close. That's essentially the behavior here: when we receive a `CLOSE_NOTIFY`, we mark `isInboundDone() == true` inside the SSLEngine. If it hasn't been received (i.e., we just sent it), we go into the loop and wait.

PR #554, opened 2020-05-14
SSLEngine - Test improvements

cipherboy:
This improves the test by reusing buffers between tests, reducing the load on the allocator and garbage collector. 

This doesn't explicitly improve the speed though.


cipherboy:
Ugh, this still doesn't work.

PR #556, opened 2020-05-18
Basic tests for PKCS#10

cipherboy:
Basic sanity tests for both of JSS's PKCS#10 interfaces.

### DIFF
--- a/org/mozilla/jss/tests/PKCS10Test.java
+++ b/org/mozilla/jss/tests/PKCS10Test.java
@@ -1,0 +1,20 @@
+package org.mozilla.jss.tests;
+
+import java.security.PublicKey;
+import java.security.KeyPair;
+import java.security.interfaces.*;
+
+import org.mozilla.jss.*;
+import org.mozilla.jss.crypto.*;
+import org.mozilla.jss.netscape.security.pkcs.*;
+import org.mozilla.jss.netscape.security.x509.*;
+
+public class PKCS10Test {
+    public static void main(String[] args) throws Exception {
+        CryptoManager cm = CryptoManager.getInstance();
+
+        CertAndKeyGen ckg = new CertAndKeyGen("RSA", "SHA256withRSA");
+        ckg.generate(4096);
+        PKCS10 csr = ckg.getCertRequest(new X500Name("CN=localhost"));
+    }
+}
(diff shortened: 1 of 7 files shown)


org/mozilla/jss/tests/PKCS10Test.java
```diff
@@ -0,0 +1,20 @@
...
+import org.mozilla.jss.netscape.security.pkcs.*;
+import org.mozilla.jss.netscape.security.x509.*;
+
+public class PKCS10Test {
+    public static void main(String[] args) throws Exception {
+        CryptoManager cm = CryptoManager.getInstance();
+
+        CertAndKeyGen ckg = new CertAndKeyGen("RSA", "SHA256withRSA");
+        ckg.generate(4096);
+        PKCS10 csr = ckg.getCertRequest(new X500Name("CN=localhost"));
```
edewata, line 18: Could we add some validations for the CSR?

org/mozilla/jss/tests/TestCertificationRequest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.mozilla.jss.asn1.*;
+import org.mozilla.jss.crypto.*;
+import org.mozilla.jss.pkcs10.*;
+import org.mozilla.jss.pkix.primitive.*;
+
+public class TestCertificationRequest {
+    public static void main(String[] argv) throws Exception {
+        if (argv.length > 2 || argv.length < 1) {
+            System.out.println("Usage: TestCertificationRequest <dbdir> [<certfile>]");
+            System.exit(0);
```
edewata, line 16: I think it should generate a non-zero return code since it fails the test.

org/mozilla/jss/tests/TestCertificationRequest.java
```diff
@@ -0,0 +1,33 @@
...
+import java.io.*;
+import java.security.KeyPair;
+
+import org.mozilla.jss.*;
+import org.mozilla.jss.asn1.*;
+import org.mozilla.jss.crypto.*;
+import org.mozilla.jss.pkcs10.*;
+import org.mozilla.jss.pkix.primitive.*;
+
+public class TestCertificationRequest {
```
edewata, line 12: Do we run this test?

cipherboy: No, it requires a CSR to run so I ran it manually. We didn't previously run this test either, but it at least moves it into the correct place in case we want to run it automatically in the future. I think this test should/could be expanded with actual uses of these classes from Dogtag PKI, but we should leave it for a future sprint and/or QE to do.

PR #560, opened 2020-05-21
Related: Bug 1710105 - JSS: add RSA PSS support

jmagne:
Add PSS cases to algorithm name translating method.


cipherboy:
@jmagne Does this need a backport to `v4.4.x`?

jmagne:
Ascheel, this doesn't really need a backport to 4.4.x for the reason that pki is using the copy of this file still present internally. This fix addresses the fact that this file is one of a several that has been moved to jss and pki code now uses the jss version. I will leave it up to you if you would like to port it anyway, for completeness sake.

cipherboy:
On the off chance we ever spin a new JSS release, we might as well include this patch in `v4.4.x`. I'll open a PR later.

PR #558, opened 2020-05-20
Add SSLSocket benchmark

cipherboy:
This benchmark supports three providers:

 1. org.mozilla.jss.ssl.SSLSocket, named JSS.legacy
 2. org.mozilla.jss.ssl.javax.JSSSocket, named JSS.SSLSocket
 3. the JDK's SunJSSE provider's SSLSocket, named SunJSSE.SSLSocket

`Signed-off-by: Alexander Scheel <ascheel@redhat.com>`

----

This is the results from siege [0] ran via:

```bash
 $ siege -c 100 -b -t 5m https://localhost:8443
```

with the benchmarker set to send a faked 1024-byte message:

```bash
 $ ./run_test.sh org.mozilla.jss.tests.BenchmarkSSLSocket JSS.legacy Server_RSA 8443 1024
```

The server certificate is 4096-bits. The selection of cipher suite and
protocol is left at their defaults. This is on a Lenovo Thinkpad P50
with a `Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz` and 32GB of RAM. Each
request gets spun off and handled by a new thread.

Using the legacy `org.mozilla.jss.ssl.SSLSocket` (old NSS-based socket)
memory stays stable and under ~1-2% of total memory:

```json
{        "transactions":                       98588,
        "availability":                        100.00,
        "elapsed_time":                        299.43,
        "data_transferred":                    96.28,
        "response_time":                       0.30,
        "transaction_rate":                    329.25,
        "throughput":                          0.32,
        "concurrency":                         99.80,
        "successful_transactions":             98588,
        "failed_transactions":                 0,
        "longest_transaction":                 29.94,
        "shortest_transaction":                0.04
}
```

Using `javax.net.ssl.SSLSocket` provided by SunJSSE (but with JSS crypto
and potentially random) and stays under 1-2% of total memory:

```json
{        "transactions":                       2417,
        "availability":                        100.00,
        "elapsed_time":                        299.36,
        "data_transferred":                    2.36,
        "response_time":                       12.12,
        "transaction_rate":                    8.07,
        "throughput":                          0.01,
        "concurrency":                         97.82,
        "successful_transactions":             2417,
        "failed_transactions":                 0,
        "longest_transaction":                 21.27,
        "shortest_transaction":                1.63
}
```

**Note** that the above option was removed from the benchmark as it was significantly slower

Using `javax.net.ssl.SSLSocket` provided by SunJSSE (without JSS crypto,
[...]

PR #562, opened 2020-05-21
Remove JSSSocket wait-on-close, fix session cache creation

cipherboy:
There's two bugs found by #558: 

 - JSSSocket waiting-on-close takes a long time and doesn't work in all scenarios (e.g., wget won't send closure notifications so the socket will remain open forever). This also happens, e.g., with siege.
 - Session cache creation can race, causing the JVM to segfault in NSS code. We use an `AtomicBoolean` to avoid this potential and ensure we only initialize the session cache once.

We also add one refactoring commit:

 - Throw `SSLException`s instead of `RuntimeException`s during `JSSEngine` initialization. 

We should probably merge this ahead of #558 merging.

### DIFF
--- a/org/mozilla/jss/ssl/SSLServerSocket.java
+++ b/org/mozilla/jss/ssl/SSLServerSocket.java
@@ -9,10 +9,13 @@ import java.io.IOException;
 import java.net.Socket;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
+import javax.net.ssl.SSLException;
+
 import org.mozilla.jss.CryptoManager;
 import org.mozilla.jss.NotInitializedException;
 import org.mozilla.jss.crypto.ObjectNotFoundException;
 import org.mozilla.jss.crypto.TokenException;
+import org.mozilla.jss.ssl.javax.JSSEngine;
 
 /**
  * SSL server socket.
@@ -266,9 +269,21 @@ public class SSLServerSocket extends java.net.ServerSocket {
      *  will contain the session cache. If null is passed, the server default
      *  is used: <code>/tmp</code> on Unix and <code>\\temp</code> on Windows.
      */
-    public static native void configServerSessionIDCache(int maxSidEntries,
-        int ssl2EntryTimeout, int ssl3EntryTimeout, String cacheFileDirectory)
-        throws SocketException;
+    public static void configServerSessionIDCache(int maxSidEntries,
+        int ssl2EntryTimeout, int ssl3EntryTimeout, String cacheFileDirectory) throws SocketException {
+        try {
+            JSSEngine.initializeSessionCache(maxSidEntries, ssl3EntryTimeout, cacheFileDirectory);
+        } catch (SSLException parent) {
+            // Because JSSEngine.initializeSessionCache is utilized in
+            // javax's SSLEngine implementations, it can only throw a
+            // SSLException. However, in the event it fails, we should
+            // re-throw it here as a SocketException, keeping the signature
+            // the same.
+            SocketException se = new SSLSocketException(parent.getMessage());
+            se.addSuppressed(parent);
+            throw se;
+        }
+    }
 
     /**
      * Sets the certificate to use for server authentication.
--- a/org/mozilla/jss/ssl/javax/JSSEngine.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngine.java
@@ -1,6 +1,7 @@
 package org.mozilla.jss.ssl.javax;
 
 import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 import javax.net.ssl.*;
 
 import org.slf4j.Logger;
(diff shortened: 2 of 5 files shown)


org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -249,6 +259,20 @@ protected static String errorText(int error) {
...
+     * Safely initializes the session cache if not already initialized.
+     */
+    public static void initializeSessionCache(
+        int maxCacheEntries, long timeout, String directory)
+    {
+        if (sessionCacheInitialized.compareAndSet(false, true)) {
+            if (SSL.ConfigServerSessionIDCache(maxCacheEntries, timeout, timeout, directory) == SSL.SECFailure) {
+                String msg = "Unable to configure server session cache: ";
+                msg += errorText(PR.GetError());
+                throw new RuntimeException(msg);
```
edewata, line 272: Could we throw an `SSLException` instead and declare it in this method? The `JSSEngineReferenceImpl.beginHandshake()` should have declared an `SSLException` so we can just let it bubble up.

Generally I'd rather not create a `RuntimeException` as the initial exception unless we can't declare the exception in the method.

org/mozilla/jss/ssl/SSLServerSocket.java
```diff
@@ -266,9 +267,21 @@ public void close() throws IOException {
...
      *  is used: <code>/tmp</code> on Unix and <code>\\temp</code> on Windows.
      */
-    public static native void configServerSessionIDCache(int maxSidEntries,
-        int ssl2EntryTimeout, int ssl3EntryTimeout, String cacheFileDirectory)
-        throws SocketException;
+    public static void configServerSessionIDCache(int maxSidEntries,
+        int ssl2EntryTimeout, int ssl3EntryTimeout, String cacheFileDirectory) throws SocketException {
+        try {
+            JSSEngine.initializeSessionCache(maxSidEntries, ssl3EntryTimeout, cacheFileDirectory);
+        } catch (RuntimeException re) {
```
edewata, line 274: I think this should catch `SSLException`. See my other comment.

org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -176,6 +177,15 @@
...
     protected static HashMap<PK11Cert, SSLFDProxy> serverTemplates = new HashMap<PK11Cert, SSLFDProxy>();
 
+    /**
+     * Whether or not the session cache has been initialized already.
+     *
+     * A session cache must always be created in order to utilize a
+     * server-side JSSEngine. However, NSS isn't threadsafe when creating
+     * such a cache, so synchronize it within JSSEngine.
+     */
+    private final static AtomicBoolean sessionCacheInitialized = new AtomicBoolean();
```
edewata, line 187: The `AtomicBoolean` should work fine, but a regular `boolean sessionCacheInitialized` and a `synchronized initializeSessionCache()` probably would work as well. It's up to you, whichever simpler.

org/mozilla/jss/ssl/javax/JSSEngine.java
cipherboy: Sure, I think that's possible. I think originally `beginHandshake()` was called from `updateHandshakeState()` which in turn is always called from `getHandshakeStatus()` -- the latter declares no exceptions.

cipherboy: Yeah we just need to catch one location of `SSLException` and re-throw as `RuntimeException`, but otherwise we can handle that better now.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -196,31 +196,31 @@ private void createBufferFD() {
...
             ssl_fd.SetClientCert(cert);
             if (SSL.AttachClientCertCallback(ssl_fd) != SSL.SECSuccess) {
-                throw new RuntimeException("JSSEngine.init(): Unable to attach client certificate auth callback.");
+                throw new SSLException("Unable to attach client certificate auth callback.");
             }
         }
     }
 
-    private void initServer() {
+    private void initServer() throws SSLException, IllegalArgumentException {
```
edewata, line 223: `IllegalArgumentException` is a `RuntimeException` so it doesn't have to be declared.

org/mozilla/jss/ssl/SSLServerSocket.java
```diff
@@ -266,9 +269,21 @@ public void close() throws IOException {
...
-        int ssl2EntryTimeout, int ssl3EntryTimeout, String cacheFileDirectory)
-        throws SocketException;
+    public static void configServerSessionIDCache(int maxSidEntries,
+        int ssl2EntryTimeout, int ssl3EntryTimeout, String cacheFileDirectory) throws SocketException {
+        try {
+            JSSEngine.initializeSessionCache(maxSidEntries, ssl3EntryTimeout, cacheFileDirectory);
+        } catch (SSLException parent) {
+            // Because JSSEngine.initializeSessionCache is utilized during
+            // init of JSSengine implementations, it can only throw a
+            // RuntimeException. However, in the event it fails, we should
```
edewata, line 279: It should say `SSLException` instead.

cipherboy:
Okie doke, thanks @edewata!

PR #565, opened 2020-05-22
Introduce token-backed (KeyStore-backed) JSSKeyManager

cipherboy:
This introduces `KeyStore`-backed `JSSKeyManager`s. We:

 - Make `JSSKeyManager` an interface, allowing multiple implementations and making it easier to detect a compatible `KeyManager`,
 - Add the new `KeyStore`-backed `JSSTokenKeyManager` (this name comes from the fact that a `KeyStore` wraps an individual token). This falls back to `CryptoManager` when no `KeyStore` instance has been provided.
 - Since `KeyStore` is already tested, we just need to test the factory construction &c; we modify an existing test pass in `TestSSLEngine` to use the new initialization routines.


cipherboy:
@edewata -- Interestingly, the current `KeyStore` implementation doesn't restrict to the current token only. Do you think we should split the `KeyStore` implementation in two, one for `CryptoManager`-backed searches (like it currently does) and one imposing a hard requirement on it living on the specified token? We could probably initialize `KeyStore` differently, allowing us to replace `JSSCryptoKeyManager` and combine it into `JSSTokenKeyManager`, perhaps with a `JSSKeyManagerImpl` approach.

(That is, currently both implementations will currently find the same sets of keys/certs).

org/mozilla/jss/provider/javax/crypto/JSSTokenKeyManager.java
```diff
@@ -0,0 +1,131 @@
...
+            org.mozilla.jss.crypto.X509Certificate cert = getCertificate(alias);
+
+            org.mozilla.jss.crypto.X509Certificate[] chain = cm.buildCertificateChain(cert);
+            logger.debug("JSSKeyManager: cert chain:");
+
+            Collection<org.mozilla.jss.pkcs11.PK11Cert> list = new ArrayList<>();
+            for (org.mozilla.jss.crypto.X509Certificate c : chain) {
+                if (!(c instanceof PK11Cert)) {
+                    logger.error("JSSKeyManager: Can't cast to PK11Cert: " + c);
+                }
```
edewata, line 95: This is from the existing code, but it's probably not necessary since all classes implementing the `org.mozilla.jss.crypto.X509Certificate` are subclasses of `PK11Cert`.

cipherboy:
@edewata Thanks for the feedback. I've consolidated into one implementation. This makes it mostly a refactoring commit.

org/mozilla/jss/provider/javax/crypto/JSSTokenKeyManager.java
```diff
@@ -0,0 +1,138 @@
...
+    @Override
+    public String chooseServerAlias(String keyType, Principal[] issuers, Socket socket) {
+        logger.debug("JSSKeyManager: chooseServerAlias() - not implemented");
+        return null;
+    }
+
+    public org.mozilla.jss.crypto.X509Certificate getCertificate(String alias) {
+        try {
+            if (jks == null) {
+                CryptoManager cm = CryptoManager.getInstance();
```
edewata, line 76: This line can be removed. We can use the `cm` field.

org/mozilla/jss/provider/javax/crypto/JSSTokenKeyManager.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @Override
+    public PrivateKey getPrivateKey(String alias) {
+
+        logger.debug("JSSKeyManager: getPrivateKey(" + alias + ")");
+
+        try {
+            if (jks == null) {
+                CryptoManager cm = CryptoManager.getInstance();
```
edewata, line 122: This line can be removed. We can use the `cm` field.

org/mozilla/jss/provider/javax/crypto/JSSKeyManagerFactory.java
```diff
@@ -8,7 +8,26 @@
 import org.mozilla.jss.provider.java.security.*;
 
 public class JSSKeyManagerFactory extends KeyManagerFactorySpi {
+    private KeyStore internalStorage = null;
+    char[] password = null;
```
edewata, line 12: The `null` initialization is redundant :)

PR #570, opened 2020-05-27
HMAC initialization can leak ref counted symmetric key

ZuluForce:
In my company we've seen very bad memory leaks and one place we've tracked this down to is handling of the ref counted symmetric key in hmac initialization.

The problem stems from the assumption about `PK11_CopySymKeyForSigning` that if it returns a non-null pointer that it was copied. The actual implementation will return the exact same pointer if the key is already in the current slot and `CKA_SIGN` can be added to it.

```
    /* first just try to set this key up for signing */
    PK11_SETATTRS(&setTemplate, CKA_SIGN, &ckTrue, sizeof(ckTrue));
    pk11_EnterKeyMonitor(originalKey);
    crv = PK11_GETTAB(slot)->C_SetAttributeValue(originalKey->session,
                                                 originalKey->objectID, &setTemplate, 1);
    pk11_ExitKeyMonitor(originalKey);
    if (crv == CKR_OK) {
        return PK11_ReferenceSymKey(originalKey);
    }
```

The problem was reproduced by repeatedly calling `init` on an hmac with different keys that were created on the PK11 token. After applying these changes the memory usage is very flat although I suspect there's another leak because after ~92 million hmac `init` calls I'm seeing an increase in memory usage of ~100MB with no change in any java controlled memory pools (I fixed their sizes).

The change to `initHMAC` was just cleanup because it appeared unnecessary to call it when we haven't set the new key yet and the same native `initHMAC` method is called at the bottom.

### DIFF
--- a/org/mozilla/jss/pkcs11/PK11MessageDigest.c
+++ b/org/mozilla/jss/pkcs11/PK11MessageDigest.c
@@ -57,6 +57,7 @@ Java_org_mozilla_jss_pkcs11_PK11MessageDigest_initHMAC
     CK_MECHANISM_TYPE mech;
     SECItem param;
     jobject contextObj=NULL;
+    bool freeNewKey = true;
 
     mech = JSS_getPK11MechFromAlg(env, algObj);
     PR_ASSERT( mech != CKM_INVALID_MECHANISM ); /* we checked already in Java */
@@ -74,6 +75,7 @@ Java_org_mozilla_jss_pkcs11_PK11MessageDigest_initHMAC
 
     if( newKey == NULL ) {
         newKey = origKey;
+        freeNewKey = false;
     }
 
     param.data = NULL;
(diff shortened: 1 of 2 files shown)


cipherboy:
@jmagne Any chance you can take a look at this while I'm out?

jmagne:
Yes, I will take a look.

ZuluForce:
> what is the purpose of the removal of the call to "reset"?

Mentioned in the PR description the `reset()` was redundant and is only tangentially related to the memory leak. The implementation of `reset()` is this:

```
    public void reset() throws DigestException {
        if (!(this.alg instanceof HMACAlgorithm)) {
            this.digestProxy = initDigest(this.alg);
        } else if (this.hmacKey != null) {
            this.digestProxy = initHMAC(this.token, this.alg, this.hmacKey);
        } else {
            this.digestProxy = null;
        }
    }
```

The `reset()` call in `initHMAC` (java method) was happening before the `hmacKey` field was set so it's creating a new context with the _old_ key. Since `reset()` makes the same `initHMAC` (native) call as the java method was doing I removed the `reset()` and it ends up in the same state without the extra context being created.

ZuluForce:
Thanks for reviewing @cipherboy . I don't have the ability to merge so do so at your leisure. We have already applied this to an internal branch at my company so we're not blocked.

PR #575, opened 2020-06-11
bug1846565 - JSS: Extensions in CertificateExtensions class could get…

ladycfu:
… out of order and cause signature discrepency

This patch replaces Hashtable with LinkedHashMap to ensure order of
extensions.

https://bugzilla.redhat.com/show_bug.cgi?id=1846565

### DIFF
--- a/org/mozilla/jss/netscape/security/x509/CertificateExtensions.java
+++ b/org/mozilla/jss/netscape/security/x509/CertificateExtensions.java
@@ -27,8 +27,9 @@ import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.security.cert.CertificateException;
+import java.util.Collections;
 import java.util.Enumeration;
-import java.util.Hashtable;
+import java.util.LinkedHashMap;
 import java.util.Vector;
 
 import org.mozilla.jss.netscape.security.util.DerInputStream;
@@ -59,7 +60,7 @@ public class CertificateExtensions extends Vector<Extension>
      */
     public static final String NAME = "extensions";
 
-    private Hashtable<String, Extension> map;
+    private LinkedHashMap<String, Extension> map;
 
     // Parse the encoded extension
     public void parseExtension(Extension ext) throws IOException {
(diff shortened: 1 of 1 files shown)


org/mozilla/jss/netscape/security/x509/CertificateExtensions.java
```diff
@@ -27,8 +27,10 @@
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.security.cert.CertificateException;
+import java.util.Collections;
+import java.util.Collection;
```
SilleBille, line 31: Nitpick: this impor seems unused?

org/mozilla/jss/netscape/security/x509/CertificateExtensions.java
```diff
@@ -253,19 +255,19 @@ public void delete(String name) throws IOException {
...
      * attribute.
      */
     public Enumeration<Extension> getAttributes() {
-        return (map.elements());
+        return new Vector(map.values()).elements();
     }
 
     public Enumeration<String> getAttributeNames() {
-        return (map.keys());
+        return new Vector(map.keySet()).elements();
```
SilleBille, line 270: Any reason why we are returning a `Vector`? Can we reuse the getName() method? Like this?
````
return getNames();
````

File: org/mozilla/jss/netscape/security/x509/CertificateExtensions.java [hunk 3]
```diff
@@ -253,19 +255,19 @@ public void delete(String name) throws IOException {
...
+        return Collections.enumeration(map.keySet());
     }
 
     /**
      * Return an enumeration of names of attributes existing within this
      * attribute.
      */
     public Enumeration<Extension> getAttributes() {
-        return (map.elements());
+        return new Vector(map.values()).elements();
```
SilleBille, line 266: Why not use Collections class to convert to Enumeration?
````
return Collections.enumeration(map.values());
````

cipherboy:
+1 to Dinesh's comments then I'll approved :)

cipherboy:
Thanks @ladycfu! Merged :)

PR #599, opened 2020-07-13
Fixed TestSSLEngine

edewata:
The `SSLEngine.getSupportedProtocols()` on Fedora returns at least
2 protocols in FIPS mode, but on RHEL it returns only 1. The
`TestSSLEngine` has been changed such that it works with 1 protocol.

### DIFF
--- a/org/mozilla/jss/tests/TestSSLEngine.java
+++ b/org/mozilla/jss/tests/TestSSLEngine.java
@@ -1,17 +1,28 @@
 package org.mozilla.jss.tests;
 
-import java.lang.*;
-import java.nio.*;
-import java.net.*;
-import java.util.*;
-import java.security.*;
-import javax.net.ssl.*;
+import java.nio.ByteBuffer;
+import java.security.KeyStore;
+import java.util.ArrayList;
+import java.util.Arrays;
 
-import org.mozilla.jss.*;
-import org.mozilla.jss.nss.*;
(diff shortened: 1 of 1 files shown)


org/mozilla/jss/tests/TestSSLEngine.java
```diff
@@ -66,13 +77,16 @@ public static void testBasics(SSLContext ctx) throws Exception {
...
         // Tests {get,set}EnabledProtocols()
         String[] protocols = ssle.getSupportedProtocols();
-        assert protocols.length >= 2;
-        String secondProtocol = protocols[1];
-        String[] oneProtocols = new String[]{ secondProtocol };
+
+        // Fedora returns at least 2 supported protocols
+        // in FIPS mode, but RHEL returns only 1.
+        assert protocols.length >= 1;
+        String firstProtocol = protocols[1];
```
SilleBille, line 84: this still is the `secondProtocol` right?

ladycfu, line 84: protocols[0] maybe?

edewata: Fixed. It should be the first protocol.

edewata: Yes, fixed.

edewata:
The test passed on RHEL in FIPS mode. Thanks!

PR #602, opened 2020-07-22
Fix SSLEngine memory leaks

cipherboy:
This fixes two memory leaks in SSLEngine, and reduces memory pressure by handling certain other cases.

 1. Leaked `PRFDProxy` in `JSSEngineReferenceImpl.createBufferFD()`.
     When `SSL_ImportFD` is called, NSS merges the second argument into the return value, by way of pushing a NSPR on top of the argument. This layer is the SSL layer and the second argument is "consumed" by the SSL `PRFileDesc *`. Java doesn't know about this, so we need to manually clear (not close!!) the `PRFDProxy` instance and set it to `null` to avoid leaking it. (Otherwise, `NativeProxy` retains a reference to it, preventing the Java object from being garbage collected).
 2. Leaked `TokenProxy` in `JSSTokenKeyManager`.
    When the `CryptoManager` returns the new `PK11Cert` instance, it contains a `CertProxy` and a `TokenProxy`; the latter wasn't explicitly cleared/freed by `PK11Cert`. This meant `NativeProxy` still had a reference to it, preventing the `TokenProxy` Java object from being garbage collected. As the comment indicates, this isn't a memory leak in NSS, because Tokens (er, Slots) are long-term, reference-counted objects with the same pointer value. NSS presumably remains initialized throughout the lifetime of the program, so no leak really occurs until shutdown (if `CryptoManager.shutdown()` is actually called). 

Additionally, we:

 - Handle `null` return from `SSL_ImportFD` correctly,
 - Allow `SSLSessions` to "close", allowing us to release their resources. Note that invalidating a session is separate from closing a session. This gives us more control over when a session's certificates are freed, than leaving it to the garbage collector.
 - Ensure `JSSSocketChannel` closes the inbound side of the wrapped `JSSEngine`.
 - Ensure `JSSSocket` isn't used after being closed.

### DIFF
--- a/org/mozilla/jss/provider/javax/crypto/JSSTokenKeyManager.java
+++ b/org/mozilla/jss/provider/javax/crypto/JSSTokenKeyManager.java
@@ -118,8 +118,10 @@ public class JSSTokenKeyManager implements JSSKeyManager {
 
         try {
             if (jks == null) {
-                org.mozilla.jss.crypto.X509Certificate cert = cm.findCertByNickname(alias);
-                return cm.findPrivKeyByCert(cert);
+                try (PK11Cert cert = (PK11Cert) cm.findCertByNickname(alias)) {
+                    PrivateKey key = cm.findPrivKeyByCert(cert);
+                    return key;
+                }
             }
 
             return (PrivateKey) jks.getKey(alias, password);
--- a/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -195,6 +195,13 @@ public class JSSEngineReferenceImpl extends JSSEngine {
 
         // Initialize ssl_fd from the model Buffer-backed PRFileDesc.
         ssl_fd = SSL.ImportFD(model, fd);
+        if (ssl_fd == null) {
+            PR.Close(fd);
+            throw new SSLException("Error creating SSL socket on top of buffer-backed PRFileDesc.");
+        }
+
+        fd.clear();
+        fd = null;
         closed_fd = false;
 
         // Turn on SSL Alert Logging for the ssl_fd object.
@@ -1422,6 +1429,12 @@ public class JSSEngineReferenceImpl extends JSSEngine {
 
         // Then clean up the NSS state.
         cleanupSSLFD();
+
+        // Clean up the session.
+        if (session != null) {
+            session.close();
+            session = null;
+        }
     }
 
     private void cleanupLoggingSocket() {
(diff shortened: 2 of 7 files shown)


org/mozilla/jss/provider/javax/crypto/JSSTokenKeyManager.java
```diff
@@ -118,8 +118,10 @@ public PrivateKey getPrivateKey(String alias) {
 
         try {
             if (jks == null) {
-                org.mozilla.jss.crypto.X509Certificate cert = cm.findCertByNickname(alias);
-                return cm.findPrivKeyByCert(cert);
+                try (PK11Cert cert = (PK11Cert) cm.findCertByNickname(alias)) {
```
edewata, line 121: Does it mean we need to do this in PKI as well? Or will `PK11Cert.close()` get called automatically when it's garbage collected?

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -195,6 +195,12 @@ private void createBufferFD() throws SSLException {
 
         // Initialize ssl_fd from the model Buffer-backed PRFileDesc.
         ssl_fd = SSL.ImportFD(model, fd);
+        if (ssl_fd == null) {
+            throw new SSLException("Error creating SSL socket on top of buffer-backed PRFileDesc.");
+        }
+
+        fd.clear();
```
edewata, line 202: Do we need to call `fd.clear()` if `ssl_fd == null`?

org/mozilla/jss/provider/javax/crypto/JSSTokenKeyManager.java
cipherboy: No, this isn't strictly necessary. It'll get called automatically when it is garbage collected. However, because the scope is small here, it is good to use the try-with-resources pattern to ensure it gets closed and freed quickly. This means the GC pass will just clean up the destroyed Java objects (`PK11Cert`, `CertProxy`, and `TokenProxy` instances), and not need to call into the JNI layer to free NSS stuff.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
cipherboy: We should probably close it, not clear it in that case. I'll update it.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -195,6 +195,12 @@ private void createBufferFD() throws SSLException {
...
+        fd = null;
```
edewata, line 203: `fd` is a local variable, and usually it's not necessary to nullify a local variable, but do you think it's necessary here?

cipherboy: It isn't necessary to nullify a local variable to trigger the GC, I agree. 

However, it just ensures that if we were to use `fd` again later without thinking about it, we'd get a NPE rather than a segfault (if we tried to call into the JNI layer using `fd` again and something didn't adequately defend against a cleared pointer).

IMO this is preferable so I nulled it.

PR #605, opened 2020-07-23
More JSSEngine Memory Leak Fixes

cipherboy:
This introduces more memory fixes for JSSEngine:

 1. Fix missing destructor call in the `PR_Close` implementation for NSPR PR/IO layers. This meant the `BufferPRFD` layer leaked.
 2. Various small leak fixes in native (C) test case for `BufferPRFDSSL`. Not an issue in production.

~Additional fixes include~

 - ~Allow duplication of `PK11Cert`, `PK11PrivKey` objects, to take ownership of new copies.~
 - ~Make SSLEngine take ownership of certs, keys, and then release memory when cleanup occurs.~

----

I've decided not to include the latter two fixes. This makes the PR smaller and we can figure out how to deal with copying certs&keys later.

### DIFF
--- a/org/mozilla/jss/ssl/javax/BufferPRFD.c
+++ b/org/mozilla/jss/ssl/javax/BufferPRFD.c
@@ -94,6 +94,7 @@ static PRStatus PRBufferClose(PRFileDesc *fd)
     PR_ASSERT(fd->identity == buffer_layer_id);
     PR_ASSERT(fd->higher == NULL);
     PR_ASSERT(fd->lower == NULL);
+    fd->dtor(fd);
 
     return rv;
 }
--- a/org/mozilla/jss/tests/TestBufferPRFDSSL.c
+++ b/org/mozilla/jss/tests/TestBufferPRFDSSL.c
@@ -142,6 +142,8 @@ static CERTCertificate *get_cert(char *host)
         }
     }
 
+    CERT_DestroyCertList(clist);
+
     return NULL;
 }
 
@@ -180,7 +182,7 @@ static SECKEYPrivateKey *get_privkey(CERTCertificate *cert, char *password)
     return PK11_FindPrivateKeyFromCert(slot, cert, NULL);
 }
 
-static PRFileDesc *setup_nss_server(PRFileDesc *s_nspr, char *host, char *password, char *nickname)
+static PRFileDesc *setup_nss_server(PRFileDesc *s_nspr, char *host, char *password, char *nickname, SECKEYPrivateKey **priv_key)
 {
     /* Set up the server end of the SSL connection and find certificates. */
     /* Adapted from aforementioned Fedora developer guide and mod_nss. */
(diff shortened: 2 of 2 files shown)


org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -980,6 +980,15 @@ protected static SSLFDProxy getServerTemplate(PK11Cert cert, PK11PrivKey key) {
...
         SSLFDProxy fd = serverTemplates.get(cert);
         if (fd == null) {
+            // Here we clone the certificate. This certificate is functionally
+            // the same, but its scope is owned by the serverTemplates map.
+            // This allows the certificate to persist, even when the original
+            // key is closed by the owning JSSEngine instance. Note that we
+            // need not clone the key: ConfigServerCert will maintain its own
+            // copy of the key and cert, so we are free to let it get garbage
+            // collected.
+            PK11Cert cert_clone = cert.duplicate();
```
edewata, line 990: Should this be done later when it's actually needed, i.e. in line 1000?

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1435,21 +1439,45 @@ public void cleanup() {
...
+
+        cleanupKeys();
+    }
+
+    private void cleanupKeys() {
+        // Clean up the keys used in this transaction.
+        if (cert != null) {
+            try {
+                cert.close();
+            } catch (Exception e) {}
```
edewata, line 1451: Could we log a warning if there's an exception so at least we know if there's something wrong?

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1435,21 +1439,45 @@ public void cleanup() {
...
+
+            cert = null;
+        }
+
+        if (key != null) {
+            try {
+                key.close();
+            } catch (Exception e) {}
```
edewata, line 1459: Same here.

org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1435,21 +1439,45 @@ public void cleanup() {
...
+
+            key = null;
+        }
     }
 
     private void cleanupLoggingSocket() {
-        if (debug_port > 0) {
+        if (debug_port > 0 && s_socket != null && c_socket != null && ss_socket != null) {
```
edewata, line 1466: Should we create separate if-statements for each socket?
```
if (s_socket != null) {
   ... close s_socket ...
}
if (c_socket != null) {
   ... close c_socket ...
}
```

lib/jss.map
```diff
@@ -485,3 +485,10 @@ Java_org_mozilla_jss_nss_SECErrors_getUntrustedCert;
     local:
         *;
 };
+JSS_4.7.1 {
+    global:
+Java_org_mozilla_jss_pkcs11_PK11Cert_duplicate;
+Java_org_mozilla_jss_pkcs11_PK11PrivKey_duplicate;
```
edewata, line 491: It's not really an issue, but usually this type of operation is implemented either as `clone()` or copy constructor. Do we want to use that instead of `duplicate()`? I'll let you decide.

cipherboy:
@edewata I'm going to merge this without the top two changes with your comments. It broke and I need more time to figure out why. They're also not strictly memory leaks (since they'll eventually get GC'd anyways). Cleaner key management is something I'll work on more next release.

PR #607, opened 2020-07-24
Prevent usage of ssl_fd after closing

cipherboy:
In the previous commit, 76396ae47adf740aac0db38f143d959e5d6c39ec, by
calling the destructor on BufferPRFD layer, we finally clean it up
properly. However, this resulted in multiple calls to `closeInbound` or
`closeOutbound` resulting in a use-after-free.


cipherboy:
This looks better. BadSSL failing test will take a little longer to track down.

PR #612, opened 2020-07-28
NativeProxy: Use Weak References 

cipherboy:
This fixes a memory leak caused by `NativeProxy` maintaining strong references to all pointer objects. By switching them to `WeakReferences`, the registry is ignored in the GC reference count, and we can again allow items to be freed. 

Moving to `WeakReference` wasn't immediately possible due to a bug in SSLEngine: SSLEngine's server template didn't clear the underlying `PRFileDesc` pointer, meaning that -- when GC was again possible -- it would get freed, invaliding the model unintentionally. This resulted in SSLEngine tests failing. So, make `SSL.ImportFD` clear the consumed pointer.

### DIFF
--- a/lib/jss.map
+++ b/lib/jss.map
@@ -485,3 +485,9 @@ Java_org_mozilla_jss_nss_SECErrors_getUntrustedCert;
     local:
         *;
 };
+JSS_4.7.1 {
+    global:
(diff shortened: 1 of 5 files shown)


cipherboy:
This was better -- but we're still chasing one more leak.

PR #620, opened 2020-08-11
Allow optional MacData field

cipherboy:
When generating on a HSM in FIPS mode, SHA-1 HMAC isn't allowed as a PBE
algorithm. For the use of certificate chain export, MAC data isn't
necessary, as the PKCS#12 is temporary and will be written to a PKCS#7
PEM chain, stored in the NSS DB directory. Thus, `MacData` isn't strictly
necessary.


cipherboy:
Closing this in favor of dogtagpki/pki#516.

Issue #621, opened 2020-08-13
port from commons-lang to commons-lang3


decathorpe:
apache commons-lang 2.x releases are no longer compatible with Java 11 (due to usage of `assert` and `enum` as identifiers), and fedora packages are all getting ported to commons-lang3.

A search of this repository shows no usages of APIs that were removed with the 3.x release of commons-lang, so the port should be straightforward.

- replace BuildRequires and Requires in jss.spec: `s/apache-commons-lang/apache-commons-lang3/`
- replace import statements in Java files: `s/org\.apache\.commons\.lang\./org.apache.commons.lang3./`
- replace package in [jss/tools/Dockerfiles/fedora_sandbox][dockerfile]: `s/apache-commons-lang/apache-commons-lang3/`
- replace library in [cmake/JSSConfig.cmake][cmake]: `s/apache-commons-lang commons-lang/apache-commons-lang3 commons-lang3/`

[dockerfile]: https://github.com/dogtagpki/jss/blob/de4d9d7/tools/Dockerfiles/fedora_sandbox#L7
[cmake]: https://github.com/dogtagpki/jss/blob/6d78828/cmake/JSSConfig.cmake#L214

PR #622, opened 2020-08-17
Port from Apache Commons Lang 2 -> 3

cipherboy:
Credits: Fabio "decathorpe" Valentini

Resolves: #621


cipherboy:
Note that this is blocked on it merging in Dogtag PKI: https://github.com/dogtagpki/pki/pull/520 

That in turn is blocked because we don't have a new enough package in F31/F32. :/

PR #601, opened 2020-07-20
Use JSS-provided TrustManagers by default

cipherboy:
When SSLContext is called with no TrustManagers, we end up defaulting to
the default internal NSS validation logic. This results in connections
which fail to verify when no hostname is provided, because NSS strictly
requires hostnames to be present and matching. Because the JDK otherwise
requires hostnames to validated (by e.g., a HostnameVerifier), we
shouldn't use the NSS handler by default.

`Signed-off-by: Alexander Scheel <ascheel@redhat.com>`

PR #625, opened 2020-08-31
Fix hostname verification and improve JCA compatibility

cipherboy:
This PR improves JCA compatibility by bypassing NSS's hostname verification on client `SSLEngine` connections, to mirror the behavior of SunJSSE.

We use the approach suggested by Bob Relyea, which is to hook the bad cert handler and check for a specific error. Because hostname verification is guaranteed to be the last thing NSS does (regardless of whether or not we use NSS's default handler or libpkix verification), this should be safe. This has the added benefit of also working with our custom cert authentication callbacks. 

Eventually, we'll expose this method to all JSSEngine instances, and allow the caller a chance to override other errors if they'd like, in an asynchronous manner.

### DIFF
--- a/org/mozilla/jss/nss/SSL.java
+++ b/org/mozilla/jss/nss/SSL.java
@@ -472,8 +472,36 @@ public class SSL {
     public static native int ConfigSyncTrustManagerCertAuthCallback(SSLFDProxy fd);
 
     /**
+     * Use an asynchronous bad certificate handler which allows us to approve
+     * rejected certificates. This allows us to bypass the hostname check
+     * failure caused by the Java socket having no knowledge of the hostname
+     * we use for certificate validation; no HostnameVerifier is passed in.
+     * As a result, NSS has no value for the hostname and validation will fail.
+     *
+     * Note: This does NOT work for server-side connections.
+     *
+     * See also: SSL_BadCertHook in /usr/include/nss3/ssl.h and
+     *           JSSL_SSLFDAsyncBadCertCallback in jss/nss/SSLFDProxy.c
+     */
+    public static native int ConfigAsyncBadCertCallback(SSLFDProxy fd);
+
+    /**
+     * Use a synchronous bad certificate handler which allows us to approve
+     * rejected certificates. This allows us to bypass the hostname check
+     * failure caused by the Java socket having no knowledge of the hostname
+     * we use for certificate validation; no HostnameVerifier is passed in.
+     * As a result, NSS has no value for the hostname and validation will fail.
+     *
+     * See also: SSL_BadCertHook in /usr/include/nss3/ssl.h and
+     *           JSSL_SSLFDSyncBadCertCallback in jss/nss/SSLFDProxy.c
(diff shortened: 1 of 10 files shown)


cipherboy:
@SilleBille Do you still have your test code from #601 by any chance? I think this will work better. :)

org/mozilla/jss/nss/SSLFDProxy.java
```diff
@@ -19,9 +19,12 @@
     public int outboundOffset;
 
     public boolean needCertValidation;
+    public boolean needBadCertValidation;
+    public int badCertError;
     public boolean handshakeComplete;
 
-    public CertAuthHandler handler;
+    public CertAuthHandler certAuthHandler;
+    public BadCertHandler badCertHandler;
```
edewata, line 27: Just a note, the old public `handler` field is no longer available, so this is not a backward compatible change, but it probably doesn't matter assuming it's only used internally.

org/mozilla/jss/nss/SSL.java
```diff
@@ -471,10 +471,38 @@ public static int EnableAlertLogging(SSLFDProxy fd) {
...
+     * failure caused by the Java socket having no knowledge of the hostname
+     * we use for certificate validation; no HostnameVerifier is passed in.
+     * As a result, NSS has no value for the hostname and validation will fail.
+     *
+     * Note: This does NOT work for server-side connections.
+     *
+     * See also: SSL_BadCertHook in /usr/include/nss3/ssl.h and
+     *           JSSL_SSLFDAsyncBadCertCallback in jss/nss/SSLFDProxy.c
+     */
+    public static native int ConfigAsyncBadCertCallback(SSLFDProxy fd);
```
edewata, line 486: When should we use `ConfigAsyncBadCertCallback` as opposed to `ConfigSyncBadCertCallback`?

cipherboy: Synchronous `BadCertCallback` is good when the validation is short. It also is the only method that works on both client-side and server-side connections.

Asynchronous `BadCertCallback` is good when the validation takes a long time and/or invokes other network requests. It works **only** on client-side connections (which this is).

Because ours is short, I went with the former.

The same reasoning relates to `CertAuthHandler`, which is already present in JSSEngine. With user-supplied `TrustManager`s, we check whether or not we can be async and prefer that (issuing a `DelegatedTask` to actually run this validation).

In the future, since we might include other BadCert callback hooks, we might use async, so I added this too.

org/mozilla/jss/nss/SSLFDProxy.java
cipherboy: Yeah, `nss/SSLFDProxy` is used mostly internally. It has to be public as earlier discussed.

cipherboy:
Thanks!

SilleBille:
@cipherboy I have my CRMFPopclient code in my personal repo: https://github.com/SilleBille/pki/tree/fix_crmfpopclient

PR #627, opened 2020-09-01
[v4.7.3] Disable Fallback SCSV and add PHA to legacy SSLSocket

cipherboy:
This PR fixes two additional issues:

 - We disable Fallback SCSV. While used by browsers, it isn't present in the cipher suites negotiated by SunJSSE by default. Since we're striving to be compatible with that, drop SCSV.
 - Add PHA to SSLSocket. This addresses the failure in https://github.com/dogtagpki/pki/pull/537 and https://github.com/dogtagpki/pki/pull/543, with one additional patch.

@edewata / @jmagne -- should we make PHA default in the legacy SSLSocket? We added TLSv1.3 support to it a while ago, but TLSv1.3's behavior differs from TLSV1.2's behavior, probably why we had issues with SCSV and TLSv1.2 pinned as maximum in Dogtag PKI. If we enable PHA, TLSv1.2 and TLSv1.3 will mostly be compatible.

If so, we could avoid the need for a strict dependency on a newer JSS version.

### DIFF
--- a/org/mozilla/jss/ssl/SocketBase.java
+++ b/org/mozilla/jss/ssl/SocketBase.java
@@ -101,6 +101,7 @@ class SocketBase {
     /* ssl/sslt.h */
     static final int SSL_Variant_Stream = 33;
     static final int SSL_Variant_Datagram = 34;
+    static final int SSL_ENABLE_POST_HANDSHAKE_AUTH = 36;
 
     static final int SSL_AF_INET = 50;
     static final int SSL_AF_INET6 = 51;
@@ -173,6 +174,10 @@ class SocketBase {
 
     void enableV2CompatibleHello(boolean enable) throws SocketException {
         setSSLOption(SSL_V2_COMPATIBLE_HELLO, enable);
+    }
+
+    void enablePostHandshakeAuth(boolean enable) throws SocketException {
+        setSSLOption(SSL_ENABLE_POST_HANDSHAKE_AUTH, enable);
     }
 
     void setSSLOption(int option, boolean on)
--- a/org/mozilla/jss/ssl/common.c
+++ b/org/mozilla/jss/ssl/common.c
@@ -422,6 +422,7 @@ PRInt32 JSSL_enums[] = {
     ssl_variant_stream,           /* 33 */      /* sslt.h */
     ssl_variant_datagram,         /* 34 */      /* sslt.h */
     SSL_LIBRARY_VERSION_TLS_1_3,  /* 35 */      /* sslproto.h */
+    SSL_ENABLE_POST_HANDSHAKE_AUTH, /* 36 */      /* ssl.h */
     0
 };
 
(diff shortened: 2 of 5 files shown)


edewata:
Sory, I'm not familiar with this, so I'd defer to @jmagne or @ladycfu.

org/mozilla/jss/ssl/SocketBase.java
```diff
@@ -101,9 +101,11 @@ void setProxy(SocketProxy sockProxy) {
     /* ssl/sslt.h */
     static final int SSL_Variant_Stream = 33;
     static final int SSL_Variant_Datagram = 34;
+    static final int SSL_ENABLE_POST_HANDSHAKE_AUTH = 36;
 
     static final int SSL_AF_INET = 50;
     static final int SSL_AF_INET6 = 51;
+    /* TODO right place? */
```
cipherboy, line 108: This is, can remove this comment.

org/mozilla/jss/ssl/common.c
```diff
@@ -422,6 +422,7 @@ PRInt32 JSSL_enums[] = {
     ssl_variant_stream,           /* 33 */      /* sslt.h */
     ssl_variant_datagram,         /* 34 */      /* sslt.h */
     SSL_LIBRARY_VERSION_TLS_1_3,  /* 35 */      /* sslproto.h */
+    SSL_ENABLE_POST_HANDSHAKE_AUTH,  /* 36 */      /* sslproto.h */
```
cipherboy, line 425: Clean up diff.

org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -945,7 +945,6 @@ public boolean isOutboundDone() {
         // TLS < 1.3.
         result.put(SSL.ENABLE_RENEGOTIATION, SSL.RENEGOTIATE_REQUIRES_XTN);
         result.put(SSL.REQUIRE_SAFE_NEGOTIATION, 1);
-        result.put(SSL.ENABLE_FALLBACK_SCSV, 1);
```
ladycfu, line 948: would it makes sense to just comment this out and give explanation, so later someone who wishes to add this would know why?

cipherboy:
OK, thanks @ladycfu! I'll address it with better commit messages in `master` branch PR tomorrow.

org/mozilla/jss/ssl/javax/JSSEngine.java
ladycfu: @cipherboy I"m not sure if I've seen any response to this ;-)

cipherboy: Ah sorry, I responded at the bottom: 

https://github.com/dogtagpki/jss/pull/627#issuecomment-685190217

The version in `master` has a better commit message and description: 

https://github.com/dogtagpki/jss/commit/d0c8b2453dff73d412ba7a4ffa274c16ea57a415

```java

        // Don't enable FALLBACK_SCSV support; this is incompatible with other
        // JCA providers and makes transitioning TLS versions difficult.
        // result.put(SSL.ENABLE_FALLBACK_SCSV, 1);

```

ladycfu: Ah, I didn't know you were responding to this one, as I was suggesting you to comment out that line instead of removing it in the code and add comment in the code.

ladycfu: it's not that critical though. It's just a suggestion.

cipherboy: Yeah it was commented out (as suggested here) in `master` for posterity. :-)

ladycfu: Ah, sorry, my bad.  I didn't see. ;-) Thanks!!

PR #633, opened 2020-09-10
Add documentation about CryptoManager/JSSProvider

cipherboy:

Per IRC discussions, likely an interesting read. :-) LMK if you still have questions or have suggestions for updates or more material.

### DIFF
--- a/docs/usage/cryptomanager.md
+++ b/docs/usage/cryptomanager.md
@@ -1,0 +1,156 @@
+# `CryptoManager`
+
+## Design
+
+### `CryptoManager`, `JSSProvider`, and `JSSLoader` interactions
+
+`CryptoManager` is the central singleton of JSS. It controls access to the
+NSS database and an instance (available to developers via the
+`CryptoManager.getInstance()` call) signals that both JSS and NSS are properly
+initialized. The existing relationship between `CryptoManager` and
+`JSSProvider` is that a single `CryptoManager` instance has a single
+`JSSProvider` instance, and moreso that the reverse is also true: the
+`JSSProvider` instance has a single `CryptoManager` instance. Currently the
+code assumes that there is only ever one `CryptoManager` instance, making
+both singletons.
+
+_Aside:_
+
+> Future work could be done to enable either multiple `CryptoManager` instances
+> with separate NSS DBs, or to enable multiple `JSSProvider` instances to
+> reference one or more `CryptoManager` instances during normal operation.
+> This currently will require significant restructuring as many internal
+> provider methods (such as `Cipher`, `Signature` and others) get the global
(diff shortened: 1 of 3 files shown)


docs/usage/cryptomanager.md
```diff
@@ -0,0 +1,156 @@
...
+
+```java
+import java.security.*;
+
+import org.mozilla.jss.CryptoManager;
+
+public class Example {
+  public static void main(String[] args) throws Exception {
+    Provider p = Security.getProvider("Mozilla-JSS");
+    CryptoManager cm = CryptoManager.getInstance();
```
edewata, line 89: I'm thinking eventually we might do something like this:
```
JSSProvider provider = (JSSProvider) Security.getProvider("Mozilla-JSS");
CryptoManager cm = provider.getCryptoManager();
```
and deprecate/drop `CryptoManager.getInstance()`. Or maybe even replace `CryptoManager` with `JSSProvider`.

edewata (approved):
Thanks for the explanation! You might want to add a reference to this doc from the relevant code.

docs/usage/cryptomanager.md
cipherboy: I think neither of these are quite right. If I were to make these changes (they're big, time-consuming and not immediately clear what the benefit is), I'd probably:

 - Make `JSSProvider` load out of a config file, keep state internally.
 - Make `CryptoManager` an internal implementation detail and have it correlate to exactly one NSS DB. It will back-end the `KeyStoreSpi` and `TrustStoreSpi` code. Instances will be singleton-per-NSS-DB and stored/cached on the associated `JSSProvider` instance for use throughout the JCA.
 - Drop all non-JCA interfaces when duplicated by JCA interfaces.
 - Make JCA interfaces both compatible with SunJSSE/... and more compliant. I'd like to see keys &c interop when possible and other improvements.

But, who knows. :-)

cipherboy: Most of the magic also occurs in the "more compliant" and "compatible" steps -- which the rest isn't necessary for and could be done now. The benefit of the restructuring is that it would let us do the .p12/jks stuff, which isn't as likely to work well otherwise as you can't control where keys come from as easily. So I think it is unlikely this restructuring is likely to occur now.

cipherboy:
OK, mentioned the doc in CryptoManager and JSSLoader. Thanks! Will merge when CI is green again.

docs/usage/cryptomanager.md
edewata: +1 to all of the above. There might be no major immediate benefit, but I think whenever we make modifications we should do it with these goals in mind. That way we won't accumulate a huge technical debt when we have an actual requirement that depends on this in the future.

PR #634, opened 2020-09-11
Release JSS v4.7.3

cipherboy:

### DIFF
--- a/jss.spec
+++ b/jss.spec
@@ -6,7 +6,7 @@ Summary:        Java Security Services (JSS)
 URL:            http://www.dogtagpki.org/wiki/JSS
 License:        MPLv1.1 or GPLv2+ or LGPLv2+
 
-Version:        4.7.2
+Version:        4.7.3
 Release:        1%{?_timestamp}%{?_commit_id}%{?dist}
 #global         _phase -a1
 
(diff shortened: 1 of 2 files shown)


jss.spec
```diff
@@ -108,13 +108,26 @@ export CFLAGS
 # Check if we're in FIPS mode
 modutil -dbdir /etc/pki/nssdb -chkfips true | grep -q enabled && export FIPS_ENABLED=1
 
+# RHEL's CMake doesn't support -B flag.
+%if 0%{?rhel}
+%{__mkdir_p} %{_vpath_builddir}
+cd %{_vpath_builddir}
+%endif
```
SilleBille, line 115: This seems to be opposite to one in dogtagpki/pki. That is, we ask to use `_vpath_builddir` in Fedora and `build` in RHEL. 

Here is the code from PKI repo: https://github.com/dogtagpki/pki/pull/554/files#diff-309669ccd4cb6f6fdf531483d210a34dR833-R836

SilleBille (commented):
One question when comparing it with dogtagpki/pki repo

jss.spec
cipherboy: Endi wrote the one in PKI and kept it at `build/` in RHEL. In JSS, since `%{_vpath_builddir}` is defined on RHEL, I kept it, and only stubbed around the missing CMake functionality.

SilleBille (approved):

jss.spec
cipherboy: (in JSS it is harder to keep `build/` vs `%{_vpath_builddir}` since we use the latter in the `%install` section.)

PR #635, opened 2020-09-16
Bump JSS version to v4.8.0 alpha 1

cipherboy:

### DIFF
--- a/jss.spec
+++ b/jss.spec
@@ -6,9 +6,9 @@ Summary:        Java Security Services (JSS)
 URL:            http://www.dogtagpki.org/wiki/JSS
 License:        MPLv1.1 or GPLv2+ or LGPLv2+
 
-Version:        4.7.2
-Release:        1%{?_timestamp}%{?_commit_id}%{?dist}
-#global         _phase -a1
+Version:        4.8.0
+Release:        0.1%{?_timestamp}%{?_commit_id}%{?dist}
+%global         _phase -a1
 
 # To generate the source tarball:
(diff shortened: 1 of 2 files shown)


cipherboy:
@edewata Did I do this one correctly? :-)

edewata (approved):
I never changed `jss_config_version` but it looks correct.

The spec file looks correct, but if you want you can insert `alpha1` like this:
https://github.com/dogtagpki/pki/blob/master/pki.spec#L16
This is to clarify the state of the code to the users.

jss.spec
```diff
@@ -6,9 +6,9 @@ Summary:        Java Security Services (JSS)
 URL:            http://www.dogtagpki.org/wiki/JSS
 License:        MPLv1.1 or GPLv2+ or LGPLv2+
 
-Version:        4.7.2
-Release:        1%{?_timestamp}%{?_commit_id}%{?dist}
-#global         _phase -a1
+Version:        4.8.0
+Release:        0.1%{?_timestamp}%{?_commit_id}%{?dist}
+global         _phase -a1
```
edewata, line 11: Sorry, missed this one. You need to change it to `%global`:
```
%global           _phase -alpha1
```
The `Release` number is used in RPM, the `_phase` is used in the source tarball. Perhaps they can be combined somehow.

cipherboy: Yep, thanks @edewata!

cipherboy:
We can add `alpha1` when we get to an alpha release of JSS :-)

PR #636, opened 2020-09-30
Mavenize JSS

cipherboy:
This will be a large pull request.

We're attempting to mavenize JSS here. This is challenging mostly due to our native components:

 - [x] We need to build a native component. This seems to require we depend on `native-maven-plugin`. (Currently this is stubbed and doesn't work)
    - Should we use this plugin?
    - Should we use a Makefile instead, that we could share across CMake and this?
 - [ ] However, it'd be nice to support our existing `sandbox/` builds with a checked-out NSS directory.
 - [ ] We need to make maven builds reproducible. 
 - [ ] We need to figure out how to template some configuration values like we do in CMake.
 - [ ] We need to pass flags from CFLAGS into our native build parameters.
 - [ ] We need to figure out how to replicate our test environment under maven.
    - Should we use an external testing environment? Something like Python?
    - Should we try to build Maven artifacts into the existing CMake paths so we can simply compile and run tests from CMake?

And probably other things. :-)

---

@edewata do you know how to make the native component work? I'd expect (because of the issues above, including not specifying NSS search paths at all!) it to fail miserably. But currently the build passes, but doesn't generate any output.


pom.xml
```diff
@@ -0,0 +1,79 @@
...
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+            <version>2.2.12</version>
+        </dependency>
+
+        <dependency>
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+            <version>2.2.12</version>
+        </dependency>
```
edewata, line 39: Duplicate declaration.

pom.xml
```diff
@@ -0,0 +1,79 @@
...
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
```
edewata, line 54: This plugin needs a version number:
```
<version>3.8.1</version>
```

cipherboy: Doesn't that strictly tie it to a version number? That adds a lot of work (manually updating version numbers...) and I'd prefer not to... Could we add a minimum instead somehow?

pom.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.mozilla</groupId>
+    <artifactId>jss</artifactId>
+    <version>4.8.0-SNAPSHOT</version>
+
```
edewata, line 9: Apparently we need this to start compiling native sources:
```
<packaging>so</packaging>
```

pom.xml
edewata: I'm not too familiar with this either, but apparently that specifies the minimum version number:
https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html

Here is the example that they provided:
http://maven.apache.org/plugins/maven-compiler-plugin/usage.html
And that is the latest available version in Maven Central.

pom.xml
```diff
@@ -0,0 +1,73 @@
...
+                <artifactId>native-maven-plugin</artifactId>
+                <version>1.0-alpha-11</version>
+                <extensions>true</extensions>
+                <configuration>
+                    <compilerProvider>generic-classic</compilerProvider>
+                    <compilerExecutable>gcc</compilerExecutable>
+                    <sources>
+                        <source>
+                            <directory>${basedir}/src/main/native</directory>
+                        </source>
```
edewata, line 66: Probably needs an `<includes>` as well.

pom.xml
cipherboy: Ah yuck. So we've gotta split JSS into two maven projects and build them both separately. Gotcha.

cipherboy:
@edewata \o/ I think I did it!

native/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.mozilla</groupId>
+        <artifactId>jss-parent</artifactId>
+        <version>4.8.0-SNAPSHOT</version>
+    </parent>
+
+    <groupId>org.mozilla</groupId>
+    <artifactId>jss-native</artifactId>
+    <packaging>so</packaging>
```
edewata, line 14: How will the dependency be defined in other projects that use JSS?
Do we need both of these, or just one?
```
<dependency>
    <groupId>org.mozilla</groupId>
    <artifactId>jss</artifactId>
    <version>4.8.0</version>
</dependency>

<dependency>
    <groupId>org.mozilla</groupId>
    <artifactId>jss-native</artifactId>
    <version>4.8.0</version>
    <type>so</type>
</dependency>
```
Or is it possible to have something like this?
```
<dependency>
    <groupId>org.mozilla</groupId>
    <artifactId>jss</artifactId>
    <version>4.8.0</version>
    <type>so</type>
</dependency>
```

edewata (commented):
Nice! Please see my comments below.

native/pom.xml
cipherboy: Don't you just grab:

```
<dependency>
    <groupId>org.mozilla</groupId>
    <artifactId>jss-parent</artifactId>
    <version>4.8.0</version>
</dependency>
```

(from `pom.xml` in the root of the git tree)

That has two modules, and presumably, wouldn't both get pulled in?

edewata: IIUC we need to specify the `<type>` too. Some examples:
https://www.mojohaus.org/maven-native/native-maven-plugin/usage.html
https://stackoverflow.com/questions/3064045/using-native-dependencies-inside-maven
Since there are two modules here with different types, I'm not sure.

Regardless, should we call the parent module `jss` and the submodules maybe `jss-java` and `jss-native`?

BTW, I'm getting this error (the regular cmake build works fine):
```
$ mvn package
...
Failed tests:   testControlCharacters(org.mozilla.jss.tests.DirStrConverterTest): arrays first differed at element [0]; expected:<20> but was:<12>
  testNullCharacters(org.mozilla.jss.tests.DirStrConverterTest): arrays first differed at element [0]; expected:<20> but was:<12>
  testEmptyString(org.mozilla.jss.tests.DirStrConverterTest): arrays first differed at element [0]; expected:<19> but was:<12>
  testMultibyteCharacters(org.mozilla.jss.tests.DirStrConverterTest): array lengths differed, expected.length=14 actual.length=11; arrays first differed at element [0]; expected:<28> but was:<12>
  testPrintableCharacters(org.mozilla.jss.tests.DirStrConverterTest): arrays first differed at element [0]; expected:<19> but was:<12>

Tests in error: 
  testConstructorDER(org.mozilla.jss.tests.GenericASN1ExtensionTest)
```

cipherboy: Maven isn't enabled for testing. I wonder if there's a way to disable it completely?

Tests won't pass under maven because:

 1. This test isn't enabled in the cmake test suite.
 2. Tests generally require java.security to exist for JSS to be loaded.
 3. Tests depend on each other in a way that isn't easy to express in JUnit files/dependencies.
 4. Tests have side-effects.
 5. Not everything in a Java or JUnit tests. There's a lot of custom Java code but also some nss-utils calls as well. 

So it isn't surprising that this is failing.

edewata: We can skip tests with `-DskipTests`, but now I got a different error during install:
```
$ mvn install -DskipTests
...
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-install-plugin:3.0.0-M1:install (default-install) on project jss-native: NoFileAssignedException: The packaging for this project did not assign a file to the build artifact -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :jss-native
```

edewata: Found this example (not sure if it will help though):
https://github.com/StephanU/native-maven-plugin-example

jss/pom.xml
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.mozilla</groupId>
```
edewata, line 7: We might want to use `org.dogtagpki` instead of `org.mozilla` to avoid conflicts in case Mozilla publishes something in the future. I think the `<groupId>` is more like a vendor ID. It doesn't have to match the actual Java package names.

edewata:
This PR has been rebased and merged in PR #869. Thanks @cipherboy!

PR #640, opened 2020-10-06
Add ALPN support in SSLSocket

frasertweedale:
This is a prerequisite for ACME tls-alpn-01 challenge verification.
See https://gist.github.com/frasertweedale/dfac386579cb2d1fe892f39d22c2975d
for usage / testbed.

```
ab29dd6c (Fraser Tweedale, 7 minutes ago)
   alpn: support via new SSLSocket constructor

   Add a new SSLSocket constructor that allows the caller to specify ALPN
   protocols.

555134db (Fraser Tweedale, 2 hours ago)
   alpn: add SSLSocket.getNegotiatedProtocol()

   Add a method to retrieve the identifier of the negotiated protocol, if
   available.

465ac1d9 (Fraser Tweedale, 4 days ago)
   alpn: format and supply ALPN data if user provides it

   Update the private SSLSocket constructor to accept a String[] of ALPN
   protocol names in preferred order.  When given, it will prepare the ALPN
   extension data in the required format and pass it along to the
   socketConnect method.

   The change to publicly expose this capability will occur in a subsequent
   commit.

6f5a0cf8 (Fraser Tweedale, 4 days ago)
   alpn: add support in native socketConnect method

   Update the private native socketConnect method to receive optional ALPN
   data to include in handshake.  The value is performatted per 
   SSL_SetNextProtoNego().

   Support for library users to supply the ALPN data will follow in subsequent
   commits.
```

### DIFF
--- a/org/mozilla/jss/ssl/SSLSocket.c
+++ b/org/mozilla/jss/ssl/SSLSocket.c
@@ -572,6 +572,29 @@ Java_org_mozilla_jss_ssl_SSLSocket_getPort(JNIEnv *env,
     }
 }
 
+JNIEXPORT jbyteArray JNICALL
+Java_org_mozilla_jss_ssl_SSLSocket_getNegotiatedProtocol(JNIEnv *env, jobject self)
+{
+    JSSL_SocketData *sock;
+    SECStatus status;
+    SSLNextProtoState npState;
+    unsigned char buf[255];
+    unsigned int bufLen = 0;
+
+    if (JSSL_getSockData(env, self, &sock) != PR_SUCCESS || sock == NULL) {
+        /* exception was thrown */
+        EXCEPTION_CHECK(env, sock);
+        return NULL;
+    }
+
+    status = SSL_GetNextProto(sock->fd, &npState, buf, &bufLen, sizeof(buf));
+    if (status == SECSuccess && bufLen > 0 && bufLen <= sizeof(buf)) {
+        return JSS_ToByteArray(env, buf, bufLen);
+    } else {
+        return NULL;
+    }
+}
+
 JNIEXPORT void JNICALL
 Java_org_mozilla_jss_ssl_SSLSocket_socketConnect
     (JNIEnv *env, jobject self, jbyteArray addrBA, jstring hostname, jint port, jbyteArray alpn)
--- a/org/mozilla/jss/ssl/SSLSocket.c
+++ b/org/mozilla/jss/ssl/SSLSocket.c
@@ -574,7 +574,7 @@ Java_org_mozilla_jss_ssl_SSLSocket_getPort(JNIEnv *env,
 
 JNIEXPORT void JNICALL
 Java_org_mozilla_jss_ssl_SSLSocket_socketConnect
-    (JNIEnv *env, jobject self, jbyteArray addrBA, jstring hostname, jint port)
+    (JNIEnv *env, jobject self, jbyteArray addrBA, jstring hostname, jint port, jbyteArray alpn)
 {
     JSSL_SocketData *sock;
     PRNetAddr addr;
@@ -583,6 +583,9 @@ Java_org_mozilla_jss_ssl_SSLSocket_socketConnect
     PRStatus status;
     int stat;
     const char *hostnameStr=NULL;
+
+    unsigned char *alpnData = NULL;
+    int alpnLength = 0;
 
     jmethodID supportsIPV6ID;
     jclass socketBaseClass;
@@ -630,6 +633,14 @@ Java_org_mozilla_jss_ssl_SSLSocket_socketConnect
         goto finish;
     }
 
+    /* Set ALPN extension */
+    if (alpn != NULL) {
+        JSS_RefByteArray(env, alpn, (jbyte**)&alpnData, &alpnLength);
+        if (alpnLength > 0) {
+            SSL_SetNextProtoNego(sock->fd, alpnData, alpnLength);
+        }
+    }
+
     if(addrBALen != 4 && addrBALen != 16) {
         JSSL_throwSSLSocketException(env, "Invalid address in connect!");
         goto finish;
@@ -672,6 +683,7 @@ finish:
 
     JSS_DerefJString(env, hostname, hostnameStr);
     JSS_DerefByteArray(env, addrBA, addrBAelems, JNI_ABORT);
+    JSS_DerefByteArray(env, alpn, alpnData, JNI_ABORT);
 }
 
 JNIEXPORT jobject JNICALL
--- a/org/mozilla/jss/ssl/SSLSocket.java
+++ b/org/mozilla/jss/ssl/SSLSocket.java
@@ -644,6 +644,14 @@ public class SSLSocket extends java.net.Socket {
         resetHandshake();
     }
 
+    /* Get result of ALPN negotiation.
+     *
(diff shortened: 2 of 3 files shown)


frasertweedale:
@cipherboy ahaha whups, I did not even see https://github.com/dogtagpki/jss/pull/526 :S

frasertweedale:
@cipherboy ah, looks like #526 is about the server side anyway, whereas this PR is for the client side and is mostly orthogonal.

org/mozilla/jss/ssl/SSLSocket.java
```diff
@@ -547,16 +548,76 @@ public SSLSocket(InetAddress address, int port,
...
+     *      added.  Each protocol ID must have a length between 1 and 255
+     *      bytes.
+     * @param certApprovalCallback A callback that can be used to override
+     *      approval of the peer's certificate.
+     * @param clientCertSelectionCallback A callback to select the client
+     *      certificate to present to the peer.
+     */
+    public SSLSocket(
+        InetAddress address, int port,
+        String hostname, byte[][]alpn,
```
cipherboy, line 575: In your [commit message](https://github.com/frasertweedale/jss/commit/465ac1d9b679334d9531c343988e6609e12586e9) you wrote:

`   Update the private SSLSocket constructor to accept a String[] of ALPN`

But I only see `byte[][]` of ALPNs. I'd prefer we stick with JCA-standard types here rather than introducing our own in general... But at minimum, the commit message should match the code. :-)

org/mozilla/jss/ssl/SSLSocket.c
```diff
@@ -630,6 +656,14 @@ Java_org_mozilla_jss_ssl_SSLSocket_socketConnect
         goto finish;
     }
 
+    /* Set ALPN extension */
+    if (alpn != NULL) {
+        JSS_RefByteArray(env, alpn, (jbyte**)&alpnData, &alpnLength);
+        if (alpnLength > 0) {
+            SSL_SetNextProtoNego(sock->fd, alpnData, alpnLength);
```
cipherboy, line 663: ALPN data only really needs to be provided before the handshake is performed. Not that I mind the extra constructors, but should we add a separate `setApplicationProtocols(...)` function to add this after construction? This would let us set it after construction from a `JSSSocketFactory` (which I think we have in ldap-jdk and/or PKI).

cipherboy: (And IIRC, it can be used with TLSv1.2 and can be provided on later handshakes. Under TLSv1.3 there's only the initial one so I guess it doesn't matter too much.)

cipherboy (changes requested):
This is largely fine.

What's missing is a ALPN selection routine on the server. I'm not sure if you wanted to add that; this is purely client side support, as is mine in #526 (it doesn't do server selection at all--that requires a call to `SSL_SetNextProtoCallback`).

My only thing is... this is the legacy `SSLSocket`... We should really be adding features its replacement, `org.mozilla.jss.ssl.javax.JSSSocket`. 

I guess, what's the use case for still using this one? The other one is compliant with standard JDK interfaces and I'd prefer to push #526 (and then a later one adding server-side ALPN support) rather than continuing to add new features to the old one.

It is on my TODO list to rip out all of this old SSLSocket, including from PKI. 

My 2c. I'll merge if there's a compelling reason to not use the new one for something.

org/mozilla/jss/ssl/SSLSocket.java
```diff
@@ -613,6 +674,14 @@ public SSLSocket(java.net.Socket s, String host,
         resetHandshake();
     }
 
+    /* Get result of ALPN negotiation.
+     *
+     * Prior to the handshake this will always return null.
+     * After the handshake, returns the negotiated protocol
+     * or null (e.g. when the server does not support ALPN).
+     */
+    public native byte[] getNegotiatedProtocol() throws SocketException;
```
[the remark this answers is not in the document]
frasertweedale: Change the name sure.  Return String I am less sure about (discussed elsewhere).

org/mozilla/jss/ssl/SSLSocket.java
frasertweedale: No, it should be `byte[]` - this is how ALPN protocol IDs are defined.  See https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#alpn-protocol-ids and https://tools.ietf.org/html/rfc7301#section-6 .

org/mozilla/jss/ssl/SSLSocket.c
frasertweedale: I don't think we need to worry about the renegotiation use case.  But if it does turn out to be important we could add the extra function then.

Or now... I do prefer supplying as much "config" in the constructor rather than afterwards.  Having to call extra functions to do more setup is something I find distasteful (I use that word in acknowledgement that it is indeed a matter of taste ^_^).  Ideally there would be a single configuration record type that can supply SNI, ALPN and be extensible as new knobs are needed, so that there is  not a combinatorial explosion in the number of possible constructors :)

frasertweedale:
> This is largely fine.
> 
> What's missing is a ALPN selection routine on the server. I'm not sure if you wanted to add that; this is purely client side support, as is mine in #526 (it doesn't do server selection at all--that requires a call to `SSL_SetNextProtoCallback`).
> 
> My only thing is... this is the legacy `SSLSocket`... We should really be adding features its replacement, `org.mozilla.jss.ssl.javax.JSSSocket`.
> 
> I guess, what's the use case for still using this one? The other one is compliant with standard JDK interfaces and I'd prefer to push #526 (and then a later one adding server-side ALPN support) rather than continuing to add new features to the old one.
> 
> It is on my TODO list to rip out all of this old SSLSocket, including from PKI.
> 
> My 2c. I'll merge if there's a compelling reason to not use the new one for something.

I'll have a closer review of #526.  I don't mind to abandon this #526 satisfies the client side use case of ACME tls-alpn-01 challenge.

org/mozilla/jss/ssl/SSLSocket.java
frasertweedale: Just saw your comment on the other PR.  Bravo Java, bravo.  What if the sequence of bytes is not valid UTF-8?  What do you know, there are ALPN GREASE values that are not valid UTF-8 (see https://tools.ietf.org/html/rfc8701, e.g. `{0xFA, 0xFA}`.  If you want to set them, you cannot.  If you want to read them, you cannot (the UTF-8 reader converts them to `��`.

cipherboy: :-)

org/mozilla/jss/ssl/SSLSocket.c
cipherboy: It also turns out (I had forgotten until I made changes in my other PR) -- NSS doesn't support changing ALPN values on renegotiation. Even 0RTT assumes (and validates!) ALPN doesn't change any.

> "single configuration record type"
... So ... [JSSParameters](https://dogtagpki.github.io/jss/master/javadocs/org/mozilla/jss/ssl/javax/JSSParameters.html)? :-)

frasertweedale: @cipherboy yep :D

frasertweedale:
Abandoning in favour of #526 .

PR #641, opened 2020-10-06
fix a misleading comment

frasertweedale:


org/mozilla/jss/ssl/SSLSocket.c
```diff
@@ -618,9 +618,9 @@ Java_org_mozilla_jss_ssl_SSLSocket_socketConnect
         goto finish;
     }
 
-    /*
-     * Tell SSL the URL we think we want to connect to.
-     * This prevents man-in-the-middle attacks.
+    /* Set the Server Name Indication (SNI) extension, to help the
+     * server choose the correct certificate.  This does not prevent
+     * man-in-the-middle attacks (certificate validation does that).
```
cipherboy, line 623: Both comments are actually correct. 

`SSL_SetURL` sadly, does multiple things with the same value: 

 1. Sets the hostname used for certificate validation on the client side,
 2. Sets the hostname used for (e)SNI on the client side.

This makes it difficult to implement a JCA-compatible `java.net.ssl.SSLSocket` implementation with NSS. We assume the hostname (when present) is what is used for cert validation. See my much more detailed comment and commit message in the JSSEngine code:

https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java#L166-L185
https://github.com/dogtagpki/jss/commit/f10c7b96ffd20e17ab62ce4e200acb8c56b62b6b

This was in relation to a QE bug that we ended up fixing for 8.3:
https://github.com/dogtagpki/jss/pull/631

cipherboy (changes requested):
I'd prefer we have both versions of this comment. :-) See justification inline.

org/mozilla/jss/ssl/SSLSocket.c
frasertweedale: Yes, I think you're right - the *default* validation callback appears to use the name given in `SSL_SetURL`.  Thanks for the info, closing PR.

cipherboy: I don't think that's strictly true. `SSL_SetURL` is the setter for `SSL_RevealURL` which at least JSS uses internally in its custom callback verifications:

https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/ssl/callbacks.c#L496

IOW, it should probably be used in all verification callbacks. :-) (Except those where hostname explicitly isn't checked ~cough Java cough~).

PR #642, opened 2020-10-09
JSSSocketChannel.read(): deliver all data from buffer fd

frasertweedale:
```
5e223293 (Fraser Tweedale, 17 minutes ago)
   JSSSocketChannel.read(): deliver all data from buffer fd

   JSSSocketChannel.read() returns 0 when the underlying socket does not yield
   any new data.  But in the case of small output buffers, earlier invocations
   of read() may result in more decrypted data sitting in the unwrapped side
   of the j_buffer, never to be delievered to the application.

   To resolve this issue, continue on to unwrap() even when the underlying
   socket does not yield new data.

   However, only do this after the handshake has been completed. Empirically,
   calling unwrap() at early stages of the handshake with empty buffers breaks
   /something/, causing no data to ever be read from the socket.  (I do not
   understand the exact cause.)

   As part of this change, also refactor how the remote read is handled.  The
   choice of which channel to read and handling of blocking channels is now
   abstracted behind the remoteRead() method, which subsumes boundRead().

   This commit also addresses a subtle bug.  If the readBuffer contains more
   data than the capacity of the j_buffer to which it is written, and if the
   dst buffer(s) get filled, then JSSSocketChannel.read() can terminate with
   leftover data in readBuffer.  Upon the next invocation of read(),
   readBuffer.clear() is called, discarding those data.  This commit modifies
   read() to ensure that leftover data in readBuffer are compacted and
   preserved.

778729ee (Fraser Tweedale, 20 minutes ago)
   JSSSocket.getChannel(): always return channel

   We can always init a channel regardless of whether the parent socket has a
   channel or not.  Indeed, init() always creates a channel, so we can always
   return a channel.
```

### DIFF
--- a/org/mozilla/jss/ssl/javax/JSSSocketChannel.java
+++ b/org/mozilla/jss/ssl/javax/JSSSocketChannel.java
@@ -38,6 +38,8 @@ public class JSSSocketChannel extends SocketChannel {
     private ByteBuffer readBuffer;
     private ByteBuffer writeBuffer;
 
+    private boolean handshakeCompleted = false;
+
     public JSSSocketChannel(JSSSocket sslSocket, SocketChannel parent, Socket parentSocket, ReadableByteChannel readChannel, WritableByteChannel writeChannel, JSSEngine engine) throws IOException {
         super(null);
 
@@ -86,33 +88,28 @@ public class JSSSocketChannel extends SocketChannel {
         autoClose = on;
     }
 
-    /**
-     * Internal helper to bound the size of a blocking read to the maximum
-     * data available.
-     */
-    private int boundRead(int suggested) throws IOException {
-        // When there's consumed data left to read, ensure we bound by the
-        // amount available there before continuing.
-        if (consumed != null && consumed.available() > 0) {
-            return Math.min(consumed.available(), suggested);
+    private int remoteRead() throws IOException {
+        if (consumed != null) {
+            int n = consumedChannel.read(readBuffer);
+            if (n < 0) {
+                consumed = null;
+                consumedChannel = null;
+                return 0;
+            }
+            return n;
+        } else if (isBlocking()) {
+            // blocking channel; we have to bound the read to what is available
+            ByteBuffer slice = readBuffer.slice();
+            int available = parentSocket.getInputStream().available();
+            if (slice.limit() > available)
+                slice.limit(available);
+            int n = readChannel.read(slice);
+            readBuffer.position(readBuffer.position() + Math.max(n, 0));
+            return n;
+        } else {
+            // non-blocking; read whatever is available
+            return readChannel.read(readBuffer);
         }
-
-        // By setting consumed = null when consumed no longer has bytes
-        // available, we provide an easy check for which channel to read
-        // from.
-        consumed = null;
-        consumedChannel = null;
-
-        // If its a non-blocking underlying socket, then return suggested;
-        // it'll read as much as currently available.
-        if (!isBlocking()) {
-            return suggested;
-        }
-
-        // In both remaining cases (no channel or channel is blocking), bound
-        // the read above by the available data in the socket's input stream.
-        int available = parentSocket.getInputStream().available();
-        return Math.min(suggested, available);
     }
 
     public boolean finishConnect() throws IOException {
@@ -200,6 +197,7 @@ public class JSSSocketChannel extends SocketChannel {
             throw new IOException(msg, ssle);
         }
 
+        handshakeCompleted = true;
         sslSocket.notifyHandshakeCompletedListeners();
 
         return true;
@@ -242,53 +240,52 @@ public class JSSSocketChannel extends SocketChannel {
             return -1;
         }
 
-        readBuffer.clear();
-
-        long remoteRead = 0;
         long unwrapped = 0;
         long decrypted = 0;
 
         try {
+            SSLEngineResult result;
             do {
-                int buffer_size = boundRead(readBuffer.remaining());
-                ByteBuffer src = ByteBuffer.wrap(readBuffer.array(), readBuffer.position(), buffer_size);
+                int n = remoteRead();
 
-                int this_read = 0;
-                if (consumed != null) {
-                    this_read += consumedChannel.read(src);
-                } else {
-                    this_read += readChannel.read(src);
-                }
-
-                readBuffer.position(readBuffer.position() + this_read);
-                remoteRead += this_read;
-
-                if (remoteRead == 0) {
-                    return 0;
+                if (readBuffer.position() == 0) {
+                    // We didn't read anything and there is no left-over data.
+                    // If handshake already completed, we can continue because
+                    // there may be data in the unwrapped buffer FD that has yet
+                    // to be delivered to the application.
+                    //
+                    // But if we did not finish the handshake and there were no
+                    // new data, we have to return here.  Calling unwrap() with
+                    // no data, prior to handshake completion, causes the
+                    // handshake to never complete.  This might be a bug, but
+                    // for now we have this workaround.
+                    //
+                    // Note we cannot call engine.getHandshakeStatus() to perform
+                    // this check; it is NOT a passive routine and reading it
+                    // early triggers the same failure.
+                    //
+                    if (!handshakeCompleted) {
+                        return decrypted > 0 ? decrypted : n;
+                    }
                 }
 
                 readBuffer.flip();
 
-                SSLEngineResult result = engine.unwrap(readBuffer, dsts, offset, length);
+                result = engine.unwrap(readBuffer, dsts, offset, length);
                 if (result.getStatus() != SSLEngineResult.Status.OK &&
                     result.getStatus() != SSLEngineResult.Status.BUFFER_UNDERFLOW &&
                     result.getStatus() != SSLEngineResult.Status.CLOSED) {
                     throw new IOException("Unexpected status from unwrap: " + result);
                 }
-
-                readBuffer.flip();
-                readBuffer.compact();
-
                 unwrapped += result.bytesConsumed();
                 decrypted += result.bytesProduced();
 
-                if (unwrapped < remoteRead && result.bytesConsumed() == 0 && result.bytesProduced() == 0) {
-                    String msg = "Calls to unwrap stalled, consuming and ";
-                    msg += "producing no data: unwrapped " + unwrapped;
-                    msg += " bytes of " + remoteRead + " bytes.";
-                    throw new IOException(msg);
-                }
-            } while (unwrapped < remoteRead);
+                readBuffer.compact();
+
+                // If we consumed bytes, there is now room in readBuffer for some
+                // more.  Even if dsts are full, we may be able to consume more
+                // bytes in another call to unwrap().
+            } while (result.bytesConsumed() > 0);
         } catch (SSLException ssle) {
             String msg = "Unable to unwrap data using SSLEngine: ";
             msg += ssle.getMessage();
--- a/org/mozilla/jss/ssl/javax/JSSSocketChannel.java
+++ b/org/mozilla/jss/ssl/javax/JSSSocketChannel.java
@@ -272,10 +272,14 @@ public class JSSSocketChannel extends SocketChannel {
                 readBuffer.flip();
 
                 result = engine.unwrap(readBuffer, dsts, offset, length);
-                if (result.getStatus() != SSLEngineResult.Status.OK &&
-                    result.getStatus() != SSLEngineResult.Status.BUFFER_UNDERFLOW &&
-                    result.getStatus() != SSLEngineResult.Status.CLOSED) {
-                    throw new IOException("Unexpected status from unwrap: " + result);
+                switch (result.getStatus()) {
+                    case CLOSED:
+                        shutdownInput();
+                    case OK:
+                    case BUFFER_UNDERFLOW:
+                        break; // CLOSED, OK and BUFFER_UNDERFLOW are expected
+                    default:
+                        throw new IOException("Unexpected status from unwrap: " + result);
                 }
                 unwrapped += result.bytesConsumed();
                 decrypted += result.bytesProduced();
--- a/org/mozilla/jss/nss/PR.c
+++ b/org/mozilla/jss/nss/PR.c
@@ -176,6 +176,10 @@ Java_org_mozilla_jss_nss_PR_Read(JNIEnv *env, jclass clazz, jobject fd,
     while (read_amount < amount) {
         this_read = PR_Read(real_fd, buffer + read_amount, amount - read_amount);
         if (this_read <= 0) {
+            if (PR_GetError() == 0) {
+                /* End of data */
+                break;
+            }
             if (PR_GetError() == PR_WOULD_BLOCK_ERROR && read_amount > 0) {
                 /* If we've previously gotten data and we would block this
                  * time, then we're at the end of our data. Reset the error


frasertweedale:
LMK what you think about this @cipherboy.  It might need more work  / more testing.

frasertweedale:
Test program for observing `read()` behaviour with small dst buffer size.
(Also available at https://gist.github.com/frasertweedale/9a6c608819a9ba498a86144c14ec9d3c)

```java
import java.net.InetAddress;
import java.nio.charset.StandardCharsets;
import java.nio.channels.SocketChannel;
import java.nio.ByteBuffer;
import java.util.ArrayList;

import javax.net.ssl.KeyManagerFactory;
import javax.net.ssl.KeyManager;
import javax.net.ssl.SSLParameters;
import javax.net.ssl.SSLSocket;
import javax.net.ssl.TrustManagerFactory;
import javax.net.ssl.TrustManager;
import javax.net.ssl.X509TrustManager;

import org.mozilla.jss.CryptoManager;
import org.mozilla.jss.InitializationValues;
import org.mozilla.jss.provider.javax.crypto.JSSKeyManager;
import org.mozilla.jss.ssl.javax.JSSSocketFactory;
import org.mozilla.jss.ssl.javax.JSSParameters;

class Main {

    public static void main(String[] args) {
        String dbdir = "/etc/pki/nssdb";
        InitializationValues iv = new InitializationValues(dbdir);

        String host = "frase.id.au";
        
        try {
            CryptoManager.initialize(iv);

            KeyManagerFactory kmf = KeyManagerFactory.getInstance("NssX509", "Mozilla-JSS");
            KeyManager[] kms = kmf.getKeyManagers();

            TrustManagerFactory tmf = TrustManagerFactory.getInstance("NssX509", "Mozilla-JSS");
            ArrayList<X509TrustManager> tms = new ArrayList<>();
            for (TrustManager tm : tmf.getTrustManagers()) {
                if (tm instanceof X509TrustManager) {
                    tms.add((X509TrustManager) tm);
                }
            }
            X509TrustManager[] xtms = tms.toArray(new X509TrustManager[0]);

            JSSSocketFactory sf = new JSSSocketFactory(
                    "TLS", (JSSKeyManager) kms[0], xtms);

            InetAddress addr = InetAddress.getByName(host);

            SSLSocket sock = sf.createSocket(addr, 443);

            JSSParameters params = new JSSParameters();
            params.setHostname(host);

            sock.setSSLParameters(params);
            System.out.println("> start handshake");
            sock.startHandshake();
            System.out.println("> done handshake");

            String req = "GET /fraser.crt HTTP/1.1\r\nHost: " + host + "\r\n\r\n";
            byte[] reqBytes = req.getBytes(StandardCharsets.UTF_8);
            ByteBuffer buf = ByteBuffer.allocate(128);
            buf.put(reqBytes);
            buf.flip();

            SocketChannel chan = sock.getChannel();
            while (buf.remaining() > 0) chan.write(buf);

            buf.clear();
            int n;
            while ((n = chan.read(buf)) >= 0) {
                buf.flip();
                System.out.println("> read " + n);
                if (n > 0) {
                    System.out.println("vvvvvv");
                    System.out.write(buf.array(), 0, n);
                    System.out.println("\n^^^^^^");
                }
                buf.clear();
                Thread.sleep(1000);
            }

        } catch (Throwable e) {
            e.printStackTrace();
        }

        System.out.println("done");
    }

}
```

frasertweedale:
Note that the program never terminates because it keeps returning 0, even after underlying socket has (presumably) closed.
This is a different bug / shortcoming to be addressed.  But you can just press ^C to kill the test program at your pleasure.

frasertweedale:
You might also need the fix from https://github.com/dogtagpki/jss/pull/643 to test this.

org/mozilla/jss/ssl/javax/JSSSocket.java
```diff
@@ -234,10 +234,6 @@ public JSSEngine getEngine() {
      * @see java.net.Socket#getChannel()
      */
     public JSSSocketChannel getChannel() {
-        if (parent.getChannel() == null) {
```
cipherboy, line 237: I wouldn't do this either.

We've discussed this with @edewata earlier. See #532, and https://github.com/dogtagpki/jss/pull/532#discussion_r421152265.

As I understand more about this architecture... a Socket is either a blocking or non-blocking file descriptor, but a SocketChannel is non-blocking. Only in the event that a Socket is blocking will a SocketChannel be created. 

In JSS we abuse this slightly. We use the `JSSSocket` as the front to SSL operations, but in order to "do" SSL, we use the `JSSSocketChannel` as an intermediary between the `JSSSocket` and the `JSSEngine`.

Note that callers using a `SocketChannel` are mostly expected to do SSL operations themselves directly with a `SSLEngine`. There's no relevant methods in the standard to upcast a `SocketChannel` into a `SSLSocketChannel`.

cipherboy (commented):
I'll need to take a closer look at this.

But one comment I wanted to make is that `consumed` exists only in the case where we use a non-SSL raw [`ServerSocketFactory`](https://docs.oracle.com/javase/8/docs/api/javax/net/ServerSocketFactory.html) to create a `ServerSocket`, and then, as the handshake starts, we read the initial `ClientHello` and use that to inform ourselves what certificate to select, passing both the initial `ClientHello` and the relevant parameters into [the `SSLSocketFactory`](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLSocketFactory.html#createSocket-java.net.Socket-java.io.InputStream-boolean-). 

Note that it is trivially impossible to read `created` post-handshake except with 0RTT in TLSv1.3 (and perhaps session resumption?) because there it requires a message from the client, in response to one sent from the server before any body is sent. 

In those two scenarios (0RTT and session resumption), `created` doesn't contain the information that Java would need anyways, unless they maintained a session cache of their own. I'm not sure how that works with NSS either; with multiple certs, if we configure the wrong one and then attempt continuation, will NSS point us to the correct one, or will it fail? Hmmmm...

jss.spec
```diff
@@ -117,7 +117,7 @@ modutil -dbdir /etc/pki/nssdb -chkfips true | grep -q enabled && export FIPS_ENA
 cd %{_vpath_builddir}
 %{__make} all
 %{__make} javadoc
-ctest --output-on-failure
+#ctest --output-on-failure
```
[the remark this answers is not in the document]
frasertweedale: Sorry, didn't mean to commit this - was just making the build speed up :-)

org/mozilla/jss/ssl/javax/JSSSocket.java
cipherboy: (IOW, it is tempting to grab the underlying `JSSSocketChannel`, but if you truly want non-blocking sockets, you should create a non-blocking socket to begin with...)

frasertweedale:
@cipherboy a couple more commits fixing two more issues:

- the never returns `-1` issue mentioned [above](https://github.com/dogtagpki/jss/pull/642#issuecomment-706020006)
- an error where when peer closes their send end, the data from the final read preceding end-of-stream is lost on our end.

I also removed the "always return socket channel" commit as requested, and removed the change to the spec file that I committed by mistake.

Updated test program (changed back to `InputStream`/`OutputStream` interface):
```java
import java.io.ByteArrayOutputStream;
import java.io.InputStream;
import java.io.OutputStream;
import java.net.InetAddress;
import java.nio.charset.StandardCharsets;
import java.util.ArrayList;

import javax.net.ssl.KeyManagerFactory;
import javax.net.ssl.KeyManager;
import javax.net.ssl.SNIServerName;
import javax.net.ssl.SNIHostName;
import javax.net.ssl.SSLParameters;
import javax.net.ssl.SSLSocket;
import javax.net.ssl.TrustManagerFactory;
import javax.net.ssl.TrustManager;
import javax.net.ssl.X509TrustManager;

import org.mozilla.jss.CryptoManager;
import org.mozilla.jss.InitializationValues;
import org.mozilla.jss.provider.javax.crypto.JSSKeyManager;
import org.mozilla.jss.ssl.javax.JSSSocketFactory;
import org.mozilla.jss.ssl.javax.JSSParameters;

class Main {

    public static void main(String[] args) {
        String dbdir = "/home/ftweedal/dev/jss/nssdb";
        InitializationValues iv = new InitializationValues(dbdir);

        String host = "frase.id.au";
        String[] alpn = { "wat1", "wat2", "wat3" };

        try {
            CryptoManager.initialize(iv);

            KeyManagerFactory kmf = KeyManagerFactory.getInstance("NssX509", "Mozilla-JSS");
            KeyManager[] kms = kmf.getKeyManagers();

            TrustManagerFactory tmf = TrustManagerFactory.getInstance("NssX509", "Mozilla-JSS");
            ArrayList<X509TrustManager> tms = new ArrayList<>();
            for (TrustManager tm : tmf.getTrustManagers()) {
                if (tm instanceof X509TrustManager) {
                    tms.add((X509TrustManager) tm);
                }
            }
            X509TrustManager[] xtms = tms.toArray(new X509TrustManager[0]);

            JSSSocketFactory sf = new JSSSocketFactory(
                    "TLS", (JSSKeyManager) kms[0], xtms);

            InetAddress addr = InetAddress.getByName(host);

            SSLSocket sock = sf.createSocket(addr, 443);

            /* this doesn't work because JSSEngine ignores the SNI params.
             * Need to use JSSParameters and setHostname().
             *
             * Should update JSSParameters "downcast" constructor to set
             * hostname from SNI names, if given.
             */
            /*
            SSLParameters params = new SSLParameters();
            ArrayList<SNIServerName> sni = new ArrayList<>();
            sni.add(new SNIHostName(host));
            params.setServerNames(sni);
            */

            JSSParameters params = new JSSParameters();
            params.setHostname(host);

            params.setApplicationProtocols(alpn);

            sock.setSSLParameters(params);
            System.out.println("> start handshake");
            sock.startHandshake();
            System.out.println("> done handshake");

            String proto = sock.getApplicationProtocol();
            if (proto == null) {
                System.out.println("> ALPN STATUS UNKNOWN");
            } else if (proto.isEmpty()) {
                System.out.println("> ALPN NO PROTO");
            } else {
                System.out.println("> ALPN PROTO = " + proto);
            }

            String req = "GET /fraser.crt HTTP/1.1\r\nHost: " + host + "\r\n";
            req += "Connection: close\r\n";
            req += "\r\n";
            byte[] reqBytes = req.getBytes(StandardCharsets.UTF_8);
            OutputStream out = sock.getOutputStream();
            out.write(reqBytes);
            out.flush();
            //sock.shutdownOutput();

            try (InputStream in = sock.getInputStream()) {
                ByteArrayOutputStream result = new ByteArrayOutputStream();
                byte[] buffer = new byte[256];
                int n;
                while ((n = in.read(buffer)) >= 0) {
                    System.out.println("> read " + n);
                    if (n > 0) {
                        System.out.println("vvvvvv");
                        System.out.write(buffer, 0, n);
                        System.out.println("\n^^^^^^");
                    }
                    Thread.sleep(500);
                }
            }

        } catch (Throwable e) {
            e.printStackTrace();
        }

        System.out.println("done");
    }

}
```

cipherboy:
@frasertweedale -- Sorry, been busy. I'll review this later this week and land before 4.8.0 GA.

frasertweedale:
@cipherboy no worries, don't rush this one; a proper review will take some time.

cipherboy:
@frasertweedale Do we also need a compact in the write case?

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -242,53 +240,54 @@ public synchronized long read(ByteBuffer[] dsts, int offset, int length) throws
...
 
-                SSLEngineResult result = engine.unwrap(readBuffer, dsts, offset, length);
-                if (result.getStatus() != SSLEngineResult.Status.OK &&
-                    result.getStatus() != SSLEngineResult.Status.BUFFER_UNDERFLOW &&
-                    result.getStatus() != SSLEngineResult.Status.CLOSED) {
-                    throw new IOException("Unexpected status from unwrap: " + result);
+                result = engine.unwrap(readBuffer, dsts, offset, length);
+                switch (result.getStatus()) {
+                    case CLOSED:
+                        shutdownInput();
```
cipherboy, line 275: I think the same applies for write, no? We should shutdown output if the engine reports closed?

frasertweedale: @cipherboy probably?  Would you be willing to investigate and, if necessary, deal with that in a separate patch?

cipherboy: Sure, I can do that.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -86,33 +88,28 @@ public void setAutoClose(boolean on) {
...
-        consumedChannel = null;
-
-        // If its a non-blocking underlying socket, then return suggested;
-        // it'll read as much as currently available.
-        if (!isBlocking()) {
-            return suggested;
+    private int remoteRead() throws IOException {
+        if (consumed != null) {
+            int n = consumedChannel.read(readBuffer);
+            if (n < 0) {
```
cipherboy, line 94: According to https://docs.oracle.com/javase/8/docs/api/java/nio/channels/ReadableByteChannel.html, 0 is a valid return result and we should probably set `consumed` and `consumedChannel` to `null` in both cases.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -242,53 +240,54 @@ public synchronized long read(ByteBuffer[] dsts, int offset, int length) throws
...
+                    // But if we did not finish the handshake and there were no
+                    // new data, we have to return here.  Calling unwrap() with
+                    // no data, prior to handshake completion, breaks something.
+                    //
+                    // Note we cannot call engine.getHandshakeStatus() to perform
+                    // this check; it is NOT a passive routine and reading it
+                    // early actually causes the same breakage.
+                    //
+                    if (!handshakeCompleted) {
+                        return decrypted > 0 ? decrypted : n;
```
cipherboy, line 266: I'm not sure I agree with this.

When `decrypted > 0`, we obviously return `decrypted`.

However, in the case where `decrypted <= 0` (really, it must be 0), the contract of our `read()` call is the number of bytes decrypted, which is 0. It isn't the number of bytes read from the remote peer.

The best example of this is during handshake: `read` during handshake returns 0 if no data is present -- not because no wire data was transferred (there presumably might have been) -- but because none of it resulted in any decryptable data. (Ignoring 0RTT). 

IMO, since `decrypted < 0` won't ever occur, it should be safe to return `decrypted`.

cipherboy (changes requested):
IMO both of these should be fixed prior to merging. But I think that's it, thanks @frasertweedale! Do you mind if I add your test case to the the suite in a follow-up PR?

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
frasertweedale: 0 is indeed a valid return value.  It means that zero bytes were read.  It doesn't mean end of stream was reached, so I think it's wrong to discard the `consumedChannel` in this case.

Based on `consumedChannel = Channels.newChannel(InputStream consumed)`, return value of 0 seems to be a "can't happen" and therefore your proposal is (vacuously) safe.  But I'd rather code to the interface than rely on these implementation details.

If you still feel strongly we sohuld change it, I will change it and add some commentary.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
frasertweedale: The goal here is to catch the case where `remoteRead() < 0` indicating end of stream.  I think this can occur if the connection gets terminated abnormally, i.e. the TLS termination alerts didn't get sent/processed (which needs to happen so `inboundClosed` gets set).

If all streams get closed "properly" then I agree with your proposal.  But otherwise, I want to be very careful to avoid situations where `JSSSocketChannel.read()` could return `0` indefinitely.

What do you think about this scenario @cipherboy?  Is there a better way to handle it?

frasertweedale:
> Do you mind if I add your test case to the the suite in a follow-up PR?

That is a fine idea.  Not only do I not mind... I insist!

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
cipherboy: If `remoteRead` returns non-zero, shouldn't we just return `remoteRead`'s value? When the underlying TCP connections terminate unexpectedly, IMO we should never attempt to wrap/unwrap, even if we had old or uncompacted data lying around. We should probably close the TLS SSLEngine's inbound and exit with `-1`, regardless of handshake status. I think then this entire nested if statement gets simplified and removed, replaced with one that always checks the `remoteRead` value. 

This might require a little restructuring, as I think the `close()` handler doesn't yet understand that the inbound was already closed and to only close the outbound in that case (or visa-versa) but IMO it is a worthwhile improvement.

Am I missing a case where this wouldn't be correct? I thought it was one of the core guarantees from TLS's RFC.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
```diff
@@ -242,53 +240,54 @@ public synchronized long read(ByteBuffer[] dsts, int offset, int length) throws
...
+                    // We didn't read anything and there is no left-over data.
+                    // If handshake already completed, we can continue because
+                    // there may be data in the unwrapped buffer FD that has yet
+                    // to be delivered to the application.
+                    //
+                    // But if we did not finish the handshake and there were no
+                    // new data, we have to return here.  Calling unwrap() with
+                    // no data, prior to handshake completion, breaks something.
+                    //
+                    // Note we cannot call engine.getHandshakeStatus() to perform
```
cipherboy, line 261: What was this failure, btw? Do you still happen to have the logs?

frasertweedale: The handshake never completes.  I am not sure the precise mechanism.  I will clarify the comment.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
frasertweedale: > However, in the case where decrypted <= 0 (really, it must be 0), the contract of our read() call is the number of bytes decrypted, which is 0. It isn't the number of bytes read from the remote peer.

> If remoteRead returns non-zero, shouldn't we just return remoteRead's value?

The return contract of `SocketChannel.read()` is number of bytes read (possibly `0`) or `-1` when *this channel* reaches end of stream.  If `remoteRead()` returns `-1`, there might still be data in `readBuffer` or in the engine's internal read buffer, so we can't return `-1` until `unwrap` yields no more data.

> IMO we should never attempt to wrap/unwrap, even if we had old or uncompacted data lying around.

Disagree.  Because there may be undecrypted data sitting in buffers waiting to be processed, we *cannot even know* if the closure of the parent socket's input stream is expected or abnormal.  Therefore even when `remoteRead()` returns `-1`, we have to wait until `unwrap` yields no more data before cleaning up and returning `-1` from `read()`.

> This might require a little restructuring

I'm sure the structure can be improved.  But... have I convinced you that the statements / expressions we're arguing about are at least not *obviously* wrong or harmful?  If so, could we merge this PR, add a regression test adapted from my test program, and go forward from there?

Otherwise... could you write a program to demonstrate bug(s) in the code fragments we're arguing about.  I suggest it not because I'm sure my code is correct (I am not!), but because concrete demonstrations of issues provided a basis for fixing them (and then for regression tests).

tl;dr this PR fixes some major issues (proven by the test program); there may well be problems with my approach but let's either make them concrete so I can understand and fix them, or merge this PR.

frasertweedale:
Latest push is a trivial rebase with minor clarifications to some comments.  No other changes.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
cipherboy: Ah I'm with you now.

I keep thinking -1 return value is only on an _error_ -- i.e., stream _incorrectly_ closed or was otherwise terminated.

Since it returns -1 on mundane "no more data for now" things, yeah, I think this is a good. I'll go ahead and merge.

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
cipherboy: Interesting. I'm fairly certain we test calling `getHandshakeStatus()` in a conditional in the existing test suite.

cipherboy (approved):

cipherboy:
Thanks @frasertweedale! Merging... :)

frasertweedale:
Thanks heaps @cipherboy!

org/mozilla/jss/ssl/javax/JSSSocketChannel.java
frasertweedale: @cipherboy small correction: -1 means (or is supposed to mean) "no more data *ever*", which may or may not be mundane :)

PR #647, opened 2020-10-20
Merge RHEL's jss.spec

edewata:


cipherboy (changes requested):
NACK'd. 

These hacks were necessary in RHEL 8.3. For RHEL 8.4, we should have a new CMake version and poke the relevant bug if not.

edewata:
OK, it's removed from RHEL. Thanks!

PR #648, opened 2020-10-26
Log exceptions when libjss4.so failed to load

cipherboy:

---

This was helpful in debugging @jmagne's issue, so I think we should include it. In particular, when `libjss4.so` is linked incorrectly, all three loads will fail, but only the second one will contain useful information.


jmagne (approved):
Looks good and I tested this myself and it was helpful, aproving.

PR #653, opened 2020-10-29
Replace RSAPrivateKey with RSAKey

cipherboy:
Per discussion with Andrew Hughes in #java, JCA takes `RSAPrivateKey` to mean that the private exponent is extractable. Because the JCA is careful to use Key most places, we might get away with making `PK11RSAPrivateKey` not implement `RSAPrivateKey` and instead only implement `RSAKey` -- signalling to other providers that the private exponent isn't extractable.

This patch requires careful testing with SunJSSE and Dogtag PKI to ensure we don't rely on `PK11RSAPrivateKey` implementing `RSAPrivateKey` somewhere.

----

This patch targets v4.6.x branch -- I'm torn whether or not we should fix v4.7.x+ as we replace the SunJSSE TLS implementation with our own.

### DIFF
--- a/org/mozilla/jss/pkcs11/PK11RSAPrivateKey.java
+++ b/org/mozilla/jss/pkcs11/PK11RSAPrivateKey.java
@@ -7,7 +7,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 class PK11RSAPrivateKey
-    extends PK11PrivKey implements java.security.interfaces.RSAPrivateKey
+    extends PK11PrivKey implements java.security.interfaces.RSAKey
 {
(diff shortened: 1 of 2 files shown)


edewata:
Looks like the F30 test needs to be removed.

Could `PK11RSAPrivateKey` inherit from `org.mozilla.jss.crypto.PrivateKey` or `java.security.PrivateKey`, so at least it's still a `PrivateKey` instead of a generic `Key`?

cipherboy:
@edewata  -- `PK11RSAPrivateKey` inherits from `PK11PrivateKey`, which implements `java.security.PrivateKey`, so it is already a `PrivateKey` instance instead of a generic `Key` instance. :-)

edewata (approved):
Sounds good then. ACK.

PR #654, opened 2020-11-02
Bump tomcatjss, pki-core conflicts due to lang3

cipherboy:
This version of JSS requires apache-commons-lang3 and shouldn't be used
with older version of pki-core as a result.

As reported by @flo-renaud


edewata (approved):
There are issues with the CI, but the changes should be harmless.

PR #656, opened 2020-11-05
Only check PKCS11Constants on beta builds -- v4.7.x backport

cipherboy:
Recent errors with PKCS11Constants have shown that we shouldn't be
running these tests on release builds for backports: only for
pre-release content. Only run them when the beta bit is set.


cipherboy:
`pki-build` is incorrectly configured; the rest looks fine.

PR #660, opened 2020-11-18
Add support for CKM_AES_KEY_WRAP_KWP

cipherboy:
Add support for `CKM_AES_KEY_WRAP_KWP` algorithm

This algorithm is the only mechanism supported by certain lines of
nCipher HSMs. Unlike other key-wrapping mechanisms, this one generates a
temporary AES key to encrypt the wrapped key with.


cipherboy:
The GitHub actions runners seem to have problems. NSS rebuild fails due to uninitialized variable, but otherwise, it is good shape when run locally.

PR #663, opened 2020-11-23
Fix JSS with SunJSSE's TLSv1.3

cipherboy:
SunJSSE's new TLSv1.3 support (backported to JDK8 on RHEL 8.2.z) changes two things over the previous TLSv1.2-compatible stack:

 1. Changes handling of MAC creating, using native JDK provider key type instead of SymmetricKey directly.
 2. Requires RSA/PSS signature support.

To do this, we backport four patches from the master branch:

 1. Allow JSSMacSpi to handle external, exportable symmetric keys by reporting them using our SecretKeyFactory.
 2. My original forward-port of @jmagne's v4.4.x-branch PSS patch.
 3. One additional fix in the SHA512withRSA/PSS identifier which accidentally used SHA256.
 4. One additional fix by @jmagne in netscape.security's Cert handling.

Note that the latter patch, while not strictly required, is useful in case we ever decide to backport PSS support in pki-core (unlikely) -- but at least we have a completed, working implementation.


cipherboy:
As this is all backports and has already seen review, I'm going ahead and merging. Any review feedback should probably be applied to master branch first.

PR #662, opened 2020-11-20
Attempt to use SecretKeyFactory to rewrite keys prior to failing JSSMacSpi

cipherboy:
In `JSSCipherSpi`, if a key from another provider is passed in, we first
attempt to export it and then import it into NSS. Only when this fails
do we raise an exception. This fixes a failure in JDK8 272-b10 when
JSS is used in conjunction with SunJSSE:

    javax.net.ssl.SSLHandshakeException: Could not generate secret
            at sun.security.ssl.ECDHKeyExchange$ECDHEKAKeyDerivation.t13DeriveKey(ECDHKeyExchange.java:479)
            at sun.security.ssl.ECDHKeyExchange$ECDHEKAKeyDerivation.deriveKey(ECDHKeyExchange.java:419)
            at sun.security.ssl.ServerHello$T13ServerHelloProducer.produce(ServerHello.java:596)
            at sun.security.ssl.SSLHandshake.produce(SSLHandshake.java:421)
            at sun.security.ssl.ClientHello$T13ClientHelloConsumer.goServerHello(ClientHello.java:1152)
            at sun.security.ssl.ClientHello$T13ClientHelloConsumer.consume(ClientHello.java:1088)
            at sun.security.ssl.ClientHello$ClientHelloConsumer.onClientHello(ClientHello.java:725)
            at sun.security.ssl.ClientHello$ClientHelloConsumer.consume(ClientHello.java:693)
            at sun.security.ssl.SSLHandshake.consume(SSLHandshake.java:377)
            at sun.security.ssl.HandshakeContext.dispatch(HandshakeContext.java:444)
            at sun.security.ssl.SSLEngineImpl$DelegatedTask$DelegatedAction.run(SSLEngineImpl.java:968)
            at sun.security.ssl.SSLEngineImpl$DelegatedTask$DelegatedAction.run(SSLEngineImpl.java:955)
            at java.security.AccessController.doPrivileged(Native Method)
            at sun.security.ssl.SSLEngineImpl$DelegatedTask.run(SSLEngineImpl.java:902)
            at org.apache.tomcat.util.net.SecureNioChannel.tasks(SecureNioChannel.java:443)
            at org.apache.tomcat.util.net.SecureNioChannel.handshakeUnwrap(SecureNioChannel.java:507)
            at org.apache.tomcat.util.net.SecureNioChannel.handshake(SecureNioChannel.java:238)
[...]

PR #665, opened 2020-11-23
Backport of Attempt to use SecretKeyFactory to rewrite keys prior to failing JSSMacSpi

cipherboy:
Backport of #662.


cipherboy:
Debain test known failure, signature MD2/MD5/SHA-1 fixes weren't backported from master.

pki_build also known failure, it doesn't build with the right PKI version.

PR #672, opened 2020-12-17
Fix more issues with old tests

cipherboy:
- In KeyFactoryTest, we were still creating 512-bit RSA/DSA keys. Remove DSA, remove references to IBM, and rely on policy to decide keysize.
 - Small changes to BadSSL. Note that the test still doesn't pass with strict OCSP validation due to https://bugzilla.redhat.com/show_bug.cgi?id=1861495


edewata (approved):
LGTM

PR #673, opened 2020-12-18
Bug1908541-jss broke SCEP - missing PasswordChallenge class

ladycfu:
This patch, together with the fix for Bug1664435 Error instantiating class for challenge_password with SCEP request, addresses the issue where  the class PasswordChallenge cannot be loaded due to Class Loader differences.
      jss is installed in the common CL (/usr/share/pki/server/common/lib/jss4.jar)
      the servlet classes are in webapp CL (/usr/share/pki/server/webapps/pki/WEB-INF/lib/pki-cms.jar)

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1664435

### DIFF
--- a/org/mozilla/jss/netscape/security/x509/ExtensionsRequested.java
+++ b/org/mozilla/jss/netscape/security/x509/ExtensionsRequested.java
@@ -1,0 +1,176 @@
+// --- BEGIN COPYRIGHT BLOCK ---
+// This program is free software; you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation; version 2 of the License.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


org/mozilla/jss/netscape/security/x509/ExtensionsRequested.java
```diff
@@ -0,0 +1,176 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.security.cert.CertificateException;
+import java.util.Enumeration;
+import java.util.Vector;
+
+import org.mozilla.jss.netscape.security.util.DerInputStream;
+import org.mozilla.jss.netscape.security.util.DerValue;
+import org.mozilla.jss.netscape.security.x509.CertAttrSet;
```
cipherboy, line 30: Technically these are redundant (and in the file above), but its fine to keep it if you want a "move only" patch. :)

cipherboy (approved):

PR #674, opened 2021-01-18
Only provision OAEP parameters when non-NULL

cipherboy:


cipherboy:
@jmagne Do you remember why I filed this?

fmarco76:
Is there a case for OAEP without parameters requiring this change? If it is then the implementation looks OK to me and we could rebase and merge.

fmarco76:
@cipherboy I am not sure there are implicit OAEP parameters but in any case the PR looks safe and should not provide any problem to the current supported cases and be safer for the other. Do you remember why this has not been merged?

cipherboy:
@fmarco76 Honestly, no clue.

Maybe this was reported by covscan or similar?

fmarco76:
@cipherboy OK. I think we can go ahead with this . Could you rebase to solve the conflicts?

jmagne (commented):
This looks totally fine and non destructive. It can't hurt even if it is not needed for some reason.

cipherboy:
@fmarco76 Updated

fmarco76:
@cipherboy Thanks!

PR #676, opened 2021-02-15
Add CertificateChain converters

edewata:
The `CertificateChain` class has been modified to support importing/exporting certificates from/to a series of PEM certificates or PKCS #7 data. This is needed for https://github.com/dogtagpki/pki/pull/3459.

Note: The some of the CI tests are failing due to an existing issue.

### DIFF
--- a/org/mozilla/jss/netscape/security/x509/CertificateChain.java
+++ b/org/mozilla/jss/netscape/security/x509/CertificateChain.java
@@ -20,7 +20,9 @@ package org.mozilla.jss.netscape.security.x509;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.PrintWriter;
 import java.io.Serializable;
+import java.io.StringWriter;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -28,6 +30,7 @@ import java.util.List;
 
 import org.mozilla.jss.netscape.security.pkcs.PKCS7;
 import org.mozilla.jss.netscape.security.util.Cert;
+import org.mozilla.jss.netscape.security.util.Utils;
 
 public class CertificateChain implements Serializable {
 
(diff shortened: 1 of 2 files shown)


org/mozilla/jss/netscape/security/x509/CertificateChain.java
```diff
@@ -150,6 +153,85 @@ private void readObject(java.io.ObjectInputStream in)
...
+    public void addCertificate(X509Certificate cert) {
+        certs.add(cert);
+    }
+
+    public void addCertificateChain(CertificateChain certChain) {
+        certs.addAll(certChain.certs);
+    }
+
+    public void addPKCS7(PKCS7 pkcs7) {
+        certs.addAll(Arrays.asList(pkcs7.getCertificates()));
```
cipherboy, line 165: I hate to be a spoilsport, but I don't think this is strictly true all the time. For instance, CA bundles are usually a concat'd list of PKCS#7 format PEM certificates (with headers/footers & the works), but they're not strictly a _chain_ (with a connected path from root -> end-entity cert). For what you're doing in https://github.com/dogtagpki/pki/pull/3459, this will suffice, but in general, it isn't strictly true that a PKCS#7 will always contain a strict chain of certs.

Where I'm going with this is, you might want to decide: do you want to validate the contents of the PKCS#7 and ensure it is a single path from root -> leaf, or do you want to ignore the problem and hand-wave it away (does it break sorting?) or what. :-)

My 2c.

edewata: OK, I've updated the patch to sort (and thus validate) the cert chain so it will only accept a single chain and reject a CA bundle. This patch will still be used by the `pki pkcs7` tools in dogtagpki/pki#3459, but for OCSP installation I think we can get the CA cert from subsystem cert chain instead of the CA bundle, so we don't have to deal with multiple chains. We can also try to remove the CA cert from OCSP's CRL issuing point record as we discussed on IRC, but I think that can be solved separately later.

cipherboy: ty! :)

org/mozilla/jss/netscape/security/x509/CertificateChain.java
```diff
@@ -150,6 +153,89 @@ private void readObject(java.io.ObjectInputStream in)
...
+
+        return sw.toString();
+    }
+
+    /**
+     * Convert the certificate chain into a PKCS #7 object.
+     */
+    public PKCS7 toPKCS7() throws Exception {
+
+        // convert X509Certificate into X509CertImpl
```
cipherboy, line 230: I think the same comment as above applies here: do we want to sort this chain first? AFAIK, we don't guarantee internally that certs are in sorted order... I don't think it is strictly required either, but we could want it so we could enforce that this is strictly a chain in a CertificateChain class? Up to you.

cipherboy (approved):

org/mozilla/jss/netscape/security/x509/CertificateChain.java
edewata: Agreed. I added `sort()` in `toPEMString()` and `toPKCS7()`.

edewata:
@cipherboy Thanks! All required tests passed. Merging...

PR #677, opened 2021-02-24
Switch to checking for AES_KEY_WRAP by default

cipherboy:

Code I had lying around, should probably be merged at some point in the next release cycle. Was pending discussions with Jack.


fmarco76:
@cipherboy this PR seems straight-forward. Do you remember the pending discussion and/or the reason it is not merged?

cipherboy:
I don't, no. Perhaps I just didn't get around to it in time.

I suspect we wanted a better algorithm for key transit to/from HSMs that we
could control at the higher level than what NSS does under the covers for
us. But maybe we've not run into any of those issues yet.

I think on my NSS fork, there's an issue discussing this.


- A

fmarco76:
> I think on my NSS fork, there's an issue discussing this.

I found the PR in your NSS fork about key mobility to/from HSM. The current NSS code has evolved and there are differences so I need to investigate more on this. Thanks!

Issue #334, opened 2019-12-18
FIPS mode - ImportKey doesn't work


cipherboy:
We import symmetric keys in non-FIPS mode using [`PK11_ImportSymKeyWithFlags(...)`](https://github.com/dogtagpki/jss/blob/master/org/mozilla/jss/pkcs11/PK11KeyWrapper.c#L767-L773). 

This doesn't work in FIPS mode. Instead, we'll need to generate a new key, encrypt the existing key, and unwrap it into the NSS DB.

Note that I broke this as I suggested we use ImportKey instead of the earlier dance with Unwrap/Wrap in PKI and the test suite; this should fix that.

cipherboy:
I think e791542e5ac9bad588cab84da50a218b799c3643 is the commit I was thinking of in JSS and https://github.com/dogtagpki/pki/commit/879114a41d8383ef544906e637fbacac14d21bc6 in PKI.

PR #678, opened 2021-02-25
Encrypt & unwrap symmetric key in FIPS mode

cipherboy:
NSS doesn't generally allow keys to be imported in FIPS mode. However,
for portability with other JCA providers, we sometimes need to import
keys from byte arrays. Do this in the JNI layer by executing a PKCS#11
encrypt and then unwrap using the same key. This lets us effectively
"import" a key into a token, if the token supports using the given
mechanism for both encryption and unwrapping operations. Some HSMs are
getting stricter about this and forbid using the same key for encrypt
and unwrap operations.

Resolves: #334

### DIFF
--- a/org/mozilla/jss/pkcs11/PK11KeyWrapper.c
+++ b/org/mozilla/jss/pkcs11/PK11KeyWrapper.c
@@ -712,6 +712,61 @@ finish:
     return keyObj;
 }
 
+PK11SymKey *JSS_PK11_ImportSymKeyWithFlagsFIPS(PK11SlotInfo *slot, CK_MECHANISM_TYPE type,
+        CK_ATTRIBUTE_TYPE operation, SECItem *key, CK_FLAGS flags,
+        PRBool isPerm, void *wincx)
+{
+    PK11SymKey *result = NULL;
+    PK11SymKey *wrapper = NULL;
+    SECStatus ret = SECFailure;
+    unsigned int wrapped_len = 0;
+    unsigned int wrapped_max = key->len + 64;
+    unsigned char *wrapped_key = calloc(wrapped_max, sizeof(unsigned char));
+    SECItem wrapped_item = { siBuffer, wrapped_key, 0 };
+    SECItem *param = NULL;
+
+    /* Steps:
+     * 1. Generate a temporary key to encrypt and unwrap with,
+     * 2. Encrypt our key to import using the wrapping key,
+     * 3. Unwrap into the token using the wrapping key.
+     */
+
+#define FIPS_KEYGEN_ALGO CKM_AES_KEY_GEN
+#define FIPS_ENCRYPT_UNWRAP_ALGO CKM_AES_KEY_WRAP_PAD
+
+    wrapper = PK11_KeyGen(slot, FIPS_KEYGEN_ALGO, NULL, 32, wincx);
+    if (wrapper == NULL) {
+        goto done;
+    }
+
+    param = PK11_GenerateNewParam(FIPS_ENCRYPT_UNWRAP_ALGO, wrapper);
+    if (param == NULL) {
+        goto done;
+    }
+
+    ret = PK11_Encrypt(wrapper, FIPS_ENCRYPT_UNWRAP_ALGO, param,
+            wrapped_key, &wrapped_len, wrapped_max,
+            key->data, key->len);
+    if (ret != SECSuccess) {
+        goto done;
+    }
+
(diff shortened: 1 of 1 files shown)


ladycfu (approved):
Looks good.  And I tested it to work as well.

org/mozilla/jss/pkcs11/PK11KeyWrapper.c
```diff
@@ -712,6 +712,59 @@ Java_org_mozilla_jss_pkcs11_PK11KeyWrapper_nativeUnwrapSymWithPriv
...
+    SECItem wrapped_item = { siBuffer, wrapped_key, 0 };
+    SECItem *param = NULL;
+
+    /* Steps:
+     * 1. Generate a temporary key to encrypt and unwrap with,
+     * 2. Encrypt our key to import using the wrapping key,
+     * 3. Unwrap into the token using the wrapping key.
+     */
+
+#define FIPS_ENCRYPT_UNWRAP_ALGO CKM_AES_CBC_PAD
```
cipherboy, line 734: The one concern I have about this is we need to make sure we choose an algorithm that does both encrypt and unwrap -- this algorithm should, but if certain HSM vendors take away that capabilities later, we might have problems.

cipherboy: We should probably replace this with `CKM_AES_KEY_WRAP_PAD`. @ladycfu do you want to re-test this PR with HSM with that change?

ladycfu: Yes sounds like CKM_AES_KEY_WRAP_PAD would be the one to work on the hsm.  I'll give it a try.

ladycfu: @cipherboy So, unfortunately, CKM_AES_KEY_WRAP_PAD doesn't work for nss in FIPS mode.  I see
-668080384[7fc8d033f690]:   phKey = 0x7fc8d04bbef8
-668080384[7fc8d033f690]:     CKA_VALUE_LEN = 32
-668080384[7fc8d033f690]:     CKA_ENCRYPT = CK_TRUE [1]
-668080384[7fc8d033f690]:     CKA_SIGN = CK_TRUE [1]
-668080384[7fc8d033f690]:       mechanism = CKM_GENERIC_SECRET_KEY_GEN
-668080384[7fc8d033f690]:   *phKey = 0x1
-668080384[7fc8d033f690]:   rv = CKR_OK
.
.
-668080384[7fc8d033f690]:       mechanism = 0x210a
-668080384[7fc8d033f690]:   rv = CKR_KEY_TYPE_INCONSISTENT

While CKM_AES_CBC_PAD in your original patch works.  Thoughts?

===
now, about hsm.  I am only able to get the keys generated when building archival options using CRMFPopClient, if I change the code ot use temporary=true and sensitive=false.  So that's not exercising your code yet.  To exercise your code in this patch, I'll need to run CMCRequest.  Some things need to be setup first which might take some time.  
Just thought I should give you a preliminary report for now.

ladycfu: per @cipherboy 's suggestion, replacing the PK11_KeyGen algorithm with CKM_AES_KEY_GEN did the trick.  So the diff from both his suggestions for this PR is
diff --git a/org/mozilla/jss/pkcs11/PK11KeyWrapper.c b/org/mozilla/jss/pkcs11/PK11KeyWrapper.c
index 8580067e..5ff5a7cc 100644
--- a/org/mozilla/jss/pkcs11/PK11KeyWrapper.c
+++ b/org/mozilla/jss/pkcs11/PK11KeyWrapper.c
@@ -731,9 +731,9 @@ PK11SymKey *JSS_PK11_ImportSymKeyWithFlagsFIPS(PK11SlotInfo *slot, CK_MECHANISM_
      * 3. Unwrap into the token using the wrapping key.
      */
 
-#define FIPS_ENCRYPT_UNWRAP_ALGO CKM_AES_CBC_PAD
+#define FIPS_ENCRYPT_UNWRAP_ALGO CKM_AES_KEY_WRAP_PAD
 
-    wrapper = PK11_KeyGen(slot, FIPS_ENCRYPT_UNWRAP_ALGO, NULL, 32, wincx);
+    wrapper = PK11_KeyGen(slot, CKM_AES_KEY_GEN, NULL, 32, wincx);
     if (wrapper == NULL) {
         goto done;
     }

PR #681, opened 2021-03-04
Enable BETA flag

edewata:
The `BETA` flag in `JSSConfig.cmake` has been set to `1` to
enable some pre-release tests.

### DIFF
--- a/cmake/JSSConfig.cmake
+++ b/cmake/JSSConfig.cmake
@@ -1,8 +1,9 @@
 macro(jss_config)
     # Set the current JSS release number. Arguments are:
     #   MAJOR MINOR PATCH BETA
-    # When BETA is zero, it isn't a beta release.
-    jss_config_version(4 9 0 0)
+    # When BETA is 1, it is a pre-release (it enables some tests).
+    # When BETA is 0, it is a final release.
+    jss_config_version(4 9 0 1)
 
     # Configure output directories
     jss_config_outputs()


edewata:
Eventually we probably want to parse the version number from the spec file, and determine the `BETA` flag from the release number (i.e. anything with `-0.x` is a pre-release). Or, we might want to run all tests (including pre-release tests) at all times (if possible).

cipherboy (approved):
The pre-release test I'm thinking of right now is the `PKCS11Constants` one. We don't really want to run it after release, because it depends on NSS an JDK versions -- if a JDK or NSS update changes a constant, we don't really want to fail the build because of it, unless we also want to update our version of the constants. In general, it should be a harmless operation (perhaps, we want to build it at compile time from our current NSS sources) -- but because we can only check against the JDK in Java 8 (and JDK 9+ we don't have access to the internal values any more except via JDK source code), I think its better to do it prior to a release.

(E.g., in 8.2 we shipped some values and then later in 8.2.z NSS and JDK both updated changing some values to match the PKCS#11v3.0 spec version of the constants, which are the correct ones -- we could've updated our copy then but it didn't really matter and it causes unnecessary FTBFS failures).

My 2c. 

(And yes, I did mention about the `.spec` file changes in my TOI doc. You can probably set Beta based on `alpha||beta` in the release version if you wanted to. But I think some of this might change in the future if we get mavenized build system instead. Not sure how Maven handles templating.)

(We also need the version string for JDK Provider interface FYI).

edewata:
@cipherboy Thanks for the explanation! I added a link to this PR from the wiki so we don't lose it:
https://github.com/dogtagpki/jss/wiki/Testing-PKCS11Constants

PR #683, opened 2021-04-13
Convert KeyPairGeneratorSpi.Usage into enumeration

edewata:
The `KeyPairGeneratorSpi.Usage` class has been converted into
an enumeration that defines the same elements. These elements
store the values of the corresponding PKCS #11 constants.

The `getVal()` method has been deprecated, but it will continue
to work as before for backward compatibility.

The `opFlags` and `opFlagsMask` fields in `PK11KeyPairGenerator`
have been converted to `long` to match the PKCS #11 constants.
The values will be down-casted to `int` whenever needed.

The `opFlagForUsage` array in `PK11KeyPairGenerator` has become
redundant since the PKCS #11 constants can be obtained from
the enumeration elements, so it has been removed.

### DIFF
--- a/org/mozilla/jss/crypto/KeyPairGeneratorSpi.java
+++ b/org/mozilla/jss/crypto/KeyPairGeneratorSpi.java
@@ -10,6 +10,8 @@ import java.security.KeyPair;
 import java.security.SecureRandom;
 import java.security.spec.AlgorithmParameterSpec;
 
+import org.mozilla.jss.pkcs11.PKCS11Constants;
+
 public abstract class KeyPairGeneratorSpi {
 
     public KeyPairGeneratorSpi() {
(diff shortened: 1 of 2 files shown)


org/mozilla/jss/crypto/KeyPairGeneratorSpi.java
```diff
@@ -44,24 +46,34 @@ public abstract void initialize(AlgorithmParameterSpec params,
...
-        public static final Usage SIGN = new Usage(2);
-        public static final Usage SIGN_RECOVER = new Usage(3);
-        public static final Usage VERIFY = new Usage(4);
-        public static final Usage VERIFY_RECOVER = new Usage(5);
-        public static final Usage WRAP = new Usage(6);
-        public static final Usage UNWRAP = new Usage(7);
-        public static final Usage DERIVE = new Usage(8);
+    public enum Usage {
+
+        ENCRYPT        (PKCS11Constants.CKF_ENCRYPT),
```
cipherboy, line 51: I like the usage of `PKCS11Constants` :)

jmagne (approved):
This looks solid and much nicer. Approving.

edewata:
@cipherboy @jmagne Thanks!

PR #685, opened 2021-04-15
Convert SymmetricKey.Usage into enumeration

edewata:
The `SymmetricKey.Usage` class has been converted into an
enumeration that defines the same elements. These elements
store the corresponding PKCS #11 `CKF_` and `CKA_` values
although the `CKA_` values are not actually used.

The `getVal()` and `getPKCS11Constant()` method have been
deprecated, but they will continue to work as before for
backward compatibility.

The `opFlags` field in `PK11KeyGenerator` has been converted
to `long` to match the PKCS #11 value. The value will be
downcasted to `int` whenever needed.

The `opFlagForUsage` array in `PK11KeyGenerator` has become
redundant since the PKCS #11 value can be obtained from
the enumeration elements, so it has been removed.


jmagne (approved):
This looks good to me. Apparently we've also done some local testing with pki. Approving.

edewata:
@jmagne Thanks!

PR #686, opened 2021-04-20
Add CA installation test

edewata:
A `Dockerfile` has been added to define the builder image and test
runner image for JSS. The builder image is used to install the
build dependencies and then build JSS. The test runner image is
used to install JSS with its runtime dependencies then run tests.

A new test has been added to install DS and CA with the newly
built JSS packages.

Some scripts have been added to initialize the test runner, to
create and remove DS, and to save the test artifacts.

### DIFF
--- a/.github/workflows/pki-tests.yml
+++ b/.github/workflows/pki-tests.yml
@@ -1,0 +1,108 @@
+name: PKI Tests
+
+on: [push, pull_request]
+
+jobs:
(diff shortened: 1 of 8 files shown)


jmagne (approved):
Looks good, tests are passing. Approving.

edewata:
@jmagne Thanks!

PR #687, opened 2021-04-30
Disable failing tests

edewata:
The CentOS 7 and 8 tests have been disabled since they keep
failing due to a missing dependency.

The optional test workflow have been merged into the required
test workflow. All tests should now pass. This will make it
easier to determine if a commit has broken something.


.github/workflows/required.yml
```diff
@@ -13,8 +13,15 @@ jobs:
           - 'fedora_33'
           - 'fedora_latest_jdk11'
           - 'symbolcheck'
+          - 'debian_jdk11'
+          - 'ubuntu_jdk8'
+          # Disable tests due to missing dependencies
+          # - 'centos_7'
+          # - 'centos_8'
```
cipherboy, line 20: Aww they still haven't gotten an updated cmake version? :'(

edewata: Yeah, I have no idea what the process is for CentOS 7 & 8. We're working on CentOS 9 though, so we can make sure it's working there.

edewata: BTW, there seems to be an issue with pkcs11 constants again. I'm trying to investigate it.

jmagne (approved):
Approving this because it cleans up many of the current ci errors. I know there are some more to investigate, but these appear to be a separate issue.

edewata:
@jmagne Thanks!
@cipherboy Are you OK if I merge this? We can certainly reenable them again when they update cmake.

edewata:
@cipherboy Per IRC discussion I'm merging this PR. Thanks!

PR #688, opened 2021-04-30
Move pkcs11check into Fedora 33 test

edewata:
Currently the PKCS #11 constants are only tested on Rawhide
and it's failing, but it's actually working fine on Fedora 33.
For now the test has been moved into the Fedora 33 test so the
CI will pass, but later the tests will need to be expanded to
include newer Fedora versions. That way the PKCS #11 constants
will be tested separately on each platform, so it will be
easier to isolate the problem.


jmagne (approved):
This makes sense. Just moving stuff from rawhide to f33. Approving.

cipherboy:
Note that the NSS team usually rebases stuff frequently across most all releases. So if things are failing in Rawhide, it is an indication it'll change in Fedora 33 as well soon enough.

Also F34 is the current most-recently supported & released distro. :)

(But don't let that hold this up if you want to merge it as-is). Just comments from the peanut gallery ;-)

edewata:
@jmagne @cipherboy Thanks!
We need the F33 test as a baseline. I'll clone it to F34 and Rawhide next. If Rawhide fails but F33/F34 works, at least that can show us that it's not caused by us.

PR #690, opened 2021-05-05
Add build.sh --without-test option

edewata:


jss.spec
```diff
@@ -119,7 +128,10 @@ modutil -dbdir /etc/pki/nssdb -chkfips true | grep -q enabled && export FIPS_ENA
 cd %{_vpath_builddir}
 %{__make} all
 %{__make} javadoc
+
+%if %{with_test}
```
ladycfu, line 132: OK with my limited knowledge in these script languages, I think I understand all up until here.  Does %if %(with_test) means it will check for the "test" value?  I guess I am not familiar with the syntax to see if "%(with_xxx) is what I think it does.  Of course, overall, if it is tested to work then it must be right.  Thanks.

edewata: Yes, `%bcond_without test` means the `test` variable is enabled by default (see [this doc](https://rpm.org/user_doc/conditional_builds.html)). If you specify `--without test` the `test` variable will be disabled so the test will not be executed. I have tested it to work as expected.

ladycfu (approved):

jss.spec
edewata: As discussed on IRC, the `build.sh` is actually replacing `%bcond_without test` with a `%global with_test 0` or `%global with_test 1` depending on whether you specify a `--without-test` or not. This is indeed a bit confusing, but it works as expected, so I'll merge it for now. Perhaps we can clean it up later. Thanks!

PR #691, opened 2021-05-10
Enable DirStrConverterTest

edewata:


edewata:
The test was probably disabled because it was broken: https://github.com/dogtagpki/jss/commit/c15418a714ba2ea0b99f61c5ae301a064c904913

PR #692, opened 2021-05-11
Add pom.xml and test for Maven

edewata:
This PR adds a basic `pom.xml` to build JSS and run the unit tests using Maven on Java 11.

A GitHub workflow is added to execute the build and the tests automatically on every check-in.

There is an existing PR (#636) by @cipherboy but it's not ready to be merged as is. Basically we're breaking it into smaller PRs and incrementally incorporate the elements from that PR. Once all elements are merged we can close it too.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.dogtagpki.jss</groupId>
+    <artifactId>jss</artifactId>
+    <version>4.9.0-SNAPSHOT</version>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>javax.xml.bind</groupId>
+            <artifactId>jaxb-api</artifactId>
+            <version>2.2.12</version>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.13.2</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+            <version>3.11</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.25</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-jdk14</artifactId>
+            <version>1.7.25</version>
+            <scope>runtime</scope>
+        </dependency>
+
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.1</version>
+                <configuration>
+                    <release>11</release>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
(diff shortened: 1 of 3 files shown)


ckelleyRH (approved):
I was able to build and test JSS using this pom with OpenJDK11, so I think this is good.

pom.xml
```diff
@@ -0,0 +1,60 @@
...
+
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.1</version>
+                <configuration>
```
ckelleyRH, line 53: We might want to consider adding `<outputDirectory>~/build/pki/JARS</outputDirectory>` into the `<configuration>` tag, then the build products are outside the repo. The path I chose was just to illustrate the point, it could be wherever as long as it's outside the repo.

ckelleyRH (changes requested):
Actually one point, the build location is `$PATH/git/jss/target/jss-4.9.0-SNAPSHOT.jar` - this puts it inside the remit of git. I only just noticed after I reviewed, by switching back to my `master` branch and having a thousand new files waiting to be staged!

pom.xml
edewata: I don't have much experience with Maven, so I don't have a strong opinion about this, but someone [said](https://stackoverflow.com/questions/23053576/combining-output-of-maven-modules-and-parent):
> Maven projects should never, ever write outside their project's folder!


According to [this](https://stackoverflow.com/questions/4757426/maven-specify-the-outputdirectory-only-for-packaging-a-jar) we can change the output folder with a command line parameter:
```
$ mvn ... -DoutputDirectory=<path>
```
I'm thinking we can continue to use the `build.sh` as the primary build script, and internally it will call Maven with `-DoutputDirectory=~/build/...`, but if you call Maven directly it will build inside the project folder.

I also added `target/` into `.gitignore`, so once we merge this PR git shouldn't see the build artifacts anymore. We could make the same changes to `.gitignore` in other branches that we're still maintaining to avoid that problem.

What do you think?

ckelleyRH: OK, I didn't realise there was this Maven convention of putting everything in `project/`. At my last place we built into all sorts of other places with Maven, but I think we should follow the convention if we're starting from scratch.

In that case, I think your plan to use `build.sh` as a wrapper for `mvn package` (or whatever that command eventually bloats out to be) is sound.

Yes, the `.gitignore` addition is critical if we're going to keep building into the repo! I don't know how often we are going to be switching to old branches, but I expect that the answer is "all the time". I think we should definitely add the `.gitignore` change to all active branches as:
1. On first encountering this, it is going to be confusing for the dev - and then they're going to waste time trying to figure it out.
2. Subsequently, this will just become very annoying, and people will locally commit some workaround to get around it, so we might as well do it for them and save all the hassle.

ckelleyRH (changes requested):

pom.xml
ckelleyRH: One objectionable "advantage" of _not_ using the `.gitignore` is you can version control the jars too. I don't think we want to go down this road given the objections to bundling binaries in our discussions around the future supply chain, I just wanted to highlight that I've seen it done and it's a maintenance nightmare to be avoided at all costs!

edewata: Agreed. I can add the code to call `mvn` from `build.sh` in a separate PR later (right now it's calling `rpmbuild`), also update `.gitignore` for old branches in separate PRs. For this PR is there anything I need to change?

Just to clarify, we're not planning to bundle JARs in the source git (although some other projects might do that), but the rules for dist-git are different so I'm trying to explore what we can/cannot do there.

ckelleyRH: I've been keeping quiet on the rules email chain because I don't know any of the current rules! It's been interesting to follow though, learning a lot from that discussion.

I agree that the refactoring of `build.sh` is work for another PR. I think the other `.gitignore`s have to be separate PRs anyway, because they're different branches? If so then I think your work here is done :1st_place_medal:

ckelleyRH (approved):

ckelleyRH:
Sorry about random messages, it took a few minutes to figure out how to approve when I had initially requested changes. I was expecting a big LGTM button, but it was a bit more subtle :smile:

ckelleyRH:
LGTM.

edewata:
@ckelleyRH Thanks! I created PR #694 and #695 to update `.gitignore` in 4.8 and 4.4 branches. I think those are the branches that we're still maintaining. I updated the `.gitignore` in this PR too for consistency. Please take a look.

PR #695, opened 2021-05-12
Update .gitignore for JSS 4.4

edewata:
There is no CI for this branch.

**Edit:** Apparently there are CI tests in this branch, but they run under a specific [branch name](https://github.com/dogtagpki/jss/blob/v4.4.x/.github/workflows/required-v4.4.x.yml#L5-L9) only. Thanks to @cipherboy, I've updated the workflow to run under all branch names, and dropped the outdated tests.

### DIFF
--- a/.github/workflows/required-v4.4.x.yml
+++ b/.github/workflows/required-v4.4.x.yml
@@ -1,28 +1,8 @@
 name: Required Tests for v4.4.x branch
 
-on:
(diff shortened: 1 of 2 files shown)


ckelleyRH (approved):
LGTM thanks!

edewata:
@ckelleyRH Thanks!

PR #689, opened 2021-05-03
Add PKCS11Constants test for Fedora 34

edewata:
This PR depends on PR #740.

Currently the PKCS #11 constants test is failing on Fedora 34 and Rawhide:
```
/tmp/tmp-jss-pkcs11constants-zbw7io_w/test.c: In function ‘main’:
/tmp/tmp-jss-pkcs11constants-zbw7io_w/test.c:6:17: error: ‘CKF_EC_FP’ undeclared (first use in this function); did you mean ‘CKF_EC_F_P’?
    6 |             if (CKF_EC_FP != 0x00100000) {
      |                 ^~~~~~~~~
      |                 CKF_EC_F_P
/tmp/tmp-jss-pkcs11constants-zbw7io_w/test.c:6:17: note: each undeclared identifier is reported only once for each function it appears in



    main()
  File "/home/sandbox/jss/./tools/build_pkcs11_constants.py", line 806, in main
    check_references(objs)
  File "/home/sandbox/jss/./tools/build_pkcs11_constants.py", line 610, in check_references
    obj.check_output(nss_args)
  File "/home/sandbox/jss/./tools/build_pkcs11_constants.py", line 229, in check_output
    raise Exception(("Unknown error with symbol '%s': cc ret: %d " +
Exception: Unknown error with symbol 'CKF_EC_FP': cc ret: 1 @ /tmp/tmp-jss-pkcs11constants-zbw7io_w; command: cc -o /tmp/tmp-jss-pkcs11constants-zbw7io_w/test.exe /tmp/tmp-jss-pkcs11constants-zbw7io_w/test.c -I/usr/include/nss3 -I/usr/include/nspr4 -lssl3 -lsmime3 -lnss3 -lnssutil3 -lplds4 -lplc4 -lnspr4 -lpthread -ldl
Container run exited with status: 1
Error: Process completed with exit code 1.
```

PR #697, opened 2021-05-13
Make GitHub workflow use F34, drop EOL F32

ckelleyRH:


ckelleyRH:
I created a `fedora_34` dockerfile, it's a carbon coby of the `fedora_33` dockerfile with `s/33/34`.

edewata:
For `required.yml` there's still a problem with the `PKCS11Constant` test (see PR #689), so you might want to remove lines 32-41 for now, then I can rebase PR #689.

ckelleyRH:
OK, thanks, I just found that PR - hadn't realised you'd already started on this, sorry!

edewata (approved):
No problem at all, this is good :)
So we isolate the problem, and continue to make progress everywhere else.

PR #699, opened 2021-05-18
Enable SonarCloud checking of new code

ckelleyRH:
Create a new CI job called "build" which builds from source using Maven,
then runs the Sonar analysis job. The quality gate is set to be advisory
only so it never fails, to prevent error fatigue. The intended use is
"free robotic review" of code quality, rather than any enforcement.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,34 @@
+name: Build
+on: [push, pull_request]
+jobs:
+  build:
+    name: Code analysis
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+      - name: Set up JDK 11
+        uses: actions/setup-java@v1
+        with:
+          java-version: 11
+      - name: Cache SonarCloud packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.sonar/cache
+          key: ${{ runner.os }}-sonar
+          restore-keys: ${{ runner.os }}-sonar
+      - name: Cache Maven packages
+        uses: actions/cache@v1
+        with:
+          path: ~/.m2
+          key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
+          restore-keys: ${{ runner.os }}-m2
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,39 @@
+name: Build
```
edewata, line 1: Could we call it "Code Analysis" or something like that? If we have other tools later we can add it to the same workflow.

.github/workflows/build.yml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    types: [opened, synchronize, reopened]
```
edewata, line 7: We usually use this trigger in other workflows:
https://github.com/dogtagpki/jss/blob/master/.github/workflows/required.yml#L3
Should we use the same trigger, or is there a specific reason for this?

.github/workflows/build.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  build:
+    name: Build
```
edewata, line 10: Maybe we could use the tool's name (i.e. SonarCloud) as the job name here.

edewata (approved):
I have some comments below, but it looks good otherwise. Feel free to update if necessary, then merge. Thanks!

.github/workflows/build.yml
ckelleyRH: Good idea, done.

.github/workflows/build.yml
ckelleyRH: I cargo-culted the whole YML file, if there is a convention then I will follow it.

.github/workflows/build.yml
edewata: Actually I was suggesting to use "Code Analysis" in line 1 so it will appear as one of the test categories here:
https://github.com/dogtagpki/jss/actions
Right now it's showing "Build" which doesn't really describe the jobs in this workflow.

.github/workflows/build.yml
edewata: Thanks for the update! It's not really a convention, but we want to run the CI on all pushes & pull requests without any branch restrictions, so that's what we've been using.

PR #701, opened 2021-05-19
Remove deprecated Integer constructor from IPAddressName

ckelleyRH:


ckelleyRH:
@edewata. I replaced a variable declaration `int something` with `var something`, which is legitimate in Java 10+, and some of the pipeline failed because they're using Java 1.8

Allowing it to be `int` again so the CI passes, but should we update/remove some of these older jobs for the master branch so we can make use of the features of Java 11?

src/main/java/org/mozilla/jss/netscape/security/x509/IPAddressName.java
```diff
@@ -209,8 +209,7 @@ public static int fillIPv4Address(String s, byte[] address, int start) {
         try {
             int end = start + nt;
             for (int i = start; i < end; i++) {
-                Integer j = new Integer(Integer.valueOf(st.nextToken()));
-                address[i] = (byte) j.intValue();
+                address[i] = Byte.valueOf(st.nextToken());
```
edewata, line 212: Here we're parsing IPv4 values in the range of `0..255`, but Java's `byte` range is `-128..127`. I suspect the original code was trying to avoid a `NumberFormatException`, so it's parsing the token into an `Integer` first, then downcast it into `byte`. We might be able to do this instead:
```
// parse token into Integer then downcast into byte to avoid NumberFormatException
address[i] = (byte) Integer.parseInt(st.nextToken());
```
Should we do this change separately since it requires further inspection?
Maybe we can add a unit test for this too.

Btw, I think we should use `parseInt/parseByte` instead of `valueOf` if the result that we want is a primitive.

src/test/java/org/mozilla/jss/tests/JSSE_SSLClient.java
```diff
@@ -529,7 +529,7 @@ public static void main(String [] args) {
                 keystoreLocation = args[0];
             }
             if ( args.length >= 2) {
-                testPort         = new Integer(args[1]).intValue();
+                testPort         = Integer.valueOf(args[1]);
```
edewata, line 532: I think we should use `parseInt()` here.

src/test/java/org/mozilla/jss/tests/JSS_FileUploadClient.java
```diff
@@ -333,7 +333,7 @@ public static void main(String[] args) {
                 System.out.println(usage);
                 System.exit(1);
             } else {
-                socketCntr = new Integer(args[0]).intValue();
+                socketCntr = Integer.valueOf(args[0]);
```
edewata, line 336: I think we should use `parseInt()` here too.

src/test/java/org/mozilla/jss/tests/JSS_SelfServClient.java
```diff
@@ -919,15 +919,15 @@ public static void main(String[] args) {
                 System.out.println(usage);
                 System.exit(1);
             } else {
-                numOfThreads = new Integer(args[0]).intValue();
+                numOfThreads = Integer.valueOf(args[0]);
```
edewata, line 922: Should be `parseInt()` as well.

src/test/java/org/mozilla/jss/tests/JSS_SelfServServer.java
```diff
@@ -113,7 +112,7 @@ public void doIt(String[] args) throws Exception {
             if (args[3].equalsIgnoreCase("true") == true)
                 TestInetAddress = true;
             if (args.length >= 5)
-                port = new Integer(Integer.valueOf(args[4]));
+                port = Integer.valueOf(args[4]);
```
edewata, line 115: Should be `parseInt()` as well.

src/test/java/org/mozilla/jss/tests/SSLClientAuth.java
```diff
@@ -89,38 +115,38 @@ public static Certificate makeCert(String issuerName, String subjectName,
...
         CryptoToken tok = cm.getInternalKeyStorageToken();
-        
+
         PasswordCallback cb = new FilePasswordCallback(args[1]);
         tok.login(cb);
-        
+
         if (args.length >= 3) {
-            port = new Integer(args[2]).intValue();
+            port = Integer.valueOf(args[2]);
```
edewata, line 144: Should be `parseInt()` as well.

src/test/java/org/mozilla/jss/tests/SSLClientAuth.java
```diff
@@ -459,15 +485,15 @@ static Extension makeBasicConstraintsExtension() throws Exception {
...
-    
+
     static int nextRandInt(SecureRandom rand) throws Exception {
         int i;
         byte[] bytes = new byte[4];
         rand.nextBytes(bytes);
-        i =  ((int)bytes[0])<<24 | ((int)bytes[1])<<16 |
-                ((int)bytes[2])<<8 | ((int)bytes[3]);
+        i =  (bytes[0])<<24 | (bytes[1])<<16 |
+                (bytes[2])<<8 | (bytes[3]);
```
edewata, line 494: I think the parentheses can be removed too, but actually there might be a flaw in this code. Java automatically promotes `byte` to `int` before the bitwise shift and or operations, so if one the byte values are negative, the sign bit would be extended, and it will override other values, thus reducing the entropy.

Let's do this change separately and probably add a test too.

edewata (changes requested):
Please see my comments below. I think we should move some of the changes into separate PRs because they need additional inspection.

ckelleyRH:
I'm going to break this up into a few PRs.

ckelleyRH:
I kept the change for `IPAddressName` on this PR, as that had the most discussion, I'll make another PR for the others and look at the `valueOf/parseInt` suggestions.

The byte->int casting I'm going to leave for now, as it's not really logically part of this change.

ckelleyRH:
I haven't added the test yet, obviously!

edewata (approved):
@ckelleyRH Sorry, I thought you were still working on the test. It's not required for this PR and all other issues have been addressed separately. ACK.

ckelleyRH:
@edewata thanks!

PR #702, opened 2021-05-20
Access static fields/methods statically, not via instances

ckelleyRH:


edewata (approved):
LGTM. There are some unused vars, but we need to be a bit careful before changing JSS API since there are some other projects using this library.

ckelleyRH:
Yeah, "unused" needs taking with a pinch of salt anyway:

- Unused, except it's actually used by Spring
- Unused, except it's actually used by a Jython call
- Unused, except it's actually used in an XML plugin definition

I've been burned by other lies told to me by Eclipse, these are just the ones I remember :laughing: :fire:

Issue #705, opened 2021-05-21
Add in missing annotations


ckelleyRH:
Deprecated methods should be annotated so they can be easily identified and removed later.
Overridden methods should be annotated to benefit from compile-time checking of interface contracts.

cipherboy:
@ckelleyRH Just a heads up, you've added the "Resolves:" keyword to all of the above PRs.. Maybe "related to" works better as they all need to merge before the issue is resolved (and presently the first PR merged will close this issue). Doesn't really matter this time as you've opened all the PRs right away, but for any other efforts, something to know for the future :)

ckelleyRH:
@cipherboy Ah I see, thanks for the tip!

PR #707, opened 2021-05-21
Add missing annotations to org.mozilla.jss.util

ckelleyRH:
Resolves #705.

### DIFF
--- a/src/main/java/org/mozilla/jss/util/Base64OutputStream.java
+++ b/src/main/java/org/mozilla/jss/util/Base64OutputStream.java
@@ -9,159 +9,158 @@
 package org.mozilla.jss.util;
 
 import java.io.FilterOutputStream;
-import java.io.PrintStream;
 import java.io.IOException;
+import java.io.PrintStream;
 
 /**
  * An output stream filter that takes arbitrary bytes and outputs their
- * base64 encoding.  Call flush() or close() to write out the final padding.
+ * base64 encoding. Call flush() or close() to write out the final padding.
  * The class also automatically puts line breaks in the output stream.
  */
 public class Base64OutputStream extends FilterOutputStream {
 
-	/**
-	 * Create a stream that does not insert line breaks.  To have line
-	 * breaks, use the other constructor.
-	 */
-	public Base64OutputStream(PrintStream out) {
-		super(out);
-		charsOnLine = 0;
-		inputCount = 0;
-		input = new byte[3];
-		doLineBreaks = false;
-	}
+    /**
+     * Create a stream that does not insert line breaks. To have line
+     * breaks, use the other constructor.
+     */
+    public Base64OutputStream(PrintStream out) {
+        super(out);
+        charsOnLine = 0;
+        inputCount = 0;
+        input = new byte[3];
+        doLineBreaks = false;
+    }
 
-	/**
-	 * @param quadsPerLine Number of 4-character blocks to write before
-	 * 	outputting a line break.  For example, for 76-characters per line,
-	 *	pass in 76/4 = 19.
-	 */
-	public Base64OutputStream(PrintStream out, int quadsPerLine) {
-		this(out);
-		doLineBreaks = true;
-		assert(quadsPerLine > 0);
-		charsPerLine = quadsPerLine*4;
-	}
-		
+    /**
+     * @param quadsPerLine Number of 4-character blocks to write before
+     *            outputting a line break. For example, for 76-characters per line,
+     *            pass in 76/4 = 19.
+     */
+    public Base64OutputStream(PrintStream out, int quadsPerLine) {
+        this(out);
+        doLineBreaks = true;
+        assert (quadsPerLine > 0);
+        charsPerLine = quadsPerLine * 4;
+    }
 
-	@Override
+    @Override
     public void write(int oneByte) throws IOException {
-		input[inputCount++] = (byte)(0xff & oneByte);
-		if(inputCount==3) {
-			outputOneAtom();
-			if(doLineBreaks) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/mozilla/jss/util/NativeErrcodes.java
```diff
@@ -246,12 +246,13 @@
...
      * @deprecated As of NSS 3.11, FORTEZZA is no longer supported.
-     * XP_SEC_FORTEZZA_BAD_CARD, XP_SEC_FORTEZZA_NO_CARD, 
+     * XP_SEC_FORTEZZA_BAD_CARD, XP_SEC_FORTEZZA_NO_CARD,
      * XP_SEC_FORTEZZA_NONE_SELECTED, XP_SEC_FORTEZZA_MORE_INFO
      * XP_SEC_FORTEZZA_PERSON_NOT_FOUND, XP_SEC_FORTEZZA_NO_MORE_INFO
-     * XP_SEC_FORTEZZA_BAD_PIN and XP_SEC_FORTEZZA_PERSON_ERROR are 
+     * XP_SEC_FORTEZZA_BAD_PIN and XP_SEC_FORTEZZA_PERSON_ERROR are
      * placeholder for backward compatibility.
      */
+    @Deprecated
```
cipherboy, line 255: Wouldn't this `@Deprecated` warning also need to be added on the later elements as well? If not, wouldn't you need to mark the `SEC_ERROR` and beyond as not deprecated?

src/main/java/org/mozilla/jss/util/NativeProxy.java
```diff
@@ -132,6 +134,7 @@ public int hashCode() {
      * @deprecated finalize() in Object has been deprecated. Use close(...)
      * from the AutoCloseable interface instead.
      */
+    @Override
```
cipherboy, line 137: My understanding was that finalize wasn't strictly an override in all versions of JDK? I thought in the future (e.g., JDK17?) they were removing the finalize member on Object. Maybe I'm misremembering and they've not made plans to fully remove it? Might be worth looking into.

(Sorry, comment was on https://github.com/dogtagpki/jss/pull/724/files#r636855103 because that's the first PR I noticed this on).

src/main/java/org/mozilla/jss/util/NativeErrcodes.java
edewata: I think Java requires a separate deprecation annotation for each constant, so the other constants aren't actually deprecated yet. We probably should do that in a separate PR though.

src/main/java/org/mozilla/jss/util/NativeProxy.java
edewata: To my understanding there are issues with `finalize()` in general so it's no longer recommended to use and we should replace it before it's actually removed in the future. Right now we're still relying on the method as an extension of the `Object.finalize()`, so the `@Override` is simply affirming that dependency (and making sure that it's named correctly).

src/main/java/org/mozilla/jss/util/Base64OutputStream.java
```diff
@@ -44,7 +44,8 @@ public Base64OutputStream(PrintStream out, int quadsPerLine) {
 	}
 		
 
-	public void write(int oneByte) throws IOException {
+	@Override
```
edewata, line 47: Not sure if this is already mentioned somewhere else, but it looks like this file uses tabs for indentation. I think generally we want to use 4 spaces for indentation, but this probably can be addressed separately later.

src/main/java/org/mozilla/jss/util/NativeProxy.java
ckelleyRH: I've not seen a hard plan to get rid of `Object.finalize()` but as you say it's bad form to depend on it, we should definitely consider not doing this.

src/main/java/org/mozilla/jss/util/NativeErrcodes.java
ckelleyRH: I am making a separate PR to add in these additional annotations. This keeps the situation where we have not deprecated something when we should logically distinct from this PR, where we're simply adding annotations to things we have already deprecated,

src/main/java/org/mozilla/jss/util/NativeProxy.java
cipherboy: I believe it will be eventually removed in favor of https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Cleaner.html

src/main/java/org/mozilla/jss/util/Base64OutputStream.java
ckelleyRH: I added in an extra commit that does Eclipse auto-formatting to get rid of any tabs.

src/main/java/org/mozilla/jss/util/NativeErrcodes.java
ckelleyRH: See #736

ckelleyRH:
This looks very missing:

`curl https://registry.fedoraproject.org/v2/`
`{}`

I did this a few minutes ago and got a blob of JSON so looks like an intermittent failure.

edewata (approved):
LGTM. Sonar reported a missing `hashCode()`, but I think that's an existing issue. Feel free to add it or just merge as is. Thanks!

ckelleyRH:
Yeah it's a bit annoying it can't do a diff with `master HEAD` and say "there were 30 issues, and you've resolved 29, nice work!" rather than "there's a bug, that was already there". I'll do it as a separate change.

PR #708, opened 2021-05-21
Add missing annotations to org.mozilla.jss.ssl.javax

ckelleyRH:
Resolves #705.


ckelleyRH:
I thought this might happen, might have a case here of methods with the same name but different signatures. Needs another look.

cipherboy:
@ckelleyRH 
> I thought this might happen, might have a case here of methods with the same name but different signatures. Needs another look.

Is this because of the failing JDK8 test? I didn't add `@Override` because the ones in the failing don't strictly override: depending on JDK versions, these features just haven't been backported (into older JDK versions) and thus aren't actually an override there.

You'd need conditional annotations based on versions... :)

ckelleyRH:
I've updated that Java 8 workflow in #739 to use Java 11 so should pass this time.

edewata (approved):
LGTM

PR #709, opened 2021-05-21
Add missing annotations in org.mozilla.jss.ssl

ckelleyRH:
Resolves #705.

### DIFF
--- a/src/main/java/org/mozilla/jss/ssl/SSLClient.java
+++ b/src/main/java/org/mozilla/jss/ssl/SSLClient.java
@@ -24,17 +24,17 @@ import org.mozilla.jss.crypto.AlreadyInitializedException;
 /**
  * Parameters supported by this socket test:
  *
- * filename 	file to be read from https server (default: /index.html)
- * port 	port to connect to (default: 443)
- * ipaddr 	address to connect to (overrides hostname, no default)
- * hostname 	host to connect to (no default)
- * clientauth 	do client-auth or not (default: no client-auth)
+ * filename file to be read from https server (default: /index.html)
+ * port port to connect to (default: 443)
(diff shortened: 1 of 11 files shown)


src/main/java/org/mozilla/jss/ssl/SSLClient.java
```diff
@@ -439,7 +439,8 @@ public static void main(String argv[])  throws Exception
     this.sc = sc;
   }
 
-  public void handshakeCompleted(SSLHandshakeCompletedEvent event) {
+  @Override
+public void handshakeCompleted(SSLHandshakeCompletedEvent event) {
```
cipherboy, line 443: indentation? :)

src/main/java/org/mozilla/jss/ssl/TestCertApprovalCallback.java
```diff
@@ -16,7 +16,8 @@
 public class TestCertApprovalCallback
    implements SSLCertificateApprovalCallback {
 
-	public boolean approve(
+	@Override
```
cipherboy, line 19: I think you're mixing tabs and spaces :/

ckelleyRH:
Sigh. I pushed a second commit on top of the annotations change which is just Eclipse running auto-format on the files I changed.

edewata (approved):
LGTM. The bug found by Sonar seems to be a false positive, so feel free to ignore it.

ckelleyRH:
I agree that the "bug" is not a bug, as we do eventually do `s1.equals(s2)` for value comparison rather than object comparison.

I think having both `s1.equals(s2)` and `s1 ==s2` is confusing though, it forces the dev to sit and think when all the method is intending is returning `okay` if `s1.equals(s2)` and `failed` otherwise.

I'll remove it in a further PR.

PR #724, opened 2021-05-21
Add missing annotations to org.mozilla.jss.nss

ckelleyRH:
Resolves #705.

### DIFF
--- a/src/main/java/org/mozilla/jss/nss/BufferProxy.java
+++ b/src/main/java/org/mozilla/jss/nss/BufferProxy.java
@@ -11,10 +11,12 @@ public class BufferProxy extends org.mozilla.jss.util.NativeProxy {
      *
      * But this does it for you.
      */
+    @Override
     protected void releaseNativeResources() {
         Buffer.Free(this);
     }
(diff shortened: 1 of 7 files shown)


src/main/java/org/mozilla/jss/nss/BufferProxy.java
```diff
@@ -11,10 +11,12 @@ public BufferProxy(byte[] pointer) {
      *
      * But this does it for you.
      */
+    @Override
     protected void releaseNativeResources() {
         Buffer.Free(this);
     }
 
+    @Override
```
cipherboy, line 19: My understanding was that `finalize` wasn't strictly an override in all versions of JDK? I thought in the future (e.g., JDK17?) they were removing the finalize member on `Object`. Maybe I'm misremembering and they've not made plans to fully remove it? Might be worth looking into.

cipherboy: Ah, wrong PR.

edewata (approved):

PR #729, opened 2021-05-21
Add missing annotations to org.mozilla.jss.crypto

ckelleyRH:
Resolves #705.

### DIFF
--- a/src/main/java/org/mozilla/jss/crypto/Signature.java
+++ b/src/main/java/org/mozilla/jss/crypto/Signature.java
@@ -18,177 +18,178 @@ import java.security.spec.AlgorithmParameterSpec;
  */
 public class Signature {
 
-	protected Signature() { }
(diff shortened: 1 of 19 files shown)


src/main/java/org/mozilla/jss/crypto/Signature.java
```diff
@@ -183,7 +183,8 @@ public SignatureAlgorithm getAlgorithmID() {
 	/**
 	 * Cloning is not supported yet
 	 */
-	protected Object clone() throws CloneNotSupportedException {
+	@Override
```
cipherboy, line 186: Tabs versus spaces here too. :)

cipherboy: (Rather the next line, sorry).

edewata (approved):
LGTM

PR #734, opened 2021-05-21
Fix return from finally in JSSSocketChannel

cipherboy:
Sonar reports the following on the (now removed) return statement:

> Remove this return statement from this finally block.
>
> Using `return`, `break`, `throw`, and so on from a `finally` block suppresses
> the propagation of any unhandled `Throwable` which was thrown in the
> `try` or `catch` block.

Re-do the logic to fix this bug.

`Signed-off-by: Alexander Scheel <alex.scheel@canonical.com>`

---

Note that this is likely a popular path as most sockets we construct will be `Socket` instances and not `SocketChannel`s.


ckelleyRH (approved):
LGTM cheers!

PR #735, opened 2021-05-21
Add missing synchronized keywords to Sockets

cipherboy:
Due to the new `@Override` method, Sonar has caught missing synchronized
keywords on several `JSSSocket` and `JSSServerSocket` methods.

> Make this method "`synchronized`" to match the parent class
> implementation.
>
> When `@Overrides` of `synchronized` methods are not themselves
> `synchronized`, the result can be improper synchronization as callers
> rely on the thread-safety promised by the parent class.

`Signed-off-by: Alexander Scheel <alex.scheel@canonical.com>`


ckelleyRH (approved):
LGTM thanks!

PR #736, opened 2021-05-24
Deprecate constants in NativeErrcodes

ckelleyRH:
A comment here declares Fortezza deprecated, but the Javadoc deprecation
only applies to the first constant, so deprecate the rest too.

### DIFF
--- a/src/main/java/org/mozilla/jss/util/NativeErrcodes.java
+++ b/src/main/java/org/mozilla/jss/util/NativeErrcodes.java
@@ -246,19 +246,27 @@ public class NativeErrcodes {
     public static final int SEC_ERROR_DECRYPTION_DISALLOWED = 228;
     /**
      * @deprecated As of NSS 3.11, FORTEZZA is no longer supported.
-     * XP_SEC_FORTEZZA_BAD_CARD, XP_SEC_FORTEZZA_NO_CARD, 
+     * XP_SEC_FORTEZZA_BAD_CARD, XP_SEC_FORTEZZA_NO_CARD,
      * XP_SEC_FORTEZZA_NONE_SELECTED, XP_SEC_FORTEZZA_MORE_INFO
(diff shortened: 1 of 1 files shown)


edewata:
I'll defer to @cipherboy, @jmagne, or @ladycfu on this area.

cipherboy (approved):
All Fortezza has been deprecated by NSS and could probably be removed if NSS doesn't expose them any more either.

ckelleyRH:
@cipherboy I read this interesting piece of history: https://bugzilla.mozilla.org/show_bug.cgi?id=668397

It seems that Fortezza has been dead for 25 years, and the NSS maintainers long ago stripped the Fortezza support, so I guess it can go?

cipherboy:
Works for me, :) Happy to remove all fortezza support then too.

ckelleyRH:
Great! I made a commit doing just this but for some reason everything I push, no matter how trivial, is breaking the CI. One of those days maybe. Will investigate tomorrow.

cipherboy:
Happy to take a look if you share some logs... Maybe I've run across it before :)

ckelleyRH:
That'd be great, thanks! Commit is on this branch in my fork: https://github.com/ckelleyRH/jss/tree/fortezza

cipherboy:
@ckelleyRH Ah I see. See:

```
PR_STATIC_ASSERT(PR_ARRAY_SIZE(kea_alg_defs) == ssl_kea_size);
```

It is asserting that `kea_alg_defs` aligns to `ssl_kea_size`. Since Fortezza is [still defined](https://github.com/nss-dev/nss/blob/master/lib/ssl/sslt.h#L75-L86) in the NSS struct, you can't remove it from that particular struct (in the C layer) though you could remove it from the Java layer and prevent using it in the JNI layer.

cipherboy:
(this is a compile time assertion hence why it is failing to build)

PR #738, opened 2021-05-25
Fix some Sonar-discovered issues

ckelleyRH:


ckelleyRH:
See #709 for context.

ckelleyRH:
It would be nice for Sonar to acknowledge that I've fixed over 50 issues here, rather than claiming I've created 39, but no good deed goes unpunished.

ckelleyRH:
Builds failing because something in the Docker image is requesting Java 8 compliance for the source and target, and I used the `var` keyword:

`#12 2.294 -- JSS JAVAC FLAGS: -classpath;/usr/share/java/slf4j/api.jar:/usr/share/java/apache-commons-lang3.jar:/usr/share/java/jaxb-api.jar;-sourcepath;/tmp/jss/build/BUILD/jss-4.9.0-alpha1/src/main/java;-target;1.8;-source;1.8;-O`

Should fix this.

ckelleyRH:
Requires #739 to pass the CI.

src/main/java/org/mozilla/jss/ssl/SSLClient.java
```diff
@@ -102,13 +102,9 @@ private String getArgument(String key) {
...
     private String cmp(String s1, String s2) {
-        if (s1 == s2)
-            return okay;
-        if (s1 == null)
+        if (s1 == null || !s1.equals(s2))
             return failed;
-        if (s1.equals(s2))
-            return okay;
-        return failed;
+        return okay;
```
edewata, line 107: Hmm.. in the original code if both `s1` and `s2` are `null` it would return `okay`, but not in the new code. Is that intentional?

ckelleyRH: Good spot. I think that's a bug though? We check `s1 == null` and fail if it is, so presumably the intended behaviour of `s1=s2=null` is to return `failed`?

edewata: Looks like everywhere `cmp()` is used, the first parameter is guarded by `isInvalid()`, so it's never `null`, so we can simplify it further.

ckelleyRH: True, that collapses this whole thing down to a one line ternary statement.

edewata (approved):
Just one comment, but it LGTM.

src/main/java/org/mozilla/jss/util/Password.java
```diff
@@ -100,6 +102,16 @@ public synchronized boolean equals(Object obj) {
...
     }
 
+    @Override
+    public int hashCode() {
+        final int prime = 31;
+        int result = 1;
+        result = prime * result + Arrays.hashCode(password);
+        result = prime * result + Arrays.hashCode(stackTrace);
+        result = prime * result + Objects.hash(cleared);
+        return result;
```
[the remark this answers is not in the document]
ckelleyRH: Yeah you're right, `equals()` only cares about `password` so I've modified `hashCode()` to do the same.

PR #739, opened 2021-05-25
Bump ubuntu_jdk8 -> 11 in github CI workflows

ckelleyRH:
I used some Java 11 syntax and this test failed (correctly, of course).

We should update this workflow to use OpenJDK11 so we can make use of new language features

### DIFF
--- a/.github/workflows/required.yml
+++ b/.github/workflows/required.yml
@@ -13,7 +13,7 @@ jobs:
           - 'fedora_latest_jdk11'
           - 'symbolcheck'
           - 'debian_jdk11'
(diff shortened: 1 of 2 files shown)


edewata (approved):
Should we rename the file to just `ubuntu` instead? We will only support Java 11 or later from now on, so I suppose can drop `jdk11` labels from the test names.

Regardless, feel free to update or merge as is. Thanks!

ckelleyRH:
I think we should keep the explicit version. Java 17 is coming in September and it will be interesting to test against it, by creating *17 jobs that are non-blocking in addition to the *11 jobs we have now.

PR #740, opened 2021-05-25
Add logger for build_pkcs11_constants.py

edewata:
The `build_pkcs11_constants.py` has been modified to send log messages to the screen instead of storing them in the output file. This way the output file will remain constant regardless how the tool was executed, and the log messages can be viewed without
having to open the output file which will make it easier to troubleshoot issues.


ckelleyRH (approved):
LGTM :+1:

PR #741, opened 2021-05-26
Reorganize examples

edewata:
The examples have been converted to become a sample Maven project that depends on JSS.
The Maven group ID for JSS has been changed such that it doesn't include the artifact ID itself.

https://github.com/edewata/jss/tree/maven/examples


ckelleyRH (approved):
Nice idea, LGTM!

edewata:
@ckelleyRH Thanks!

PR #742, opened 2021-05-26
Drop git dependency

edewata:
The `jss.spec` has been modified to drop `git` dependency so it's no longer required for building the official binaries, but `git` is still needed during development to call `build.sh --with-commit-id`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,7 +21,7 @@ COPY . /tmp/jss/
 WORKDIR /tmp/jss
 
 # Build JSS packages
-RUN dnf install -y rpm-build
+RUN dnf install -y git rpm-build
 RUN dnf builddep -y --spec jss.spec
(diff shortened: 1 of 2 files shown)


ckelleyRH (approved):
LGTM, but one question as an RPM n00b. What is the benefit of switching to `%setup` here? According to the [doc](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_autosetup) `%autosetup` with a VCS is bad (which is what was there before this commit fixed it), but your commit removes the explicit reference to a particular VCS, so is it not fine to keep `%autosetup`?

edewata:
@ckelleyRH I was under the impression that `%autosetup` requires a VCS. It's restored now without `git`. Thanks!

ckelleyRH:
Noice, I still approve :+1:

PR #743, opened 2021-05-26
Replace tabs with spaces

ckelleyRH:
This was done by executing the following in the repo top level directory:
`files=$(find -type f -name "*.java"); 'sed -i $'s/\t/    /g' $files'`

It won't fix all of the formatting problems, but it's a big enough diff already.


cipherboy (approved):
While you're doing messy changes, how about removing trailing whitespace from Java/C files? :)

emaldona:
I see problems: many files where using tabs set to equivalent to 8 spaces for example
src/test/java/org/mozilla/jss/tests/SetupDBs.java is a simple one
other files had line the start with several tabs followed by several spaces
src/test/java/org/mozilla/jss/tests/SetupDBs.java is another simple one.
I see indentation as two spaces rather than four as one would have expected.
I'm using meld to view the changes.

edewata:
I have a similar concern. The problem is the tabs might not have been used consistently even within a single file, so if we simply replace it with 4 spaces there's no guarantee that the final result will be nicely indented, and we might have to use auto format to fix it again. In that case I'd suggest to just use auto format in the first place, which should take care trailing spaces too, but maybe one or a few packages at a time so the PRs do not become too big.

edewata:
BTW, the CI failed due to git issue in F33 and F34 tests, but it's already fixed now so please rebase the PR.

ckelleyRH:
I am sympathetic to these concerns! The idea was to do the tidying up in stages to reduce the burden on the reviewer. The auto-format creates mega-diffs, even if done on a per-package basis, and I thought that removing the tabs as a separate change first would make life easier. At first glance with the auto-format it is not clear that tabs have been replaced with spaces, and this change makes it explicit that this was done. Also, for future archaeologists, this separate PR is evidence that we prefer spaces to tabs.

I don't mind doing a per-package auto-format instead, if you guys are happy to review that.

cipherboy:
Just a thought, but what about checking in the auto formatting tool you
mentioned earlier on IRC and running it as a ctest? We can each run the
script locally and validate that the before/after diff matches as expected
(i.e., we should be able to undo the format commit, run the format, save
e.g. checksums of the resulting files, then reapply the format commit from
the PR and the results should be identical).

If we're all running the same options, it should suffice for review, while
saving on number of PRs.

Thoughts?

cipherboy (changes requested):
Per Elio's comment, I'm in favor of using a proper autoformatter, though the PRs submitted seem to have a lot of... quirks. Maybe we have to use two tools (one for Java one for C?). But I think this is the hacky and probably wrong approach. My 2c.

ckelleyRH:
I'll try crowbar `uncrustify` in and see how it copes with both languages.

PR #772, opened 2021-05-27
Autoformat tools

ckelleyRH:


tools/build_pkcs11_constants.py
```diff
@@ -232,15 +232,15 @@ def check_output(self, nss_args):
         # see what the error was.
         ret = proc.returncode
         if proc.returncode != 0:
-            raise Exception(("Unknown error with symbol '%s': cc ret: %d " +
+            raise Exception(("Unknown error with symbol '%s': cc ret: %d " +
```
edewata, line 235: I don't think we want to add a trailing space..

PR #773, opened 2021-05-27
Autoformatting using uncrustify

ckelleyRH:
To run:
`yum install uncrustify`
`uncrustify --replace --no-backup -c - SSLVersionRange.h`
`Parsing: ./SSLVersionRange.h as language C-Header`

To validate:
`uncrustify --check -c - SSLVersionRange.h`
`Parsing: ./SSLVersionRange.h as language C-Header`
`PASS: ./SSLVersionRange.h (469 bytes)`

### DIFF
--- a/src/main/java/org/mozilla/jss/ssl/PrintOutputStreamWriter.java
+++ b/src/main/java/org/mozilla/jss/ssl/PrintOutputStreamWriter.java
@@ -1,8 +1,8 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-// 
-// 
+//
+//
 //
 // $Id$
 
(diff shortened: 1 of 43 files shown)


ckelleyRH:
I used the default config for both Java and C, but it's highly configurable - [this is the config ](https://github.com/uncrustify/uncrustify/blob/master/forUncrustifySources.cfg)`uncrustify` use to uncrustify `uncrustify`.

It compiled at least :1st_place_medal:

emaldona:
Regarding that default config for both Java and C is highly configurable I got some comments.  I have been looking at the different pull requests associated with this. In general I like what I seen regarding Java code but not so much what I seen on C code.  I don't know whether the JSS team has agreed on code style or not. Upstream Mozilla NSS has a sensible code guidelines at https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/NSS_Developer_Tutorial#nss_coding_style 
 my 2¢

src/main/java/org/mozilla/jss/ssl/PrintOutputStreamWriter.java
```diff
@@ -13,28 +13,28 @@
...
-    extends java.io.OutputStreamWriter
+class PrintOutputStreamWriter
+	extends java.io.OutputStreamWriter
 {
 
-    public PrintOutputStreamWriter(OutputStream out)
-    {
-    	super(out);
-    }
+public PrintOutputStreamWriter(OutputStream out)
```
cipherboy, line 20: I'd reasonably expect this to be indented, do others agree? I think this is a C++-ism more so than a Java-ism. Thoughts?

edewata: Agree. I think we probably should stick with Eclipse for autoformatting Java code, and leave C/Python code unchanged for now.

cipherboy: Did you see the issue with Eclipse in the Java code though? It completely murdered a nicely formatted enum declaration.

edewata: Ah yes, that's really bad.

PR #767, opened 2021-05-27
Autoformat org.mozilla.jss.ssl

ckelleyRH:

PR #774, opened 2021-06-04
Autoformat org.mozilla.jss.ssl using uncrustify

ckelleyRH:
To compare with the Eclipse autoformat in #767 @emaldona

### DIFF
--- a/src/main/java/org/mozilla/jss/ssl/CipherPolicy.java
+++ b/src/main/java/org/mozilla/jss/ssl/CipherPolicy.java
@@ -6,18 +6,21 @@ package org.mozilla.jss.ssl;
 
 public class CipherPolicy {
 
-    public static final CipherPolicy DOMESTIC =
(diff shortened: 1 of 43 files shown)


src/main/java/org/mozilla/jss/ssl/CipherPolicy.java
```diff
@@ -6,18 +6,21 @@
 
 public class CipherPolicy {
 
-    public static final CipherPolicy DOMESTIC =
-        new CipherPolicy(SocketBase.SSL_POLICY_DOMESTIC);
+public static final CipherPolicy DOMESTIC =
+	new CipherPolicy(SocketBase.SSL_POLICY_DOMESTIC);
```
edewata, line 10: Looks like all fields and methods in the class lose their indentations, and the code is indented with tabs instead of spaces.

ckelleyRH: There are some weird default choices with indentation, it is true.

PR #776, opened 2021-06-08
Convert RPM install script into CMake script

edewata:
The installation script in `jss.spec` has been converted into CMake script.
The `make` invocations have been modified to make it easier to read the output.

### DIFF
--- a/jss.spec
+++ b/jss.spec
@@ -125,13 +125,25 @@ modutil -dbdir /etc/pki/nssdb -chkfips true | grep -q enabled && export FIPS_ENA
 
 # The Makefile is not thread-safe
 %cmake \
+    -DVERSION=%{version} \
     -DJAVA_HOME=%{java_home} \
     -DJAVA_LIB_INSTALL_DIR=%{_jnidir} \
+    -DJSS_LIB_INSTALL_DIR=%{_libdir}/jss \
     -B %{_vpath_builddir}
 
 cd %{_vpath_builddir}
-%{__make} all
-%{__make} javadoc
+
+%{__make} \
+    VERBOSE=%{?_verbose} \
+    CMAKE_NO_VERBOSE=1 \
+    --no-print-directory \
+    all
+
+%{__make} \
+    VERBOSE=%{?_verbose} \
+    CMAKE_NO_VERBOSE=1 \
+    --no-print-directory \
+    javadoc
 
 %if %{with_test}
 ctest --output-on-failure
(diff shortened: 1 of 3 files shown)


jss.spec
```diff
@@ -124,10 +124,7 @@ export CFLAGS
 modutil -dbdir /etc/pki/nssdb -chkfips true | grep -q enabled && export FIPS_ENABLED=1
 
 # The Makefile is not thread-safe
-%cmake \
-    -DJAVA_HOME=%{java_home} \
```
cipherboy, line 128: At least `-DJAVA_HOME` is used by [`cmake`'s `FindJAVA`](https://cmake.org/cmake/help/latest/module/FindJava.html) library.

cipherboy (changes requested):
I'd look a little more closely at these variables. I think some of them influence CMake's decisions about which toolchains to use, which would help building on a multi-toolchain environment (e.g., jdk8 and jdk11 both present -- which is allowed).

jss.spec
cipherboy: I think `FindJNI` might use the other, but I'm not entirely sure.

edewata: @cipherboy Thanks, I updated the PR to do something completely different now :)

The ultimate goal is to modify the `build.sh` to call CMake directly instead of `rpmbuild`, then modify the `jss.spec` to call `build.sh` instead of CMake. That way the upstream JSS can be built with `build.sh` without depending on Fedora-specific packaging.

cipherboy: Ah nifty :) That's a good idea!

cipherboy (commented):
(Clearing previous requested changes)

cipherboy (approved):
(I guess I have to approve it even though I haven't reviewed it in order to remove the request changes. This is not a review of the patch, just a way to show the previously requested feedback has been addressed.)

edewata:
@ckelleyRH Could you take a look at this? Thanks!

ckelleyRH (commented):
I know next-to-nothing about `cmake` so this is a surface-level review. I was able to follow the build instructions and build without error using both `cmake` and `build.sh`. A comprehensive review this is not, so I won't approve, but I've no objections to the change and the rationale is a good one.

edewata:
@cipherboy @ckelleyRH Thanks for looking! I'm not an expert on CMake either, so this is mostly based on trial and error. I was able to remove some references to `DESTDIR` in the last update. The CI and PKI seem to be working just fine so I'll merge this.

PR #778, opened 2021-06-11
Update version number to 5.0.0-alpha1

edewata:
Recently JSS 4.9 was modified to require Java 11. However, older platforms that use JSS 4.8 might have other libraries that will not work with Java 11, so it will not be able to be upgraded to JSS 4.9.

To address the problem, the current JSS 4.9 has been renamed into JSS 5.0 which will only work on newer platforms, and a new JSS 4.9 branch will be created from JSS 4.8 to provide updates for older platforms.

The binaries have also been renamed to `jss.jar` and `libjss.so` to simplify future upgrades.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,14 +81,14 @@ jss_tests()
 
 install(
     FILES
-        ${CMAKE_CURRENT_BINARY_DIR}/jss4.jar
+        ${CMAKE_CURRENT_BINARY_DIR}/jss.jar
     DESTINATION
         ${JAVA_LIB_INSTALL_DIR}
 )
 
 install(
     FILES
-        ${CMAKE_CURRENT_BINARY_DIR}/libjss4.so
+        ${CMAKE_CURRENT_BINARY_DIR}/libjss.so
     DESTINATION
         ${JSS_LIB_INSTALL_DIR}
     PERMISSIONS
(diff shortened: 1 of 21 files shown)


cipherboy (commented):
No horse in this race, but within the 4.4.x series we moved from a JDK7 min-dep to a JDK8 min-dep partly by accident (though, Dogtag 10.5.x made the same change a little earlier and more intentionally). Considering that we're on the late edge of JDK11 adoption relative to the broader ecosystem, I'm thinking it might be fine to leave it as 4.9 unless you're planning on leaving 4.9+ for RHEL 8 and leave it strictly JDK8 and move the JDK11 features strictly RHEL 9.

ckelleyRH:
I think having a clean break `java 8 -> JSS4.x -> RHEL8; java 11 -> JSS5.x -> RHEL9` is nicely rationalised.

ckelleyRH (approved):
LGTM.

edewata:
Yeah, we need to do the same thing with PKI (i.e. renaming PKI 10.11 to PKI 11). Unfortunately Java 11 requires newer packages (e.g. JAXB, Activation) that are not available in RHEL 8, so we need to continue providing PKI 10.x and JSS 4.x to RHEL 8.x using the same dependencies already available in RHEL 8.x (including Java 1.8). Newer platforms can use PKI 11 with JSS 5 using Java 11.

@cipherboy @ckelleyRH Thanks!

Issue #781, opened 2021-07-02
NPE in Buffer.WriteCapacity()


edewata:
Sometimes the `Buffer.WriteCapacity()` throws an NPE:

```
2021-07-02T18:29:32.7931657Z + pki -n caadmin ca-user-cert-add caagent --serial 0x7
2021-07-02T18:29:36.2129529Z java.lang.NullPointerException
2021-07-02T18:29:36.2131194Z 	at org.mozilla.jss.nss.Buffer.WriteCapacity(Native Method)
2021-07-02T18:29:36.2133683Z 	at org.mozilla.jss.ssl.javax.JSSEngineReferenceImpl.unwrap(JSSEngineReferenceImpl.java:1192)
2021-07-02T18:29:36.2136070Z 	at org.mozilla.jss.ssl.javax.JSSSocketChannel.read(JSSSocketChannel.java:276)
2021-07-02T18:29:36.2138079Z 	at java.base/java.nio.channels.SocketChannel.read(SocketChannel.java:486)
2021-07-02T18:29:36.2140097Z 	at org.mozilla.jss.ssl.javax.JSSSocketChannel.read(JSSSocketChannel.java:236)
2021-07-02T18:29:36.2142914Z 	at org.mozilla.jss.ssl.javax.JSSSocketChannel.implCloseSelectableChannel(JSSSocketChannel.java:381)
2021-07-02T18:29:36.2146867Z 	at java.base/java.nio.channels.spi.AbstractSelectableChannel.implCloseChannel(AbstractSelectableChannel.java:242)
2021-07-02T18:29:36.2150320Z 	at java.base/java.nio.channels.spi.AbstractInterruptibleChannel.close(AbstractInterruptibleChannel.java:112)
2021-07-02T18:29:36.2155237Z 	at org.mozilla.jss.ssl.javax.JSSSocket.close(JSSSocket.java:710)
2021-07-02T18:29:36.2156889Z 	at org.apache.http.impl.BHttpConnectionBase.close(BHttpConnectionBase.java:332)
2021-07-02T18:29:36.2159757Z 	at org.apache.http.impl.conn.LoggingManagedHttpClientConnection.close(LoggingManagedHttpClientConnection.java:81)
2021-07-02T18:29:36.2162361Z 	at org.apache.http.impl.conn.CPoolEntry.closeConnection(CPoolEntry.java:70)
[...]

PR #786, opened 2021-07-08
Update PKI CA test

edewata:
The CA test has been update to match the CA test in PKI:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-tests.yml#L60-L160

This can be used to reproduce issue #781 and test the fix when it becomes available.


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH

PR #787, opened 2021-07-09
Downgrade NSS dependency

edewata:
The NSS dependency has been temporarily downgraded to
avoid intermittent CI failures reported in Issue #781.

Additional dependencies to NSS subpackages have been
added to avoid conflicts. The `PKCS11Constants` tests
have been temporarily disabled as well.


edewata:
It doesn't seem to fix the problem so I'll close this PR.

PR #789, opened 2021-07-28
Reorganize PKCS #11 constants tests

edewata:
The PKCS #11 constants tests for Fedora 33 and 34 have
been converted into a new workflow such that the platform
versions will update automatically and can be configured
using the test MATRIX.

https://github.com/dogtagpki/pki/wiki/Configuring-Test-Matrix


ckelleyRH (approved):
LGTM! It would be good to finish making the Github action work so we don't have to keep copy/pasting the set up matrix code.

c-dorney (approved):

edewata:
@ckelleyRH @c-dorney Thanks!

Let me see if I can figure out something about the duplicate code.

ckelleyRH:
@edewata the skeleton of a solution is here: https://github.com/dogtagpki/init-ci/blob/master/action.yml. I got it to work, but the test matrix override did not work for some reason.

PR #790, opened 2021-07-28
Add init-workflow.sh

edewata:
The `init-workflow.sh` has been added to configure the test
matrix based on the `BASE64_MATRIX` variable. The test matrix
needs to be base64-encoded since otherwise GitHub will mask
the value rendering it unusable.

https://github.com/dogtagpki/jss/wiki/Configuring-Test-Matrix

### DIFF
--- a/.github/workflows/pkcs11-tests.yml
+++ b/.github/workflows/pkcs11-tests.yml
@@ -10,18 +10,15 @@ jobs:
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
       - name: Set up test matrix
         id: set-matrix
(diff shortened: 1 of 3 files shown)


c-dorney (approved):
LGTM

ckelleyRH (approved):
Ah, maybe the lack of encoding hurt me with the action too! The benefit of the action over this method is we can re-use the code across multiple repos but I still like refactoring it out into a script so I approve.

edewata:
@c-dorney @ckelleyRH Thanks! I removed the unnecessary `export` statements. We can try the action again if we want. At least this is a progress.

ckelleyRH:
This is significantly better, and some might argue defining the script in each repo is better than establishing a whole new repo to harbour a github action, spreading the CI code over N repos.

PR #793, opened 2021-07-30
Convert symbol test

edewata:
The symbol test has been converted from a `Dockerfile` into a GH workflow.


ckelleyRH (approved):
LGTM!

The CA test failed because of known issues, perhaps we should temporarily block that test here too like we did in the pki repo?

c-dorney (approved):
Awesome the workflow is so much easier to read

edewata:
@ckelleyRH @c-dorney Thanks!

The CA test is actually coming from `pki-tests` package, so it should have been fixed by https://github.com/dogtagpki/pki/pull/3673, but apparently it's not sufficient. I'll look into that.

PR #794, opened 2021-07-30
Convert build tests

edewata:
The build tests for Fedora, Debian, and Ubuntu have been
converted from `Dockerfile`s into a GH workflow.


.github/workflows/required.yml
```diff
@@ -3,24 +3,40 @@ name: Required Tests
...
+      if: ${{ startsWith(matrix.os, 'fedora:') || startsWith(matrix.os, 'centos:') }}
+      run: |
+        dnf install -y dnf-plugins-core rpm-build
+        dnf builddep -y --spec jss.spec
+
+    - name: Install Debian/Ubuntu dependencies
+      if: ${{ startsWith(matrix.os, 'debian:') || startsWith(matrix.os, 'ubuntu:') }}
+      run: |
+        apt-get update
+        apt-get install -y \
```
ckelleyRH, line 33: This is fewer dependencies than specified in either of the old Docker images. I'm happy with it, we obviously don't need them and removing dead code is good, I'm just curious how you knew to remove them.

ckelleyRH (approved):
This is much improved, thanks!

.github/workflows/required.yml
edewata: @ckelleyRH Thanks! I didn't actually expect to remove anything, but when there's a static list like this sometimes we just need to revalidate it. So I started from scratch and readd the dependencies if they are still needed.

PR #797, opened 2021-08-05
Drop deprecated methods in CertAndKeyGen

edewata:

### DIFF
--- a/docs/changes/v5.0.0/API-Changes.adoc
+++ b/docs/changes/v5.0.0/API-Changes.adoc
@@ -1,0 +1,6 @@
+= API Changes =
+
+== Changes in org.mozilla.jss.netscape.security.x509.CertAndKeyGen ==
+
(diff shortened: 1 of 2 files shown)


ckelleyRH (approved):
LGTM! Both `setRandom()` and `getSelfCert()` appear unused in JSS and no references in PKI either.

One small note, that's not part of your diff so it's fine if you don't change it: https://github.com/dogtagpki/jss/pull/797/files#diff-0363f382330565caa7daee266f3dcc675c8b86e523593a1f5ffeebbbdeef530dL241-L242
^^this exception at the end of `getSelfCertificate()` contains the string `"getSelfCert: "`, which is the name of the replaced method. I found this confusing. I think this should be changed to `"getSelfCertificate: "` but I approve regardless.

edewata:
@ckelleyRH I fixed that. Thanks!

PR #798, opened 2021-08-10
Drop deprecated X509Cert class

edewata:


ckelleyRH (approved):
LGTM.

ckelleyRH:
> Looks good
>
>Out of curiosity, why is X509Cert.java being dropped? Is there something else to use in place of it?

`X509Cert` implements the deprecated `java.security.Certificate` interface from the JRE (deprecated since 1.2 :scream:) - the replacement is `java.security.cert.X509Certificate`, which is part of the package created to replace the former.

edewata:
Good question @c-dorney, and thanks @ckelleyRH for the clarification! I've updated the doc to mention the replacement. Thanks!

Issue #782, opened 2021-07-02
JSSEngine.queryEnabledCipherSuites() generates unknown cipher warnings


edewata:
The `JSSEngine.queryEnabledCipherSuites()` generates the following warnings:

```
2021-07-02T18:29:32.7931657Z + pki -n caadmin ca-user-cert-add caagent --serial 0x7
2021-07-02T18:29:34.6674821Z WARNING: Unable to get the value of cipher: SSL2_RC4_128_WITH_MD5 (65281): Unknown cipher suite to get or getting its value failed
2021-07-02T18:29:34.6678907Z WARNING: Unable to get the value of cipher: SSL2_RC4_128_EXPORT40_WITH_MD5 (65282): Unknown cipher suite to get or getting its value failed
2021-07-02T18:29:34.6687405Z WARNING: Unable to get the value of cipher: SSL2_RC2_128_CBC_WITH_MD5 (65283): Unknown cipher suite to get or getting its value failed
2021-07-02T18:29:34.6694609Z WARNING: Unable to get the value of cipher: SSL2_RC2_128_CBC_EXPORT40_WITH_MD5 (65284): Unknown cipher suite to get or getting its value failed
2021-07-02T18:29:34.6701624Z WARNING: Unable to get the value of cipher: SSL2_IDEA_128_CBC_WITH_MD5 (65285): Unknown cipher suite to get or getting its value failed
2021-07-02T18:29:34.6710035Z WARNING: Unable to get the value of cipher: SSL2_DES_64_CBC_WITH_MD5 (65286): Unknown cipher suite to get or getting its value failed
2021-07-02T18:29:34.6716870Z WARNING: Unable to get the value of cipher: SSL2_DES_192_EDE3_CBC_WITH_MD5 (65287): Unknown cipher suite to get or getting its value failed
2021-07-02T18:29:34.6723499Z WARNING: Unable to get the value of cipher: TLS_NULL_WITH_NULL_NULL (0): Unknown cipher suite to get or getting its value failed
2021-07-02T18:29:34.6730275Z WARNING: Unable to get the value of cipher: SSL3_RSA_EXPORT_WITH_RC4_40_MD5 (3): Unknown cipher suite to get or getting its value failed
2021-07-02T18:29:34.6736875Z WARNING: Unable to get the value of cipher: TLS_RSA_EXPORT_WITH_RC4_40_MD5 (3): Unknown cipher suite to get or getting its value failed
[...]

PR #799, opened 2021-08-11
Reduce warnings in JSSEngine.queryEnabledCipherSuites()

edewata:
The `JSSEngine.queryEnabledCipherSuites()` has been modified to check whether a cipher is supported before getting its default preference.
    
The change eliminates most of the warnings shown in issue #782 except for the following ciphers:
* `TLS_EMPTY_RENEGOTIATION_INFO_SCSV`
* `TLS_FALLBACK_SCSV`


ckelleyRH (commented):
It looked to me that the remaining two warnings could be fixed by changing the constructor for the fallback/empty ciphers:

```
    private SSLCipher(int id) {
        /* Should only be used with TLS_EMPTY_RENEGOTIATION_INFO_SCSV and
         * TLS_FALLBACK_SCSV. */
-      this(id, false, null, true);
+      this(id, false, null, false);
    }
```
This led me to explore the consequences of doing so:

The enum value for `TLS_EMPTY_RENEGOTIATION_INFO_SCSV` is only referenced in one test in `JSSE_SSLClient`, so that could go. The enum value for `TLS_FALLBACK_SCSV` is unused. There are static constants in `SSLSocket` with the same names, which are both unused. Are these things magically used by some non-Java process I have not seen yet, or can we safely remove the constants, the enum entries, and the constructor?

src/main/java/org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -531,8 +544,9 @@ public void setEnabledCipherSuites(SSLCipher[] suites) throws IllegalArgumentExc
         ArrayList<SSLCipher> enabledCiphers = new ArrayList<SSLCipher>();
 
         for (SSLCipher cipher : SSLCipher.values()) {
+            if (!cipher.isSupported()) continue;
             try {
-                if (SSL.CipherPrefGetDefault(cipher.getID()) && cipher.isSupported()) {
+                if (SSL.CipherPrefGetDefault(cipher.getID())) {
```
c-dorney, line 549:
```suggestion
                if (cipher.isSupported() && SSL.CipherPrefGetDefault(cipher.getID())) {
```
This might be a nicer read? The latter won't be checked if `cipher.isSupported()` is false

edewata: Sounds good. I updated the PR.

c-dorney (approved):

edewata:
Looks like the two remaining ciphers are still defined in NSS:
* https://github.com/nss-dev/nss/blob/master/lib/ssl/sslproto.h#L150
* https://github.com/nss-dev/nss/blob/master/lib/ssl/sslproto.h#L155

so technically they are still "supported" by NSS. I'm not sure whether we should change `SSLCipher.supported` for these ciphers to `false`, or whether we can just drop those ciphers.

@cipherboy What do you think?

edewata:
@ckelleyRH I just changed the `SSLCipher` constructor to get the `supported` value from NSS, and NSS does actually return `false` for those ciphers, so all the warnings should be gone now.

ckelleyRH (approved):
Awesome, LGTM!

edewata:
@c-dorney @ckelleyRH Thanks!
@cipherboy If you have any comments on this just let me know. Thanks!

PR #800, opened 2021-08-11
Add build tests for Azure pipeline

edewata:
This is an initial attempt to create an Azure pipeline based on the existing [build tests](https://github.com/dogtagpki/jss/blob/master/.github/workflows/build-tests.yml).

The results are available here:
https://dev.azure.com/edewata/jss/_build/results?buildId=31&view=results
It's supposed to be public, but please let me know if you can't see it.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,72 @@
+# Workaround to install sudo
+# https://github.com/Microsoft/azure-pipelines-agent/issues/2043#issuecomment-687983301
+resources:
+  containers:
+  - container: debian_testing
+    image: debian:testing
+    options: '--name ci-container -v /usr/bin/docker:/tmp/docker:ro'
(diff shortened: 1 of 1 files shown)


ckelleyRH:
Looks good! The `freeipa` guys have their Azure CI report back here in the checks, can we make it do that too?

edewata:
@ckelleyRH Yes, but apparently it has to be set up after adding the `azure-pipelines.yml`.

ckelleyRH (approved):
OK, then let's merge this in so we have something to play with,

edewata:
@ckelleyRH Thanks!

edewata:
Here's the pipeline:
https://dev.azure.com/dogtagpki/jss/_build?definitionId=1&_a=summary
I think the CI report will appear not in this PR but in the next one.

PR #801, opened 2021-08-11
Remove redundant superinterface implementation statements

ckelleyRH:
Does what it says on the tin.

It also happens to be a test for Azure CI pipeline reporting back to Github.


edewata (approved):
LGTM!

Too bad we can't see Azure logs directly in GitHub. We probably can try this next:
https://github.com/marketplace/actions/azure-pipelines-action

PR #803, opened 2021-08-17
Fix: Bug 1964176 - KRA PKCS12 support for nCipher sw v12.60+.

jmagne:
This bug has it's jss portion reflected here. The gist of this fix
is to register 3 new secoids into nss in a dynamic fashion. The 3
new algs are 128, 192, and 256 bit variants of the machanism,
CKM_AES_KEY_WRAP_KWP.

We have to provide support for these algs because we need a common algorithm
that works the same in the nss software db as well as in the hsm.

The key jss method that is failing is one that is reponsible for creating the
so called EncryptedPrivateKeyInfo blob, which is packaged in the final p12 file.

In order to accomplish we have had to create our own version of the nss routine,
PK11_ExportEncryptedPrivateKeyInfo, which is able to use one of the 3 new encryptiion
algs to wrap the private key in the final blob. Jss is also implemting a new version of the
routine in nss which is reponsible for performing a key exchange of a session key between
the softeare db and the hsm. This code will use the oaep wrapping alg on the hsm to do the
heavy lifting, while creating a temp private key on the hsm to unwrap the session key from
the nss software db to the hsm. As a final step the p12 code will encode one of the 3 new oids
to indicate that one of the 3 new algorithms have been used.

Here are the declarations of the 3 new algs:

 public static final EncryptionAlgorithm
   AES_128_KEY_WRAP_KWP = new EncryptionAlgorithm(SEC_OID_AES_128_KEY_WRAP_KWP,
        Alg.AES,Mode.NONE,
        Padding.PKCS5, IVParameterSpecClasses, 16,
        AES_ROOT_OID.subBranch(8), 128,"AES/None/PKCS5Padding/Kwp/128");

  public static final EncryptionAlgorithm
   AES_192_KEY_WRAP_KWP = new EncryptionAlgorithm(SEC_OID_AES_192_KEY_WRAP_KWP,
        Alg.AES,Mode.NONE,
        Padding.PKCS5, IVParameterSpecClasses, 16,
        AES_ROOT_OID.subBranch(28), 192,"AES/None/PKCSPadding/Kwp/192");

  public static final EncryptionAlgorithm
   AES_256_KEY_WRAP_KWP = new EncryptionAlgorithm(SEC_OID_AES_256_KEY_WRAP_KWP,
        Alg.AES,Mode.NONE,
        Padding.PKCS5, IVParameterSpecClasses, 16,
        AES_ROOT_OID.subBranch(48), 256,"AES/None/PKCS5Padding/Kwp/256");

Note that we have to provide kind of a custom name for each of these algs, to differentiate
between the 3 variants. When actually referring to these algs from jss, we can use that name
to look up the proper alg. Also note the 3 new SEC_OID values. Since these are not yet in nss,
we have dynamically added these 3 sec oids into jss for use.

The rest of this code is devoted to the support for adding dynmic oids into nss on behalf of jss.


src/main/java/org/mozilla/jss/pkcs11/PK11Store.c
```diff
@@ -688,9 +688,23 @@ Java_org_mozilla_jss_pkcs11_PK11Store_getEncryptedPrivateKeyInfo(
...
         slot, algTag, pwItem, privk, iterations, NULL /*wincx*/);
     if (epki == NULL) {
-        JSS_throwMsgPrErr(
-            env, TOKEN_EXCEPTION, "Failed to export EncryptedPrivateKeyInfo");
-        goto finish;
+	    //try our own version for the new AES KWP algs.
+
+	epki = JSS_ExportEncryptedPrivKeyInfoV2(
+                   slot,
+        algTag,     /* PBE algorithm to encrypt the with key */
```
ladycfu, line 695: typo? should be /" PBE algorithm to encrypt the key with */

src/main/java/org/mozilla/jss/util/jssutil.c
```diff
@@ -926,6 +926,271 @@ JSS_clearPtrFromProxy(JNIEnv *env, jobject nativeProxy)
     }
 }
 
+/* V2 refers to PKCS #5 V2 here. If a PKCS #5 v1 or PKCS #12 pbe is passed
```
ladycfu, line 929: Would be a good idea to identify which portion of the code has been copied from nss to 1. save time for the reviewers, and 2. for future

PR #804, opened 2021-08-18
This bug has it's jss portion reflected here. The gist of this fix

jmagne:
is to register 3 new secoids into nss in a dynamic fashion. The 3
new algs are 128, 192, and 256 bit variants of the machanism,
CKM_AES_KEY_WRAP_KWP.

We have to provide support for these algs because we need a common algorithm
that works the same in the nss software db as well as in the hsm.

The key jss method that is failing is one that is reponsible for creating the
so called EncryptedPrivateKeyInfo blob, which is packaged in the final p12 file.

In order to accomplish we have had to create our own version of the nss routine,
PK11_ExportEncryptedPrivateKeyInfo, which is able to use one of the 3 new encryptiion
algs to wrap the private key in the final blob. Jss is also implemting a new version of the
routine in nss which is reponsible for performing a key exchange of a session key between
the softeare db and the hsm. This code will use the oaep wrapping alg on the hsm to do the
heavy lifting, while creating a temp private key on the hsm to unwrap the session key from
the nss software db to the hsm. As a final step the p12 code will encode one of the 3 new oids
to indicate that one of the 3 new algorithms have been used.

Here are the declarations of the 3 new algs:

public static final EncryptionAlgorithm
AES_128_KEY_WRAP_KWP = new EncryptionAlgorithm(SEC_OID_AES_128_KEY_WRAP_KWP,
Alg.AES,Mode.NONE,
Padding.PKCS5, IVParameterSpecClasses, 16,
AES_ROOT_OID.subBranch(8), 128,"AES/None/PKCS5Padding/Kwp/128");

public static final EncryptionAlgorithm
AES_192_KEY_WRAP_KWP = new EncryptionAlgorithm(SEC_OID_AES_192_KEY_WRAP_KWP,
Alg.AES,Mode.NONE,
Padding.PKCS5, IVParameterSpecClasses, 16,
AES_ROOT_OID.subBranch(28), 192,"AES/None/PKCSPadding/Kwp/192");

public static final EncryptionAlgorithm
AES_256_KEY_WRAP_KWP = new EncryptionAlgorithm(SEC_OID_AES_256_KEY_WRAP_KWP,
Alg.AES,Mode.NONE,
Padding.PKCS5, IVParameterSpecClasses, 16,
AES_ROOT_OID.subBranch(48), 256,"AES/None/PKCS5Padding/Kwp/256");

Note that we have to provide kind of a custom name for each of these algs, to differentiate
between the 3 variants. When actually referring to these algs from jss, we can use that name
to look up the proper alg. Also note the 3 new SEC_OID values. Since these are not yet in nss,
we have dynamically added these 3 sec oids into jss for use.

The rest of this code is devoted to the support for adding dynmic oids into nss on behalf of jss.


src/main/java/org/mozilla/jss/pkcs11/PK11Store.c
```diff
@@ -688,9 +688,23 @@ Java_org_mozilla_jss_pkcs11_PK11Store_getEncryptedPrivateKeyInfo(
     epki = PK11_ExportEncryptedPrivKeyInfo(
         slot, algTag, pwItem, privk, iterations, NULL /*wincx*/);
     if (epki == NULL) {
-        JSS_throwMsgPrErr(
-            env, TOKEN_EXCEPTION, "Failed to export EncryptedPrivateKeyInfo");
-        goto finish;
+	    //try our own version for the new AES KWP algs.
+
+	epki = JSS_ExportEncryptedPrivKeyInfoV2(
```
cipherboy, line 693: Indentation is a little weird starting here.

src/main/java/org/mozilla/jss/util/jssutil.c
```diff
@@ -926,6 +926,271 @@ JSS_clearPtrFromProxy(JNIEnv *env, jobject nativeProxy)
...
+    if (!epki->encryptedData.len) {
+        rv = SECFailure;
+        goto loser;
+    }
+
+    crv = PK11_WrapPrivKey(pk->pkcs11Slot,  key,
+                            pk,  cryptoMechType,
+                           iv,  &epki->encryptedData, NULL);
+
+  /*  crv = PK11_GETTAB(pk->pkcs11Slot)->C_WrapKey(pk->pkcs11Slot->session, &cryptoMech, key->objectID, pk->pkcs11ID, NULL, &encBufLen); */
```
cipherboy, line 1063: Here as well about indentation.

src/main/java/org/mozilla/jss/util/jssutil.c
```diff
@@ -926,6 +926,271 @@ JSS_clearPtrFromProxy(JNIEnv *env, jobject nativeProxy)
...
+             * back to RSA PKCS#1v1.5. */
+            our_mech = CKM_RSA_PKCS;
+            mech_param = NULL;
+        }
+
+        wrapData.data = NULL;
+
+        /* Just go ahead and generate a temp private key,
+         * since this method is not going to be called often,
+	 * and we don't have to worry about finding a key with the
```
cipherboy, line 1141: Indentation ;-)

cipherboy (commented):
Just some random indentation things I mentioned on IRC earlier today ;-) Code's looking good, Jack!

src/main/java/org/mozilla/jss/CryptoManager.c
```diff
@@ -22,7 +22,7 @@
 #include <jssver.h>
 
 #include "pk11util.h"
-
+#include <Algorithm.h>
```
cipherboy, line 25: Generally I like the idea of using angle brackets for system libraries and double quotes for internal (JSS) libraries but this hasn't always been followed. I'll leave it up to you though.

PR #805, opened 2021-08-20
Deprecate RevocationReasonAdapter in JSS 4

edewata:
https://github.com/edewata/jss/blob/v4.9.x-deprecation/docs/changes/v4.9.1/API-Changes.adoc

### DIFF
--- a/src/main/java/org/mozilla/jss/netscape/security/x509/RevocationReasonAdapter.java
+++ b/src/main/java/org/mozilla/jss/netscape/security/x509/RevocationReasonAdapter.java
@@ -24,8 +24,10 @@ import org.apache.commons.lang3.StringUtils;
 /**
  * The RevocationReasonAdapter class provides custom marshaling for RevocationReason.
  *
+ * @deprecated Use the revocation reason code or label instead.
  * @author Endi S. Dewata
  */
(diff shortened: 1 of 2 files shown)


src/main/java/org/mozilla/jss/netscape/security/x509/RevocationReasonAdapter.java
```diff
@@ -24,8 +24,10 @@
 /**
  * The RevocationReasonAdapter class provides custom marshaling for RevocationReason.
  *
+ * @deprecated Use the revocation reason code or label instead.
  * @author Endi S. Dewata
  */
+@Deprecated
```
ckelleyRH, line 30: Seeing as we know we are removing this we should provide this information to future archaeologists:
`@Deprecated(since="4.9.1", forRemoval=true)`

edewata: Nice! I wasn't aware of this feature. Patch updated.

ckelleyRH (approved):
LGTM!

c-dorney (approved):

edewata:
@ckelleyRH @c-dorney Thanks!

PR #806, opened 2021-08-20
Drop RevocationReasonAdapter and JAXB dependency from JSS 5

edewata:


ckelleyRH (approved):
All references to JAXB have been removed, the CI still passes and I can get rpms with `build.sh` and a JAR using Maven so I think we're good.

edewata:
Thanks!

PR #808, opened 2021-09-23
Fix Bug 2001169 - Audit event 'ACCESS_SESSION_ESTABLISH' is not gener…

jmagne:
…ating for PKI instances acting as Server [10.2.1]

    This fix allows us to actually see ssl connection events in the audit log from the pki /server perspective.
    This fill will also require support bug fixes for both pki and tomcatjss.

Based on original code contributed by Alex Ascheel, (cipherboy).

### DIFF
--- a/src/main/java/org/mozilla/jss/ssl/javax/JSSEngine.java
+++ b/src/main/java/org/mozilla/jss/ssl/javax/JSSEngine.java
@@ -1,6 +1,8 @@
 package org.mozilla.jss.ssl.javax;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.EventListener;
 import java.util.HashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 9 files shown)


src/main/java/org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -935,6 +953,71 @@ public boolean getWantClientAuth() {
...
+        return listeners;
+    }
+
+    /**
+     * Fires any and all SSLSocketListeners on the specified alert received event.
+     *
+     * To be used by other implementations of JSSEngine.
+     */
+    protected void fireAlertReceived(SSLAlertEvent event) {
+	    System.err.println("JsSEngine.fireAlertReceived: " + event + " listeners: " + listeners);
```
ladycfu, line 976: Probably should consider removing println's as this could become excessive.  Same with all the non-error debugs.

ladycfu (approved):
Under the assumption that @jmagne is @cipherboy 's reviewer of this JSS patch, and @jmagne has tested the patch, I merely took a quick look and it seems fine other than the potential excessive messages from println that I think should probably be commented out or removed.

PR #814, opened 2021-11-01
Remove SHA-1 algorthim options from enums.

ckelleyRH:
* Replace PBE_SHA1_DES3_CBC with PBE_PKCS5_PBMAC1 as a default.

### DIFF
--- a/src/main/java/org/mozilla/jss/CryptoManager.java
+++ b/src/main/java/org/mozilla/jss/CryptoManager.java
@@ -571,13 +571,13 @@ public final class CryptoManager implements TokenSupplier
         // Force class load before we install the provider. Otherwise we get
         // an infinite loop as the Security manager tries to instantiate the
         // digest to verify its own JAR file.
-        JSSMessageDigestSpi mds = new JSSMessageDigestSpi.SHA1();
+        JSSMessageDigestSpi mds = new JSSMessageDigestSpi.SHA256();
         logger.debug("Loaded " + mds);
 
         // Force the KeyType class to load before we can install JSS as a
         // provider.  JSS's signature provider accesses KeyType.
         KeyType kt = KeyType.getKeyTypeFromAlgorithm(
-            SignatureAlgorithm.RSASignatureWithSHA1Digest);
+            SignatureAlgorithm.RSASignatureWithSHA256Digest);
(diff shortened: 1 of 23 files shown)


ckelleyRH:
I'm making a PR to `dogtagpki/pki` that will address the PKI build issues.

src/main/java/org/mozilla/jss/CryptoManager.java
```diff
@@ -571,13 +571,13 @@ public static synchronized void initialize( InitializationValues values )
         // Force class load before we install the provider. Otherwise we get
         // an infinite loop as the Security manager tries to instantiate the
         // digest to verify its own JAR file.
-        JSSMessageDigestSpi mds = new JSSMessageDigestSpi.SHA1();
```
cipherboy, line 574: I'm not entirely sure how this works. Maybe it is best to remove it altogether, though some coordination with JDK team and other upstream members will need to be done.

In particular, this code exists from an era where all JCA Provider JARs needed to be signed. I'm not sure if that has changed, and/or what signature algorithms it supports. If anyone is presently signing JCAs, using SHA-1, then we'd have a bit of a problem and/or need to encourage them to move to SHA-256. If this functionality is removed/deprecated in upstream JDKs, we can probably remove it altogether.

Just letting you know about the historical purpose of this, so you could decide the best path :-) Mickey (ask Matt) might have more information on this, outside of other sources available from Google.

ckelleyRH: I'm going to drop this PR in favour of another I'm cooking up, having got more insight from the crypto guys here. But seeing as we're already discussing it here I'll continue!

It looks like these two `SHA-1` referencing objects only exist to do class-loading magic, so I guess that switching them to `SHA-256` is safe. It would be great to not have to do this and remove it but that could have consequences and I'm trying to leave the smallest possible footprint.

cipherboy: Right, my understanding is this class loading magic needs to happen if/when the JAR library is signed with a SHA-1-type signature... If a SHA-256 signature is used, SHA-256 would of course need to be loaded, but I think my point was that we need to make sure signatures _have_ transitioned to SHA-256 before we make this change... My 2c :-)

PR #816, opened 2021-11-02
Update Fedora container and COPR repository

edewata:
JSS 4.9 is only supported up to Fedora 34 for PKI 10.12, so the Fedora container has been changed to `fedora:34` and the  COPR repository has been changed to `@pki/10.12`.


c-dorney (approved):
LGTM

edewata:
@c-dorney Thanks!

PR #817, opened 2021-11-03
Add JSS symkey library

edewata:
To consolidate Java API for NSS the PKI symkey library has been imported into JSS with the following changes:

- the package name has been renamed to `org.mozilla.jss.symkey`
- the JAR file has been renamed to `jss-symkey.jar`
- the shared library file has been renamed to `libjss-symkey.so`

The code loading the library in `SessionKey` has been simplified as well.
In the future the JAR file and the shared library file might be merged into `jss.jar` and `libjss.so`.

Doc: https://github.com/edewata/jss/blob/symkey/docs/changes/v5.1.0/API-Changes.adoc

See also: https://github.com/dogtagpki/pki/issues/1368

**Note:** Once this is merged, PKI will be [modified](https://github.com/edewata/pki/commit/bf02277425c9a4b654d89c6cd69a903a3d7e6fab) to use JSS symkey instead of PKI symkey, then PKI symkey can be [dropped](https://github.com/edewata/pki/commit/7bff44ef173446ed91f5f8790407202e47ec41da). This will reduce the number of native packages in PKI.


edewata:
ACKed by @jmagne on IRC. Thanks!

PR #818, opened 2021-11-04
Remove redundant type specifications

ckelleyRH:


cmake/JSSConfig.cmake
```diff
@@ -295,11 +295,11 @@ macro(jss_config_java)
     list(APPEND JSS_TEST_JAVAC_FLAGS "-sourcepath")
     list(APPEND JSS_TEST_JAVAC_FLAGS "${PROJECT_SOURCE_DIR}/src/main/java")
 
-    # Ensure we're compatible with JDK 8
+    # Ensure we're compatible with JDK 1
```
edewata, line 298: 11? :)

edewata (approved):
I think there's one typo in the comment, but otherwise it LGTM.

cmake/JSSConfig.cmake
ckelleyRH: Opps, thanks!

c-dorney (approved):

PR #822, opened 2021-11-05
"Fix Bug 2001169 - Audit event 'ACCESS_SESSION_ESTABLISH' is not gener…

c-dorney:
…ating for PKI instances acting as Server [10.2.1]

    This fix allows us to actually see ssl connection events in the audit log from the pki /server perspective.
    This fill will also require support bug fixes for both pki and tomcatjss.

Based on original code contributed by Alex Ascheel, (cipherboy)."

This was a change made by @jmagne to fix bug: 2001169 for RHCS 10.2.1, Originally he only made the change on the `v4.9.x` branch but it needs to go to master as well. 

Tested locally by building and installing JSS, tomcatjss, and pki and observed the AUDIT logs:
```
0.https-jsse-nio-8443-exec-15 - [05/Nov/2021:15:31:17 GMT] [14] [6] [AuditEvent=ACCESS_SESSION_ESTABLISH][ClientIP=--][ServerIP=--][SubjectID=--][Outcome=Success] access session establish success
0.https-jsse-nio-8443-exec-16 - [05/Nov/2021:15:31:17 GMT] [14] [6] [AuditEvent=ACCESS_SESSION_TERMINATED][ClientIP=--][ServerIP=--][SubjectID=--][Outcome=Success][Info=serverAlertReceived: CLOSE_NOTIFY] access session terminated

```

### DIFF
--- a/src/main/java/org/mozilla/jss/ssl/SSLAlertEvent.java
+++ b/src/main/java/org/mozilla/jss/ssl/SSLAlertEvent.java
@@ -9,6 +9,7 @@ import java.util.EventObject;
 import javax.net.ssl.SSLException;
 
 import org.mozilla.jss.nss.SSLFDProxy;
+import org.mozilla.jss.ssl.javax.JSSEngine;
 
 public class SSLAlertEvent extends EventObject {
 
@@ -17,6 +18,7 @@ public class SSLAlertEvent extends EventObject {
     int level;
     int description;
 
+    transient JSSEngine engine;
     SSLAlertLevel levelEnum;
     SSLAlertDescription descriptionEnum;
 
@@ -57,11 +59,19 @@ public class SSLAlertEvent extends EventObject {
     }
 
     public SSLSocket getSocket() {
-        return (SSLSocket) getSource();
+        Object obj = getSource();
+	if( obj != null && obj instanceof SSLSocket) {
+            return (SSLSocket) obj;
+	}
+	return null;
     }
 
     public SSLFDProxy getFileDesc() {
-        return (SSLFDProxy) getSource();
+        Object obj = getSource();
+	if( obj != null && obj instanceof SSLFDProxy) { 
+            return (SSLFDProxy) getSource();
+        }
+        return null;
     }
 
     public int getLevel() {
@@ -98,6 +108,13 @@ public class SSLAlertEvent extends EventObject {
     public void setDescription(SSLAlertDescription description) {
         this.descriptionEnum = description;
         this.description = description.getID();
+    }
+
+    public JSSEngine getEngine() {
+        return engine;
+    }
+    public void setEngine(JSSEngine new_engine) {
+        engine = new_engine;
     }
 
     public SSLException toException() {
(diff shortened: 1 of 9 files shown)


edewata (approved):
LGTM

src/main/java/org/mozilla/jss/ssl/SSLAlertEvent.java
```diff
@@ -17,6 +18,7 @@
     int level;
     int description;
 
+    transient JSSEngine engine;
```
ckelleyRH, line 21: I'm not sure I understand why `engine` is `transient` here?

src/main/java/org/mozilla/jss/ssl/SSLAlertEvent.java
```diff
@@ -57,11 +59,19 @@ public SSLAlertEvent(SSLFDProxy proxy, SSLAlertLevel level, SSLAlertDescription
     }
 
     public SSLSocket getSocket() {
-        return (SSLSocket) getSource();
+        Object obj = getSource();
+	if( obj != null && obj instanceof SSLSocket) {
+            return (SSLSocket) obj;
+	}
+	return null;
```
ckelleyRH, line 66: This (and other instances like it) could be ternary operators if one was so inclined. Also you don't need to check for null, `instanceof` does this implicitly:

```
    public SSLSocket getSocket() {
        Object obj = getSource();
        return obj instanceof SSLSocket ? (SSLSocket) obj : null;
    }
```

src/main/java/org/mozilla/jss/ssl/SSLAlertEvent.java
cipherboy: Well, I wrote this originally in #807... ;-)

`SSLAlertEvent` is an existing class that extends [`EventObject`](https://docs.oracle.com/javase/7/docs/api/java/util/EventObject.html) which in turn implements `Serializable`. Since JSSEngine isn't serializable, `transient` marks it as such here in our `SSLAlertEvent` so that it doesn't try to serialize/deserialize it. In particular, across JVM instances/..., the `JSSEngine` isn't portable (in general, because it depends on NSS state that also isn't portable) and thus can't be serializable.

cipherboy: Ah, for context, this is the signature of `this.obj` (from `EventObject.getSource()`) -- [how the parent class is implemented](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/EventObject.java#L44-L47):

```java
    /**
     * The object on which the Event initially occurred.
     */
    protected transient Object source;
```

ckelleyRH: OK, thanks!

ckelleyRH (approved):
Now that I understand a bit more what is happening I approve, my other comment is tidying rather than functional.

cipherboy:
Should make sure to close #807 when you merge this one @c-dorney :-)

PR #824, opened 2021-11-08
Deprecate SHA-1

ckelleyRH:
`PKCS11Constants` has been left alone as it is auto-generated.


edewata (approved):
I think it should be `since=5.0.1`, but otherwise it LGTM. Feel free to cherry-pick these changes to `v5.0` branch and update the version number as well. Thanks!

edewata:
Uh.. @sonarcloud complained about missing `@deprecated` tags, but we can fix that separately.

c-dorney (approved):
LGTM

Issue #820, opened 2021-11-04
Azure CI broken for Fedora because Docker can't add Azure user to sudoers


ckelleyRH:
See [this](https://dev.azure.com/dogtagpki/jss/_build/results?buildId=53&view=logs&j=2cab6c39-d18e-5e41-04d4-faa08a8995cc&t=818d1a9c-16a8-4b09-a0b0-4813da9fa1e2&l=48) for an example failure, it has only started happening in the last week.

* There were successful runs after the last change to `azure-pipelines.yml`
* I couldn't see any obvious difference between hosts/Docker versions/etc, but not my area of expertise
* It started before the release of F35

@edewata any ideas?

PR #831, opened 2021-11-23
Fix Azure pipelines

edewata:
The build tests in Azure pipelines have been updated to provide `docker` in the container for installing the dependencies as `root`, so it's no longer necessary to use `sudo`.

Resolves: #820


ckelleyRH (approved):
:tada: LGTM!

edewata:
@ckelleyRH Thanks!

PR #833, opened 2022-01-17
Related: Bug 1964176 - KRA PKCS12 support for nCipher sw v12.72+

jmagne:
Found an issue which was causing the need to add the "explicitness" flag
to the cknfracstrc file.

With this fix that flag will no longer be required.


ladycfu (approved):
LGTM. As I understand it,  the key being generated here is on there for a short time to help move the "actual" aes session key from the software token to the hardware token, and therfore it itself does not need to be extractable.  I was told that it is tested to work.

PR #834, opened 2022-01-19
Don't hold onto threads after they're gone

uplogix-mmcclain:
If threads are dynamically created and destroyed and
CryptoManager.setThreadToken() is called, a Hashtable will hold onto
references of Threads long after they should be garbage collected.


cipherboy (commented):
Makes sense to me. Definitely agree that `perThreadTokenTable` would persist thread references, preventing the GC as noted in OP's PR description.

cipherboy (approved):

edewata (approved):
@uplogix-mmcclain @cipherboy LGTM, thanks!

PR #835, opened 2022-01-19
X509Certificate.getSubjectDN()and getIssuerDN() leak memory

uplogix-mmcclain:


cipherboy (approved):
I'd imagine there's lots of these cases lying around JSS... Wonder if it makes sense to create a macro for the `NewStringAndFree` pattern?

uplogix-mmcclain:
Should I have called PORT_Free instead of free?

cipherboy:
[Functionally they're equivalent...](https://github.com/nss-dev/nss/blob/master/lib/util/secport.c#L169-L175); you'd only redefine `PORT_free` if you wanted a different allocator for portability. It doesn't really matter too much IMO, but I'd perhaps prefer it for tradition if nothing else.

fmarco76:
Hi @uplogix-mmcclain, this PR could be merged. Could you re-base the code. Thanks!

uplogix-mmcclain:
Hi @fmarco76.  It looks like git did what I intended.  It should cleanly merge now.

fmarco76:
Yes, sometime git is incredible!

PR #836, opened 2022-01-19
fix small memory leak with every open socket

uplogix-mmcclain:


cipherboy (commented):
I don't think this is quite right. `jsock_close` implements the PRIO methods. This needs to instead be:

```java
fd->dtor(fd)
```

Preferably with the asserts around it. See:

https://github.com/dogtagpki/jss/blob/master/src/main/java/org/mozilla/jss/ssl/javax/BufferPRFD.c#L97

Its a little hard to follow, but that's what occurs here: https://github.com/nss-dev/nss/blob/master/lib/ssl/ssldef.c#L209-L214 (which is also called by `ssl_SecureClose`). It then calls close on the underlying layer, which in turn must `dtor` itself. 

Note tho, that we rely on `dtor` being implemented for us by PRIO; I don't think its correctly implemented here:

https://github.com/dogtagpki/jss/blob/master/src/main/java/org/mozilla/jss/ssl/javasock.c#L989

So maybe the `PR_Free` from this PR suffices? But note the default implementation does a little more:

https://github.com/servo/nspr/blob/master/pr/src/io/prlayer.c#L21-L27
https://github.com/servo/nspr/blob/master/pr/src/io/prlayer.c#L453
https://github.com/servo/nspr/blob/master/pr/src/io/prlayer.c#L478

My 2c, but I'll let the others review and decide.

fmarco76:
Looking at the documentation you linked and the code I think you are quite right. The `PR_Free` seems not the correct way to release the memory, although it could reduce the memory leak.

> I don't think this is quite right. `jsock_close` implements the PRIO methods. This needs to instead be:
> 
> ```java
> fd->dtor(fd)
> ```
> 
> Preferably with the asserts around it. See:
> 
> https://github.com/dogtagpki/jss/blob/master/src/main/java/org/mozilla/jss/ssl/javax/BufferPRFD.c#L97
> 

@cipherboy on this I am wondering why after the `dtor` is not called the close like here:

[https://github.com/nss-dev/nss/blob/master/lib/ssl/ssldef.c#L214-L219](https://github.com/nss-dev/nss/blob/master/lib/ssl/ssldef.c#L214-L219)

cipherboy:
@fmarco76 Honestly, like the other PR, my memory is fast fading... From what I recalled, we do close the lower socket, but we don't need to call dtor explicitly (or visa-versa).

I think part of it was because the lower layer wasn't a real socket but was our file buffers, which I thought we cleaned up separately. The other part I seem to vaguely recall was fighting with this close/dtor ordering for a bit and settled on this as what worked.

I also seem to recall there was some case where aggressive freeing on close would cause issues, because some certs or similar were still referenced, but I don't know about that. Perhaps that was in the exception case; ref tracking should still mean certs can be referenced after the SSLEngine closes.

uplogix-mmcclain:
Someone a long time ago decided not to use the destructor which is why I patched jsock_close.

See in javasock.c (file appears to have been moved)
```
static void
jsockDestructor(PRFileDesc *fd)
{
    PR_ASSERT(!"This destructor shouldn't be called. close() does the work");
}
```

fmarco76:
@uplogix-mmcclain the problem is that you are trying to remove a layered structure and I am not sure your are cleaning everything in every case. I have seen the destructor not implemented and I do not know the reason but I will investigate.

I think this has to be better understood. We keep this PR open until we have a more clear idea and then we will decide.

fmarco76:
@uplogix-mmcclain I took some time to better review this PR. I think the PR_FREE should do the work but to avoid problems it is better to check the higher and lower level before to free. Something like in one of the above link:

https://github.com/servo/nspr/blob/master/pr/src/io/prlayer.c#L21-L27

As you already pointed out, dtor was not implemented and revert that decision is more complicated so we can limit the patch to the close method. Can you update this PR?

uplogix-mmcclain:
I'm out of office through the end of the year.
I'll take a look the first week of January.

Matthew

fmarco76:
> I'll take a look the first week of January.

I will be back the second week of January to review it.

uplogix-mmcclain:
The commit has been updated to follow the pl_FDDestructor pattern.

fmarco76:
@uplogix-mmcclain I think now is safe to merge. Thanks!

PR #837, opened 2022-01-19
WOULD_BLOCK_ERROR is not a buffer underflow

uplogix-mmcclain:
WOULD_BLOCK_ERROR just means the socket isn't ready to return data.

Prior to this fix I ran into 2 different issues.

On low latency networks (LAN), large posts or file uploads would hang
near the end of the transfer and report socket timeout exceptions.

On slower networks (VPN), large HTTP requests could silently fail with
the input truncated.

### DIFF
--- a/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -1197,6 +1197,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
         int this_src_write;
         int this_dst_write;
 
+        int error = 0;
         do {
             this_src_write = 0;
             this_dst_write = 0;
@@ -1222,7 +1223,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
 
             int max_dst_size = computeSize(dsts, offset, length);
             byte[] app_buffer = PR.Read(ssl_fd, max_dst_size);
-            int error = PR.GetError();
+            error = PR.GetError();
             debug("JSSEngine.unwrap() - " + app_buffer + " error=" + errorText(error));
             if (app_buffer != null) {
                 this_dst_write = putData(app_buffer, dsts, offset, length);
@@ -1264,7 +1265,11 @@ public class JSSEngineReferenceImpl extends JSSEngine {
         if (is_inbound_closed) {
             debug("Socket is currently closed.");
             handshake_status = SSLEngineResult.Status.CLOSED;
-        } else if (handshake_already_complete && src_capacity > 0 && app_data == 0) {
+        } else if ((handshake_already_complete) &&
+                   (error != PRErrors.WOULD_BLOCK_ERROR) &&
+                   (src_capacity > 0) &&
+                   (app_data == 0)) {
+
             debug("Underflowed: produced no application data when we expected to.");
             handshake_status = SSLEngineResult.Status.BUFFER_UNDERFLOW;
         }


cipherboy (changes requested):
Not quite sure I follow this one. Per [the definition of `BUFFER_UNDERFLOW`](https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLEngineResult.Status.html#BUFFER_UNDERFLOW):

> The SSLEngine was not able to unwrap the incoming data because there were not enough source bytes available to make a complete packet. 

That this occurs makes sense from your given scenarios: the application _using_ the SSLEngine needs to handle this case and provide more data to us. Notably, this only occurs after the handshake has already succeeded, so further packets (from the network) should be sized correctly before attempting to read them.

The bigger problem I think is that we have no heuristic prior to this to ensure the passed `src` buffer is adequately sized, per packet contents. This shouldn't require parsing the TLS protocol (or access to the secret keys)... Instead, we just need to parse the protocol header(s) and ensure we have at least [one full fragment](https://datatracker.ietf.org/doc/html/rfc5246#section-6.2.1).

I think this'd look like:

```java
...
// Order of operations:
// 1. Ensure we have enough data to parse at least one entire packet.
...

// Check available fragments to ensure src is sized properly.
{
    int starting_src_pos = src.position();

    do {
        int fragment_start = src.position();
        byte fragment_type = src.get();
        validate_fragment_type(fragment_type); // XXX Implement me
        // Skip the next two bytes as they're the protocol version and we're
        // going to leave that to NSS for now.
        src.get(); src.get();
        // Now validate the length.
        int fragment_length = src.getShort();
        if (src.remaining() < fragment_length) {
            // This is the underflow case; remove the latter if.
            return new SSLEngineResult(SSLEngineResult.Status.BUFFER_UNDERFLOW, handshake_state, wire_data, app_data);
            // If you implement the suggestion below (about calling into NSS)
            // while parsing fragments, you'd turn this into a:
            /*
            handshake_status = SSLEngineResult.Status.BUFFER_UNDERFLOW;
            break;
            */
            // and then close this if statement and handle the call into NSS. The
            // `else if` statements below would then occur after the call into NSS;
            // note that you'd have to be careful to reset `src.position(fragment_start)`
            // prior to the call into NSS. :-) In essence, the latter do...while would be
            // subsumed into this one, right here. :-)
        } else if (src.remaining() == fragment_length) {
            // No more application data possible; return.
            break;
        } else {
            // Update our position and try to parse the next fragment.
            src.position(fragment_start + fragment_length);
        }
    } while(true);

    // Reset our source position (only necessary in the case of.
    src.position(starting_src_position);
}

do {
...
```

Now, the risk we run here (with the above) is that we'll keep getting (more) partial packets. In which case, we actually want to parse out full fragments and provide NSS one (full) fragment at a time. Left as an exercise to the reader, but its the approach I would prefer, assuming it is within SSLEngine specs ;-)

Then we can make progress (non-empty `wire_data` and `app_data`) but still potentially return `BUFFER_UNDERFLOW` status I think.  

Thoughts? But that's my 2c. not having looked at this in a while; perhaps @edewata or @jmagne disagree though.

uplogix-mmcclain:
For context, the application is Tomcat 8.5 with TomcatJSS 7.7.1 (OpenJDK 8).
I did not see this behavior with the regular JSSE SSL Tomcat connector.

cipherboy:
Right, this doesn't occur in JSSE because it parses the packets fully. You can see this here:

1. Total srcs == zero: https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/sun/security/ssl/SSLEngineImpl.java#L596-L604 
2. Or, parse the packet and check we have enough: https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/sun/security/ssl/SSLEngineImpl.java#L681-L684 (with the code above actually performing the calculations).

The present heuristic in our JSS implementation is flawed; rather than adjusting the heuristic, I think we should instead update it to be a valid, deterministic algorithm instead. My 2c.

jmagne:
Just taking a quick look at this, it appears that cipherboy is just wanting to do the incoming buffer size integrity check at the top of the routine, instead of waiting to call nss in the middle of processing.

rjrelyea:
I wonder if this problem is causing the issue in https://bugzilla.redhat.com/show_bug.cgi?id=1921247 where we are busy waiting hang on a socket in tomcat?

cipherboy:
@rjrelyea No, I don't believe so as this is in the `SSLEngine` code (which yes, does place NSS in non-Blocking mode).

The BZ you mentioned was against RHEL 7.9 and thus doesn't use this code path (only in RHEL 8.2+ or maybe RHEL 8.3+ iirc). 

There might still be bugs in `SSLSocket` (which is used with Tomcat's BIO adapter that was deprecated from RHEL 8+), but note that in that RHEL 7.9 code, NSS owns the raw network socket, not Tomcat (like Tomcat does with `SSLEngine` on RHEL 8+ and here we give NSS a "fake"/buffer-backed socket). I think we might still place it in non-blocking mode; perhaps a hotfix forcing the socket to be in blocking mode might be worthwhile?

rjrelyea:
Being in non-blocking mode is fine as long as you can figure out that 1) the socket is done, and 2) you aren't busy waiting on the socket. I think the latter is part of the problem.

cipherboy:
@rjrelyea Wouldn't `ssl_Recv` return an error if the socket is closed? I think for us to be hot-looping there, we'd need to be calling `Java_org_mozilla_jss_ssl_SSLSocket_socketRead` which either that's ignoring an error (looks fine to me: https://github.com/dogtagpki/jss/blob/master/src/main/java/org/mozilla/jss/ssl/SSLSocket.c#L931-L962) -- or perhaps NSS isn't informing us for some reason?

Also, how are you sure that the socket in that BZ is non-blocking? AFAICT, the old `SSLSocket` code has no way of setting non-blocking... and I'd assuming blocking is the default?

uplogix-mmcclain:
I was testing @cipherboy's fixes in #838 and found that just changing JSSEngine.BUFFER_SIZE (to 5 + 1024 + 1024 + 20 + 256 + (1 << 14)) is enough to make the issue I reported in #837 go away.  I don't know how some of those constants are derived; they were copied from org.apache.tomcat.util.net.openssl.OpenSSLEngine.MAX_ENCRYPTED_PACKET_LENGTH.

At the very least, I have not been able to reproduce it with the manual tests I previously ran.

uplogix-mmcclain:
See #865
It includes @cipherboy's fix to JSSEngine.BUFFER_SIZE

PR #838, opened 2022-01-27
Only send fully-formed record fragments to NSS

cipherboy:
In the SSLEngine's `unwrap(...)` method, we previously let NSS handle all
validation on the source `ByteBuffer`. While this limits the amount of TLS
protocol knowledge `JSSEngine` needs to have, it has the unfortunate side
effect of not always being able to expose the required
`WOULD_BLOCK_ERROR` status to the application; see the related pull
request below for a discussion on why this is problematic.

This change instead parses `src` for full record fragments, passing only
completed fragments to NSS. This ensures NSS never blocks waiting for
more data (as upon getting a full record, should release back to the
network layer to fetch more data) -- and also ensure applications are
properly informed when the network hasn't provided enough data to parse
a record fragment.

See also: https://github.com/dogtagpki/jss/pull/837

`Signed-off-by: Alexander Scheel <alexander.m.scheel@gmail.com>`

--- 

This needs to be tested against Tomcat; @jmagne can you try a Dogtag deployment with this and see if [`testssl.sh`](https://testssl.sh/) or `sslscan` cause any high CPU usage when run against it?

@uplogix-mmcclain Could you take a look as well?

### DIFF
--- a/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -8,6 +8,7 @@ import java.nio.channels.WritableByteChannel;
 import java.nio.channels.Channels;
 import java.security.PublicKey;
 import java.nio.ByteBuffer;
+import java.util.Arrays;
 
 import javax.net.ssl.*;
 
@@ -131,6 +132,16 @@ public class JSSEngineReferenceImpl extends JSSEngine {
      * TrustManager instances, passing the result back to NSS.
      */
     private CertValidationTask task;
+
+    /**
+     * Used in unwrap(...) to hold fragments across calls to unwrap(...).
+     *
+     * The backing byte buffers might not have enough space for the entire
+     * fragment we've read from src. While we could resize them to be
+     * large enough (2^16 bytes), we instead store each fragment's data
+     * here between calls to ByteBuffer.Write(...) and also unwrap(...).
+     */
+    private byte fragment_data[] = null;
 
     public JSSEngineReferenceImpl() {
         super();
@@ -1143,6 +1154,21 @@ public class JSSEngineReferenceImpl extends JSSEngine {
         }
     }
 
+    private void validate_content_type(byte fragment_type) throws SSLException {
+        // See: https://datatracker.ietf.org/doc/html/rfc8446#section-5.1
+        // These are the same across SSL2.0 to TLSv1.3.
+        switch (fragment_type) {
+        case 0: // invalid
+        case 20: // change_cipher_spec
+        case 21: // alert
+        case 22: // handshake
+        case 23: // application_data
+            break;
+        default:
+            throw new SSLException("unknown record fragment header: " + fragment_type);
+        }
+    }
+
     @Override
     public SSLEngineResult unwrap(ByteBuffer src, ByteBuffer[] dsts, int offset, int length) throws IllegalArgumentException, SSLException {
         debug("JSSEngine: unwrap(ssl_fd=" + ssl_fd + ")");
@@ -1172,9 +1198,10 @@ public class JSSEngineReferenceImpl extends JSSEngine {
         logUnwrap(src);
 
         // Order of operations:
-        //  1. Read data from srcs
-        //  2. Update handshake status
-        //  3. Write data to dsts
+        //  1. Validate record fragments from srcs
+        //  2. Read a complete record fragment
+        //  3. Update handshake status
+        //  4. Write any data to dsts
         //
         // Since srcs is coming from the data, it could affect our ability to
         // handshake. It could also affect our ability to write data to dsts,
@@ -1197,18 +1224,90 @@ public class JSSEngineReferenceImpl extends JSSEngine {
         int this_src_write;
         int this_dst_write;
 
+        // We assume src begins on a fragment. This is a safe assumption given
+        // our invariant that we only process completed fragments. Incomplete
+        // fragments will yield a BUFFER_UNDERFLOW and be raised to the
+        // calling application to hand us more data; if there is no more, an
+        // IOException will likely be thrown when the connection terminates.
+        // At each stage, we leave src.position() in a state wherein it
+        // should point to the start of where the next record fragment would
+        // be, if one is present.
         do {
             this_src_write = 0;
             this_dst_write = 0;
 
-            if (src != null) {
-                this_src_write = Math.min((int) Buffer.WriteCapacity(read_buf), src.remaining());
+            // Read one record fragment and give it to NSS when we have data.
+            if (src != null && src.hasRemaining() && fragment_data == null) {
+                if (src.remaining() < 5) {
+                    // At this point, we have _some_ data but not enough to
+                    // reconstruct a full record fragment header. Return a
+                    // BUFFER_UNDERFLOW error so we get more data.
+                    return new SSLEngineResult(SSLEngineResult.Status.BUFFER_UNDERFLOW, handshake_state, wire_data, app_data);
+                }
 
-                // When we have data from src, write it to read_buf.
+                // src position is later reset.
+                int fragment_start = src.position();
+
+                // From TLSv1.3's RFC 8446 Section 5.1 Record Layer (and
+                // corresponding RFCs for older protocol versions), each
+                // record fragment consists of:
+                //
+                //  - 1 byte fragment ContentType,
+                //  - 2 bytes of protocol version (unused as of TLSv1.3),
+                //  - 2 bytes of content length
+                //  - Up to 2^16 - 1 bytes of content.
+                //
+                // Fetch and validate the ContentType as a sanity check that
+                // we're at the start of a record.
+                byte fragment_type = src.get();
+                validate_content_type(fragment_type);
+
+                // Ignore the legacy protocol version; we leave it to NSS to
+                // validate this for us.
+                src.get(); src.get();
+
+                // Get and validate the fragment's length against remaining
+                // data in src.
+                int fragment_length = src.getShort();
+                if (src.remaining() < fragment_length) {
+                    // Since we have too little data remaining in our
+                    // fragment, reset to the start of the fragment and
+                    // return.
+                    src.position(fragment_start);
+                    return new SSLEngineResult(SSLEngineResult.Status.BUFFER_UNDERFLOW, handshake_state, wire_data, app_data);
+                }
+
+                // We know we have a complete record fragment in src. This
+                // means we can compute the total bounds of this fragment:
+                int fragment_end = fragment_start + 5 + fragment_length;
+                // and can read it into fragment_data. Note that we have to
+                // reset src's position so we get the fragment header as well.
+                fragment_data = new byte[fragment_end - fragment_start];
+                src.position(fragment_start);
+                src.get(fragment_data);
+
+                debug("JSSEngine.unwrap(): Found record fragment of size " + fragment_data.length + " bytes.");
+            }
+
+            if (fragment_data != null) {
+                // We have previously validated fragment data. Copy it into
+                // our write buffers. First compute the size of this write:
+                this_src_write = Math.min((int) Buffer.WriteCapacity(read_buf), fragment_data.length);
+
                 if (this_src_write > 0) {
-                    byte[] wire_buffer = new byte[this_src_write];
-                    src.get(wire_buffer);
+                    // Create a temporary buffer containing the bytes we know
+                    // will be read into the buffer.
+                    byte[] wire_buffer = Arrays.copyOfRange(fragment_data, 0, this_src_write);
 
+                    // Update our fragment_data, either setting it to null or
+                    // removing the bytes we've read.
+                    if (this_src_write < fragment_data.length) {
+                        fragment_data = Arrays.copyOfRange(fragment_data, this_src_write, fragment_data.length);
+                    } else {
+                        fragment_data = null;
+                    }
+
+                    // Perform the underlying byte buffer call.
                     this_src_write = (int) Buffer.Write(read_buf, wire_buffer);
 
                     wire_data += this_src_write;
--- a/src/main/java/org/mozilla/jss/ssl/javax/JSSEngine.java
+++ b/src/main/java/org/mozilla/jss/ssl/javax/JSSEngine.java
@@ -53,10 +53,13 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
     /**
      * Size of the underlying BUFFERs.
      *
-     * Helps to be large enough to fit most common SSL packets during the
-     * initial handshake.
+     * Helps to be large enough to fit well-formed SSL packets during the
+     * initial handshake and subsequent data transfer.
+     *
+     * See MAX_ENCRYPTED_PACKET_LENGTH calculation from Tomcat's
+     *     org.apache.tomcat.util.net.openssl.OpenSSLEngine.
      */
-    protected static int BUFFER_SIZE = 1 << 12;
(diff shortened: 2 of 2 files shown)


src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1197,18 +1214,97 @@ public SSLEngineResult unwrap(ByteBuffer src, ByteBuffer[] dsts, int offset, int
...
+
+                    // Update our fragment_data, either setting it to null or
+                    // removing the bytes we've read.
+                    if (this_src_write < fragment_data.length) {
+                        fragment_data = Arrays.copyOfRange(fragment_data, this_src_write, fragment_data.length);
+                    } else {
+                        fragment_data = null;
+                    }
 
+                    // Perform the underlying byte buffer call.
```
cipherboy, line 1307: We don't exit while `this_src-write > 0`, but if NSS gets stuck with a partial record fragment, I think perhaps we should reset `src`... Let me update that.

uplogix-mmcclain:
Just real quick feedback, I haven't looked at the code very closely...

I replaced my patch with this and then tested a largish POST to our app over a VPN.  The connection appeared to hang before oddly giving me a duplicate submission error (not sure what that was about).

Below is a partial stack from the relevant thread.

```
 "https-jsse-nio-443-exec-3" #65 daemon prio=5 os_prio=0 tid=0x00007f1cb4006800 nid=0x1f0d0 waiting on condition [0x00007f1d07ffc000]
    java.lang.Thread.State: TIMED_WAITING (parking)
       at sun.misc.Unsafe.park(Native Method)
       - parking to wait for  <0x000000008d69f748> (a java.util.concurrent.CountDownLatch$Sync)
       at java.util.concurrent.locks.LockSupport.parkNanos(LockSupport.java:215)
       at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(AbstractQueuedSynchronizer.java:1037)
       at java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1328)
       at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:277)
       at org.apache.tomcat.util.net.NioEndpoint$NioSocketWrapper.awaitLatch(NioEndpoint.java:1201)
       at org.apache.tomcat.util.net.NioEndpoint$NioSocketWrapper.awaitReadLatch(NioEndpoint.java:1203)
       at org.apache.tomcat.util.net.NioBlockingSelector.read(NioBlockingSelector.java:212)
       at org.apache.tomcat.util.net.NioSelectorPool.read(NioSelectorPool.java:279)
       at org.apache.tomcat.util.net.NioSelectorPool.read(NioSelectorPool.java:259)
       at org.apache.tomcat.util.net.NioEndpoint$NioSocketWrapper.fillReadBuffer(NioEndpoint.java:1334)
       at org.apache.tomcat.util.net.NioEndpoint$NioSocketWrapper.read(NioEndpoint.java:1276)
       at org.apache.coyote.http11.Http11InputBuffer.fill(Http11InputBuffer.java:808)
       at org.apache.coyote.http11.Http11InputBuffer.access$300(Http11InputBuffer.java:42)
       at org.apache.coyote.http11.Http11InputBuffer$SocketInputBuffer.doRead(Http11InputBuffer.java:1196)
       at org.apache.coyote.http11.filters.IdentityInputFilter.doRead(IdentityInputFilter.java:139)
       at org.apache.coyote.http11.Http11InputBuffer.doRead(Http11InputBuffer.java:265)
       at org.apache.coyote.Request.doRead(Request.java:670)
       at org.apache.catalina.connector.InputBuffer.realReadBytes(InputBuffer.java:325)
       at org.apache.catalina.connector.InputBuffer.checkByteBufferEof(InputBuffer.java:630)
       at org.apache.catalina.connector.InputBuffer.read(InputBuffer.java:348)
       at org.apache.catalina.connector.CoyoteInputStream.read(CoyoteInputStream.java:158)
       at org.apache.catalina.connector.Request.readPostBody(Request.java:3397)
       at org.apache.catalina.connector.Request.parseParameters(Request.java:3330)
       at org.apache.catalina.connector.Request.getParameterNames(Request.java:1214)
       at org.apache.catalina.connector.Request.getParameterMap(Request.java:1193)
       at org.apache.catalina.connector.RequestFacade.getParameterMap(RequestFacade.java:443)
```

The SSLEngine javadoc says, "Applications should call SSLSession.getApplicationBufferSize() and compare that value with the space available in the destination buffer, enlarging the buffer if necessary. Similarly, if unwrap() were to return a SSLEngineResult.Status.BUFFER_UNDERFLOW, the application should call SSLSession.getPacketBufferSize() to ensure that the source buffer has enough room to hold a record (enlarging if necessary), and then obtain more inbound data. "  

In this case the application is Tomcat, so JSSEngineReferenceImpl shouldn't be resizing buffers.

Also, the ByteBuffer interface was written to avoid creating a bunch of intermediate byte arrays.  So you should be able to use src.mark(), then read the content length, and check if the data is all there.  If it's all there, process the packet through NSS.  If it's not all there, then call reset() and return BUFFER_UNDERFLOW(?).

cipherboy:
Just to provide some context, maybe it'll spark some ideas... :-)

As you know, SunJSSE is the OpenJDK default SSLEngine provider. When configured with the PKCS#11 backend (with NSS), all cryptographic primitives will be performed by NSS (and thus, could achieve FIPS validation) but TLS is still handled by OpenJDK, not by NSS. This SSLEngine takes the other approach, allowing NSS to perform all the TLS protocol logic and cryptographic primitives.

This JSSEngine is similar to [Tomcat's OpenSSL "native" SSLEngine](https://github.com/apache/tomcat/blob/main/java/org/apache/tomcat/util/net/openssl/OpenSSLEngine.java) approach in that both NSS and OpenSSL implement interfaces over system io (the BIO stream abstraction in the OpenSSL case and PRIO from NSPR in NSS's case). `JSSEngineReferenceImpl` was a take on that using internal circular ring buffers to provide the data to NSS. I had hoped to write a more efficient implementation, allowing direct proxying of the ByteBuffers to NSS, but never got the time to work out the issues with reading/writing. I think [this](https://github.com/cipherboy/jss/tree/optimized-jssengine) was the branch but it never passed even basic tests. 

That said...

> In this case the application is Tomcat, so JSSEngineReferenceImpl shouldn't be resizing buffers.

I definitely agree that we shouldn't be resizing `src` -- Tomcat owns that buffer. But we should perhaps be resizing `read_buf` and `write_buf` to handle larger packets. Currently they're set at [`2^12` bytes](https://github.com/dogtagpki/jss/blob/master/src/main/java/org/mozilla/jss/ssl/javax/JSSEngine.java#L59). This is smaller than the [spec's](https://datatracker.ietf.org/doc/html/rfc8446#section-5.1) `2^14` bytes. Given a max-sized packet (5 + 2^14 bytes, [give or take](https://github.com/apache/tomcat/blob/main/java/org/apache/tomcat/util/net/openssl/OpenSSLEngine.java#L115-L123)), we'd need twice that overhead for read/write buffers.

The benefit of this resize is that we'd be guaranteed that well-formed packets would fit within the bounds of buffer and thus take a single call to `Buffer.Write(read_buf, ...)` to fill. 

I definitely agree that `SSLSession`'s [`sizes`](https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/SSLSession.html#getPacketBufferSize()) need to be updated. Currently there the [backing buffer's size](https://github.com/dogtagpki/jss/blob/master/src/main/java/org/mozilla/jss/ssl/javax/JSSEngine.java#L223) which is wrong.

uplogix-mmcclain:
With the second change to JSSEngine.BUFFER_SIZE, this now appears to work.

1. fragment_length needs to be sanity checked or there's a potential security issue here when fragment_length < 0 (Java shorts are signed) or fragment_length > MAX_RECORD_LENGTH (BUFFER_SIZE, if that's the right sanity limit).  I don't know all the rules of TLS to know if exceptions are to be alerted in a special way.

2. Is it really a BUFFER_UNDERFLOW, if unwrap has already handled 1 full packet?


3. Since BUFFER_UNDERFLOW is handled up front, unwrap doesn't appear to need the check after the while loop.

I'm referring to this:
```
} else if (handshake_already_complete && src_capacity > 0 && app_data == 0) {
```


4. With the resized org.mozilla.jss.nss.Buffer, I don't think you need the intermediate fragment_data byte array.  If it is possible that read_buf doesn't have capacity, that would still need some logic.

This could use more cleanup, but here's what I'm thinking:
```
int fragment_length = src.getShort();
if (src.remaining() < fragment_length) {
    logger.error("UNDERFLOW: " + src.remaining() + " < " + fragment_length);
    // Since we have too little data remaining in our fragment,
    // reset to the start of the fragment and return.
    src.position(fragment_start);
    return new SSLEngineResult(SSLEngineResult.Status.BUFFER_UNDERFLOW, handshake_state, wire_data, app_data);
}

src.position(fragment_start);

final int recordLength = fragment_length + 5;
final long writeCapacity = Buffer.WriteCapacity(read_buf);
if (writeCapacity < recordLength) {
    logger.error("writeCapacity < recordLength: " + writeCapacity + " < " + recordLength);
    // TODO: is this possible?  should we throw an exception here?
} else {
    byte[] wire_buffer = new byte[recordLength];
    src.get(wire_buffer);
  
    // Perform the underlying byte buffer call.
    this_src_write = (int) Buffer.Write(read_buf, wire_buffer);
    
    // TODO: we already checked writeCapacity, but can
    // this_src_write ever be < fragment_length?
    
    wire_data += this_src_write;
    debug("JSSEngine.unwrap(): Wrote " + this_src_write + " bytes to read_buf.");
}
```


Outside the scope of this pull request:

. I don't think unwrap needs the do/while loop.  If a single call to unwrap handled a single packet, wouldn't the caller just call it again if the return was "OK" and there was unhandled data in the buffers?

It appears to work without it and the code could potentially get simpler and be easier to optimize.  Of course, wrap() performance is probably more important than unwrap() for normal web applications.


. I don't know if it's deliberate or not, but much of JSS doesn't appear to follow Java standard naming conventions.  That's actually what most of the sonarcloud "code smells" are about.

https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html

If BUFFER_SIZE was final, that complaint would probably go away.

uplogix-mmcclain:
After further testing, it appears that just changing JSSEngine.BUFFER_SIZE is enough to make the issue I reported in #837 go away.

At least, I have not been able to reproduce it with the tests I previously ran.

Limiting calling into NSS only when full records are received is probably a good performance improvement: JNI calls are expensive.

cipherboy:
@uplogix-mmcclain So what's your thoughts then? Are these changes (+ your suggestions) worthwhile? Or should we leave it as strictly updating `BUFFER_SIZE`?

@edewata, @jmagne, @ladycfu thoughts or preferences from a Dogtag PoV?

uplogix-mmcclain:
These are just my thoughts.  I don't know what workflow and process is preferred by the team in charge.

I think #837 should either be closed/rejected or updated to only set the BUFFER_SIZE.  If it's closed/rejected, a new pull request with only the BUFFER_SIZE change should be created and in the description mention the 2 issues I found that #837 addressed.

I'm not fond of reimplementing NSS parsing code, but #838 may be worthwhile.  My biggest concerns are creating new bugs that didn't exist before (particularly around handling records of invalid size).  I think it may simplify the unwrap code and possibly be a performance boost if fewer JNI calls are made, but just removing the do/while and not parsing the packet header would probably also result in fewer JNI calls.

I think org.apache.tomcat.util.net.openssl.OpenSSLEngine.MAX_ENCRYPTED_PACKET_LENGTH may be valid for TLS 1.2, but is probably invalid for TLS 1.3.  If so, this code should handle both scenarios.

From https://datatracker.ietf.org/doc/html/rfc8446#section-5.2
>   length:  The length (in bytes) of the following
>      TLSCiphertext.encrypted_record, which is the sum of the lengths of
>      the content and the padding, plus one for the inner content type,
>      plus any expansion added by the AEAD algorithm.  The length
>      MUST NOT exceed 2^14 + 256 bytes.  An endpoint that receives a
>      record that exceeds this length MUST terminate the connection with
>      a "record_overflow" alert.

cipherboy:
> I think `org.apache.tomcat.util.net.openssl.OpenSSLEngine.MAX_ENCRYPTED_PACKET_LENGTH` may be valid for TLS 1.2, but is probably invalid for TLS 1.3. If so, this code should handle both scenarios.

If by invalid you mean oversized, I agree (by `1024 + 1024 + 20` bytes). However, regardless of this code, NSS will still parse the packet and handle that error case itself, sending out a fatal alert, I believe, which will then get surfaced as an Exception here. So we should be good.

uplogix-mmcclain:
Yes (it hadn't occurred to me), if record length is > ((1 << 14) + 256) but <= (1024 + 1024 + 20 + 256 + (1 << 14)), the Java code could just pass the record to NSS and NSS would handle it fine.  If the record length is > (1024 + 1024 + 20 + 256 + (1 << 14)), the Java code would need to do the appropriate thing to reject the record and cause the record_overflow alert to be sent (not sure if it's supposed to fail the same way with TLS 1.2).

cipherboy:
Ahh good point about needing to trigger the `record_overflow` alert... This means we'd need to set the state to `need_wrap`, wait for `wrap(...)`, construct and send the alert, shutdown NSS, and then raise the exception. Lovely! :D

fmarco76:
@cipherboy @uplogix-mmcclain I have reviewed this PR, the #837 and the #865 to understand the problem. If I got correctly increasing the buffer size as suggested in #865 should eliminate the fragmentation problem because following the specification the fragment should not exceed  that size. However, the notification for `record_overflow` could be useful.

If I get it right then I would suggest to accept the #865 (after an update) which change only the size. Close this PR because the change become not relevant and open a new github issue referring this discussion and the overflow problem. Is it OK with you?

uplogix-mmcclain:
@fmarco76 A version/variation on this pull request might still be useful for future performance improvements.

Without handling the edge cases around illegal record sizes, I don't think this is ready to be merged.

NSS currently handles everything to do with parsing the TLS records including creating `record_overflow` alerts.  Once the Java code starts to do minimal parsing to handle lengths it has to also handle the edge cases where a malicious client may try to send a record that is too large.

Don't confuse the record_overflow alert with SSLEngineResult.Status.BUFFER_UNDERFLOW.  (You may not be.  I'm unsure how to interpret, "However, the notification for `record_overflow` could be useful.")

cipherboy:
Yeah, I think @uplogix-mmcclain is correct; its probably best to merge #865 for the time being.

The worry I still have is that we'd want to validate lengths here, but then use copy-free buffers for the read/write operations. I don't think we have an easy way of doing that, so perhaps its easier just to merge the other PR and table this one...

As it is, I think we allocate more than the amount of buffers we need: 

 - 1x in Tomcat for copy to/from socket and JSS
 - 1x in JSS, to copy to/from NSS,
 - 1x in NSS, to handle smaller fragments getting read.

Ideally, we'd remove the middle one, and sync Tomcat<->NSS more closely. This change increases the memory a bit (for the first and second steps -- now Tomcat must fetch a completed packet before we'll send it to NSS and we must have space for that). I believe the other PR merely increases the middle size.

However, I believe the correctness of this change is probably ideal, given NSS might not have semantics for partially formed TLS buffers and I worry about stalling without fully-formed packets. Hopefully we don't run into that scenario though...

My 2c., but I'd definitely merge the other PR.

fmarco76:
@cipherboy and @uplogix-mmcclain my plan was to merge the #865. For this PR I was proposing to close because the problem highlighted I do not think will be solved here. If this is the case I would add an issue referring to this discussion which will be reviewed  in the future. Nevertheless, if you think it is better to keep this for the moment we can merge the #865, which seem all agree, and leave this open.

uplogix-mmcclain:
I have no objections to closing #838 as won't fix.

PR #840, opened 2022-01-31
Remove SHA-1 from allowed RSA/EC signing algorithms

ckelleyRH:


edewata:
LGTM, but I'll have to defer to @ladycfu or @jmagne.

ladycfu (approved):
In theory this does the job, as it would completely cut off any support for SHA1 wherever pki code uses these static variables.  I'll approve, but make sure jmagne ok's it.

ckelleyRH:
Awesome, thanks @ladycfu @jmagne!

PR #842, opened 2022-02-09
Fix memory leak on each TLS connection

ckelleyRH:
Each TLS connection is leaking a bunch of data that isn't in the heap
and so after 25k requests Tomcat uses about 2.5GB resident memory.

There are large number of relationships that point at each other and we
need to break the cycle so JSSEngineReferenceImpl's finalizer can run
and clear all the native resources these point at.

The lowest impact place to break the cycle was at SSLAlertEvent.engine.
This relationship doesn't seem to be used anywhere.  Once the cycle is
broken, JSSEngineReferenceImpl can be garbage collected and the
finalizer can run.

Signed-off-by: Chris Kelley <ckelley@redhat.com>

### DIFF
--- a/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -4,17 +4,39 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.ServerSocket;
 import java.net.Socket;
-import java.nio.channels.WritableByteChannel;
-import java.nio.channels.Channels;
-import java.security.PublicKey;
 import java.nio.ByteBuffer;
+import java.nio.channels.Channels;
+import java.nio.channels.WritableByteChannel;
+import java.security.PublicKey;
 
-import javax.net.ssl.*;
+import javax.net.ssl.SSLEngineResult;
+import javax.net.ssl.SSLException;
+import javax.net.ssl.SSLHandshakeException;
+import javax.net.ssl.X509ExtendedTrustManager;
+import javax.net.ssl.X509TrustManager;
 
-import org.mozilla.jss.nss.*;
-import org.mozilla.jss.pkcs11.*;
-import org.mozilla.jss.provider.javax.crypto.*;
-import org.mozilla.jss.ssl.*;
+import org.mozilla.jss.nss.BadCertHandler;
+import org.mozilla.jss.nss.Buffer;
+import org.mozilla.jss.nss.BufferProxy;
+import org.mozilla.jss.nss.Cert;
+import org.mozilla.jss.nss.CertAuthHandler;
+import org.mozilla.jss.nss.PR;
+import org.mozilla.jss.nss.PRErrors;
+import org.mozilla.jss.nss.PRFDProxy;
+import org.mozilla.jss.nss.SSL;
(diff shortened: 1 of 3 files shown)


src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1638,10 +1660,10 @@ private void cleanupSSLFD() {
         if (!closed_fd && ssl_fd != null) {
             try {
                 SSL.RemoveCallbacks(ssl_fd);
-                PR.Close(ssl_fd);
```
cipherboy, line 1641: My understanding at the time of writing this was that `PR.Close(...)` freed all memory at the NSS SSL PRFileDesc wrapper level, and which in turned freed the underlying transit PRFileDesc (here, our circular buffer PRFileDesc wrapper) as well. 

By removing `PR.Close(ssl_fd)`, but adding `super.(...)` to `ssl_fd.close()`, you effectively [do the same thing](https://github.com/dogtagpki/jss/blob/master/src/main/java/org/mozilla/jss/nss/PRFDProxy.java#L10), just in a different order. I'm not really sure it matters much.

src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1638,10 +1660,10 @@ private void cleanupSSLFD() {
...
                 ssl_fd.close();
+                ssl_fd = null;
```
cipherboy, line 1664: This step I agree with.

jmagne:
I agree with what cipherboy has said. Before I delve even further, I had a quesiton. Have we been given or tried a test case where the problem happens without the fix and then after applying the fix, the leak is lessened or eliminated?

src/main/java/org/mozilla/jss/util/GlobalRefProxy.java
```diff
@@ -12,7 +12,5 @@ public GlobalRefProxy(Object target) {
     private static native byte[] refOf(Object target);
 
     @Override
-    protected synchronized void releaseNativeResources() {
-        clear();
-    }
+    protected native void releaseNativeResources();
```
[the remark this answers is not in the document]
jmagne: Yes, it looks like the version of clear in NativeProxy.java  simple nulls out a pointer, where the native releaseNativeREsources, deletes a global ref.

ckelleyRH:
> I don't particularly have the time to do a full review but figured I'd drop what thoughts I have off the top of my head.

No worries @cipherboy thanks for taking a look!

> I agree with what cipherboy has said. Before I delve even further, I had a quesiton. Have we been given or tried a test case where the problem happens without the fix and then after applying the fix, the leak is lessened or eliminated?

Yes @jmagne, I have shared the reproduction details with you via email.

edewata (approved):
@uplogix-mmcclain @ckelleyRH @cipherboy @jmagne Thanks everyone! I'm not really familiar with this area of the code, but so far I don't see anything objectionable. I think we should just merge this patch to all branches and let it be tested by the Fedora community.

There are some CI failures, but I don't think they are caused by this patch. The failure in `v5.0` branch already existed before this patch:
https://github.com/dogtagpki/jss/actions?query=branch%3Av5.0

PR #843, opened 2022-02-09
Fix memory leak on each TLS connection

ckelleyRH:
Each TLS connection is leaking a bunch of data that isn't in the heap
and so after 25k requests Tomcat uses about 2.5GB resident memory.

There are large number of relationships that point at each other and we
need to break the cycle so JSSEngineReferenceImpl's finalizer can run
and clear all the native resources these point at.

The lowest impact place to break the cycle was at SSLAlertEvent.engine.
This relationship doesn't seem to be used anywhere.  Once the cycle is
broken, JSSEngineReferenceImpl can be garbage collected and the
finalizer can run.

Signed-off-by: Chris Kelley <ckelley@redhat.com>

### DIFF
--- a/src/main/java/org/mozilla/jss/nss/SSLFDProxy.java
+++ b/src/main/java/org/mozilla/jss/nss/SSLFDProxy.java
@@ -1,6 +1,5 @@
 package org.mozilla.jss.nss;
 
-import java.lang.IllegalArgumentException;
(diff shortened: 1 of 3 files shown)


ckelleyRH:
I have verified this fix against F35.

edewata (approved):

PR #847, opened 2022-02-16
Replace deprecated X509Certificate.getSubjectDN() and getIssuerDN()

edewata:
The `X509Certificate.getSubjectDN()` and `getIssuerDN()` have been replaced with `getSubjectX500Principal()` and `getIssuerX500Principal()`, respectively, except where the values are converted into `X500Name`.


ckelleyRH (approved):
LGTM!

Small question: some places you have removed `toString()` and some places you didn't - can we remove it from the other places too?

edewata:
@ckelleyRH Thanks!

We can only remove `toString()` if it's part of `String` concatenation since Java will call the method automatically so it's redundant. In other places there's no `String` concatenation, but the value needs to be a `String` so we keep the `toString()`.

PR #848, opened 2022-02-25
Add build.sh options for RPM build

edewata:
The `build.sh` has been modified to provide options such that the RPM build can be done according to the RPM specification.
It will also set the environment variables for CMake as defined in `/usr/lib/rpm/macros.d/macros.cmake`.

### DIFF
--- a/jss.spec
+++ b/jss.spec
@@ -109,6 +109,8 @@ This package contains the API documentation for JSS.
 %build
 ################################################################################
 
+# Set build flags for CMake
+# (see /usr/lib/rpm/macros.d/macros.cmake)
 %set_build_flags
 
 export JAVA_HOME=%{java_home}
@@ -126,7 +128,12 @@ modutil -dbdir /etc/pki/nssdb -chkfips true | grep -q enabled && export FIPS_ENA
 ./build.sh \
     %{?_verbose:-v} \
     --work-dir=%{_vpath_builddir} \
+    --prefix-dir=%{_prefix} \
+    --include-dir=%{_includedir} \
     --lib-dir=%{_libdir} \
+    --sysconfig-dir=%{_sysconfigdir} \
+    --share-dir=%{_datadir} \
+    --cmake=%{__cmake} \
     --java-home=%{java_home} \
     --jni-dir=%{_jnidir} \
     --version=%{version} \
--- a/build.sh
+++ b/build.sh
@@ -12,11 +12,19 @@ SRC_DIR="$(dirname "$SCRIPT_PATH")"
 NAME=jss
 WORK_DIR="$HOME/build/$NAME"
 
+PREFIX_DIR="/usr"
+INCLUDE_DIR="/usr/include"
+
 if [ "$HOSTTYPE" = "x86_64" ]; then
    LIB_DIR="/usr/lib64"
 else
    LIB_DIR="/usr/lib"
 fi
+
+SYSCONFIG_DIR="/etc"
+SHARE_DIR="/usr/share"
+
+CMAKE="cmake"
 
 JNI_DIR="/usr/lib/java"
 INSTALL_DIR=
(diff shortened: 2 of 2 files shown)


ckelleyRH (commented):
> The build.sh has been modified to provide options such that the RPM build can be done according to the RPM specification.

I don't understand what is going on with this change - could you outline what you're trying to do please?

edewata:
Ideally in Fedora we're supposed to use `%cmake` macro as described in [CMake Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/CMake/) to ensure that CMake is called with the proper parameters for Fedora.

In older branches we used to call the `%cmake` directly in [jss.spec](https://github.com/dogtagpki/jss/blob/v4.9.x/jss.spec#L124-L129):
```
%cmake \
    -DVERSION=%{version} \
    -DJAVA_HOME=%{java_home} \
    -DJAVA_LIB_INSTALL_DIR=%{_jnidir} \
    -DJSS_LIB_INSTALL_DIR=%{_libdir}/jss \
    -B %{_vpath_builddir}
```
and the `build.sh` would call `rpmbuild` which used the `jss.spec` to create the build (i.e. RPM packages), but that would only work on Fedora.

In newer branches we're trying to support more platforms, so the `build.sh` is calling CMake directly to create a more generic build (i.e. tarball), but `jss.spec` can still call the `build.sh` to create the RPM package:
```
./build.sh \
    %{?_verbose:-v} \
    --work-dir=%{_vpath_builddir} \
    --lib-dir=%{_libdir} \
    --java-home=%{java_home} \
    --jni-dir=%{_jnidir} \
    --version=%{version} \
    %{!?with_javadoc:--without-javadoc} \
    %{?with_tests:--with-tests} \
    dist
```
However, right now the `build.sh` is calling CMake without the proper parameters for Fedora.

This PR is supposed to fix that without dropping the support for other platforms. It's trying to emulate what the `%cmake` macro does, i.e. defining some `CMAKE_*` variables. So `jss.spec` can call `build.sh` with Fedora-specific parameters to generate a Fedora-specific build like this:
```
./build.sh \
    %{?_verbose:-v} \
    --work-dir=%{_vpath_builddir} \
    --prefix-dir=%{_prefix} \
    --include-dir=%{_includedir} \
    --lib-dir=%{_libdir} \
    --sysconfig-dir=%{_sysconfigdir} \
    --share-dir=%{_datadir} \
    --cmake=%__cmake \
    --java-home=%{java_home} \
    --jni-dir=%{_jnidir} \
    --version=%{version} \
    %{!?with_javadoc:--without-javadoc} \
    %{?with_tests:--with-tests} \
    dist
```
but other platforms should still be able to call `build.sh` to generate a generic build.

jss.spec
```diff
@@ -126,7 +128,12 @@ modutil -dbdir /etc/pki/nssdb -chkfips true | grep -q enabled && export FIPS_ENA
 ./build.sh \
     %{?_verbose:-v} \
     --work-dir=%{_vpath_builddir} \
+    --prefix-dir=%{_prefix} \
+    --include-dir=%{_includedir} \
     --lib-dir=%{_libdir} \
+    --sysconfig-dir=%{_sysconfigdir} \
+    --share-dir=%{_datadir} \
+    --cmake=%__cmake \
```
ckelleyRH, line 136: Is there a reason that there are no curly braces here? I just tried a test on the command line to see what happens when you omit them and it turns out it gives the same value:
```
rpm --eval "some text printed on %{__cmake}"
some text printed on /usr/bin/cmake

rpm --eval "some text printed on %__cmake"
some text printed on /usr/bin/cmake
``` 

Could we standardise one way or the other though please, as it is certainly curious to the inexperienced observer!

ckelleyRH (approved):
Thanks for the explanation! The change seems OK but I'm not an expert here - I did catch one inconsistency though but I will approve anyway.

jss.spec
edewata: That's actually how the `%cmake` macro was defined (i.e. with inconsistent usage of curly braces), but I've changed it since it doesn't seem to cause any problem.

edewata:
@ckelleyRH Thanks!

PR #849, opened 2022-02-25
Add X509CertImpl.getSubjectName() and getIssuerName()

edewata:
The original `X509CertImpl.getSubjectDN()` and `getIssuerDN()` would get the subject name and issuer name (which are `X500Name` objects) then downcast them into `Principal` objects.

Since these methods are deprecated, new `getSubjectName()` and `getIssuerName()` methods have been added as replacements. These methods will return the subject name and issuer name directly as `X500Name` objects. They can be used to fix some deprecation warnings in PKI.

### DIFF
--- a/src/main/java/org/mozilla/jss/netscape/security/x509/X509CertImpl.java
+++ b/src/main/java/org/mozilla/jss/netscape/security/x509/X509CertImpl.java
@@ -81,6 +81,9 @@ import org.mozilla.jss.netscape.security.util.ObjectIdentifier;
  */
 public class X509CertImpl extends X509Certificate
         implements DerEncoder {
+
+    public static final org.slf4j.Logger logger = org.slf4j.LoggerFactory.getLogger(X509CertImpl.class);
+
     // Serialization compatibility with the X509CertImpl in x509v1.jar
     // supporting the subset of X509Certificate on JDK1.1.x platforms.
     static final long serialVersionUID = -2048442350420423405L;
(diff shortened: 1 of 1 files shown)


edewata:
@ckelleyRH Thanks!

Yeah, if we have a code like this:
```
X509CertImpl cert = ...
X500Name subjectName = (X500Name) cert.getSubjectDN();
```
we can replace that with:
```
X509CertImpl cert = ...
X500Name subjectName = cert.getSubjectName();
```

If we have a code like this:
```
java.security.cert.X509Certificate cert = ...
X500Name subjectName = (X500Name) cert.getSubjectDN();
```
I think we can safely replace it with this:
```
java.security.cert.X509Certificate cert = ...
X509CertImpl certImpl = (X509CertImpl) cert;
X500Name subjectName = certImpl.getSubjectName();
```

But if we have a code like this:
```
org.mozilla.jss.crypto.X509Certificate cert = ...
Principal subjectDN = cert.getSubjectDN();
```
I don't think that needs to be fixed since `org.mozilla.jss.crypto.X509Certificate` doesn't inherit from `java.security.cert.X509Certificate` so the method is not deprecated.

Ideally I would prefer if we can somehow merge `X509CertImpl` with `org.mozilla.jss.crypto.X509Certificate` and make it a subclass of `java.security.cert.X509Certificate`, but that could be tricky.

ckelleyRH:
Yeah that's the sort of refactor that would make things a bit cleaner but has the capacity to inflict significant pain if not handled carefully.

PR #850, opened 2022-03-03
Add doc for deprecated methods in X509CertImpl

edewata:
https://github.com/edewata/jss/blob/deprecation/docs/changes/v5.2.0/API-Changes.adoc

### DIFF
--- a/docs/changes/v5.2.0/API-Changes.adoc
+++ b/docs/changes/v5.2.0/API-Changes.adoc
@@ -1,0 +1,6 @@
+= API Changes =
+
+== Deprecate getSubjectDN() and getIssuerDN() in X509CertImpl ==
+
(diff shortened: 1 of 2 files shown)


ckelleyRH (approved):
Do we intend to remove these methods? If so can we have `@Deprecated(since="5.2.0", forRemoval=true)`? LGTM otherwise so approving.

edewata:
@ckelleyRH Thanks!

Right now we can't remove them since they are abstract methods in `java.security.cert.X509Certificate` and there's no plan to remove them from the base class:
https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/security/cert/X509Certificate.java#L278
https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/security/cert/X509Certificate.java#L237
But if Java removes them in the future we can remove them as well.

PR #852, opened 2022-03-24
Fix bugs identified by Sonar in x509 classes

ckelleyRH:
* Use try-with-resources in `CRLDistributionPoint`
* Put array designator on the type in `main` method
* Fix potential infinite loop by checking whether stream is empty.

### DIFF
--- a/src/main/java/org/mozilla/jss/netscape/security/x509/CRLDistributionPoint.java
+++ b/src/main/java/org/mozilla/jss/netscape/security/x509/CRLDistributionPoint.java
@@ -23,10 +23,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 
-import org.mozilla.jss.netscape.security.util.BitArray;
-import org.mozilla.jss.netscape.security.util.DerOutputStream;
-import org.mozilla.jss.netscape.security.util.DerValue;
-
 import org.mozilla.jss.asn1.ANY;
 import org.mozilla.jss.asn1.ASN1Template;
 import org.mozilla.jss.asn1.ASN1Util;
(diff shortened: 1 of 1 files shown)


src/main/java/org/mozilla/jss/netscape/security/x509/ExtensionsRequested.java
```diff
@@ -160,7 +160,7 @@ private void construct(DerValue dv) throws IOException {
         // now, the stream will be in the correct format
         stream.reset();
 
-        while (true) {
+        while (stream.available() != 0) {
```
edewata, line 163: I'm not sure about this. According to [Javadoc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/InputStream.html#available()) the method:
> Returns an **estimate** of the number of bytes that can be read (or skipped over) from this input stream **without blocking**, which may be 0, or 0 when end of stream is detected.

I think `0` availability doesn't necessarily mean that it's the end of stream. It could just mean that there's no data that can be read now, but there might be more coming later, so we probably shouldn't terminate the loop based on this.

We probably need to do more investigation. Right now the `DerInputStream.getDerValue()` always returns a non-null object. It probably needs to return a `null` if it reaches the end of stream so we can terminate the loop based on that.

edewata (commented):
I have one concern above, but everything else looks good. Maybe we should defer the change in `ExtensionsRequested` so we can merge this PR?

src/main/java/org/mozilla/jss/netscape/security/x509/ExtensionsRequested.java
ckelleyRH: > I think `0` availability doesn't necessarily mean that it's the end of stream. It could just mean that there's no data that can be read now, but there might be more coming later, so we probably shouldn't terminate the loop based on this.

Hmm, good point. I have reverted this part for now.

edewata (approved):
LGTM, thanks for the update!

ckelleyRH:
Thanks!

PR #853, opened 2022-03-24
Remove dead code/conditionals that can only ever take one value

ckelleyRH:


src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
```diff
@@ -311,10 +315,6 @@ public boolean engineVerify(byte[] sigBytes)
 		assert(algorithm!=null);
 		assert(key!=null);
 
-		if(sigBytes==null) {
-			return false;
-		}
```
edewata, line 316: I think it's actually possible for `sigBytes` to be `null` here in case the assertion is not enabled, so we probably should keep this code.

edewata (commented):
I have one concern, but the rest looks good.

src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
ckelleyRH: I'm preparing another PR to remove use of `assert`, as you say you can disable them. They're also poor control flow as nobody is expecting to handle 'Error`.

edewata: Java assertions are actually [disabled by default](https://docs.oracle.com/cd/E19683-01/806-7930/assert-4/index.html). I think it's meant to be enabled during development to complement testing (because it may not be possible to test all scenarios). But considering we never explicitly enable assertions, and the assertions we have so far might not have been used properly, we probably can just remove them or replace them with regular code.

ckelleyRH: #854 <-Like so :smile:

edewata (approved):
I assume this is going to be merged after #854, which adds the check for null `sigBytes`.

ckelleyRH:
> I assume this is going to be merged after #854, which adds the check for null `sigBytes`.

Indeed, thanks!

PR #854, opened 2022-03-25
Replace use of assert in PK11Signature

ckelleyRH:
* `assert` is disabled by default and it is confusing to see it
used for control flow.

### DIFF
--- a/src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
+++ b/src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
@@ -13,8 +13,14 @@ import java.security.SecureRandom;
 import java.security.SignatureException;
 import java.security.spec.AlgorithmParameterSpec;
 import java.security.spec.PSSParameterSpec;
-import org.mozilla.jss.crypto.*;
-import org.mozilla.jss.util.*;
+
+import org.mozilla.jss.crypto.Algorithm;
+import org.mozilla.jss.crypto.DigestAlgorithm;
+import org.mozilla.jss.crypto.NoSuchItemOnTokenException;
+import org.mozilla.jss.crypto.PrivateKey;
+import org.mozilla.jss.crypto.SignatureAlgorithm;
+import org.mozilla.jss.crypto.TokenException;
+import org.mozilla.jss.util.NativeProxy;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


ckelleyRH:
In the code I was modifying there was a mixture of tabs and spaces. It is poor practice to make whitespace changes in a functional change, apologies for that, but I could barely read the method before I fixed it!

ckelleyRH:
I did a bit more tidying up to address the Sonar issues.

src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
```diff
@@ -200,29 +209,43 @@ public void engineUpdate(byte b)
...
+        }
+        if (token == null) {
+            throw new TokenException(NO_TOKEN);
+        }
+        if (tokenProxy == null) {
+            throw new TokenException(NO_TOKEN_PROXY);
+        }
+        if (algorithm == null) {
+            throw new SignatureException(NO_ALGORTIHM);
+        }
```
edewata, line 243: I think it's not necessary to check `token` and `algorithm` here since they are set in the constructor and never change. You might want to move the `tokenProxy` check into the constructor as well.

src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
```diff
@@ -232,17 +255,25 @@ protected native void engineUpdateNative(byte[] b, int off, int len)
...
```
edewata, line 273: Similarly, no need to check these fields again here.

src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
```diff
@@ -295,21 +325,30 @@ public int engineSign(byte[] outbuf, int offset, int len)
...
```
edewata, line 348: Similarly, no need to check these fields again here.

edewata (approved):
I think it generally looks good, but please see my comments. Feel free to update as needed then merge.

src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
ckelleyRH: Ack, `tokenProxy` check moved into the constructor.

src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
ckelleyRH: Ack.

src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
```diff
@@ -262,23 +293,22 @@ protected native void engineUpdateNative(byte[] b, int off, int len)
...
         }
-		if(	(outbuf==null) || (outbuf.length <= offset) ||
-			(len < sig.length) || (offset+len > outbuf.length))
-		{
-			throw new SignatureException(
-					"outbuf is not sufficient to hold signature");
-		}
-		System.arraycopy( sig, 0, outbuf, offset, sig.length);
-		return sig.length;
+        if ((outbuf.length <= offset) || (len < sig.length) || (offset + len > outbuf.length)) {
```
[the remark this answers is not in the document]
ckelleyRH: Ack.

src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
```diff
@@ -262,23 +293,22 @@ protected native void engineUpdateNative(byte[] b, int off, int len)
...
+            throw new SignatureException("outbuf is not sufficient to hold signature");
+            }
```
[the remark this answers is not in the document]
ckelleyRH: Ack.

src/main/java/org/mozilla/jss/pkcs11/PK11Signature.java
ckelleyRH: Ack.

ckelleyRH:
Thanks!

PR #855, opened 2022-03-25
Update SonarCloud analysis to Java 17

ckelleyRH:


ckelleyRH:
It would be nice if there were a way to avoid this, I forgot about this completely.

edewata (approved):
LGTM!

I'm not sure if we actually have Java 17-specific code in JSS, but if we do I suppose SonarCloud would have complained?

ckelleyRH:
> LGTM!
> 
> I'm not sure if we actually have Java 17-specific code in JSS, but if we do I suppose SonarCloud would have complained?

The change introduced 6 new issues, all for deprecation of classes between Java 11 and Java 17, so it did work. It would appear you are correct and no Java 12+ code has been introduced yet :+1:

Issue #832, opened 2021-12-14
CI broken due to NSS 3.73


edewata:
Currently the PKCS11 test is broken due to the new constants introduced in NSS 3.73:
https://github.com/dogtagpki/jss/runs/4515074676
https://github.com/nss-dev/nss/commit/e6c170e488bbca416f81a668274f4f447b4c1d38

One option is to add the new constants into `PKCS11Constants.java`, but it might not work on platforms that do not have NSS 3.73 yet.

Another option is to define `Conflicts:` in `jss.spec` against NSS 3.73 or later.

Another option is to let the CI fail until NSS 3.73 becomes more widely available.

cipherboy:
There's only a few places where JSS actually exposes a raw PKCS#11-ish interface (and hence, less used among our consumers and low risk of failure). These new constants are all vendor-specific too (`CK*_NSS_*` -- not prone to being used, especially by people just looking for a replacement for the thing the JDK removed) so from my point of view, 1 and 3 seem the best. My 2c :-)

PR #857, opened 2022-04-02
Update PKCS11Constants

edewata:
The PKCS11Constants class has been updated to include the new constants introduced in NSS 3.73.

Resolves: https://github.com/dogtagpki/jss/issues/832


ckelleyRH (approved):
LGTM. One question though - is there a way to enforce an NSS version range or do we just need to be careful?

edewata:
@ckelleyRH Thanks!

We can require a certain version of NSS here:
https://github.com/dogtagpki/jss/blob/master/jss.spec#L73
but NSS 3.73 is not yet available on all platforms that we officially support, so for now the warning needs to stay.

PR #859, opened 2022-05-17
Fix Bug 2013674 - JSS cannot be properly initialized after using anot…

jmagne:
…her NSS-backed security provider

The solution the the bug is to provide a way to initialize JSS such that the underlying nss system is initialized as a context
instead of a full NSS initialization. This allows jss / nss to create it's own nss init scenario. This could be of use where
the main process has already inited nss with a set of params and properties. Allowing say a library loaded into the man process the ability
to do a context based nss init, allows the library to set it's own nss params and not inherit those from the main process.
For instance if a main process has initialized nss to have a read only database, the context init will allow the library to load an nss
context with a read write database.

This is accomplished in a very simple manner with respect to jss. Right now jss allows one singleton instance to be created for a program.
This fix allows the caller to choose to have the singleton to be inited either as a context or a full init. Theoretically the nss context
concept would allow for multiple contexts to be created,but for simplicity, JSS will only allow one such context or full nss init.

The original behavior of doing a full NSS init beneath JSS persists as the default. In order to do a context init the following sample code will suffice:

CryptoManager.initializeWithContext(databaseDir);

manager = CryptoManager.getInstance();
...
...
manager.shutdown().

Notice that  an explicit call to shutdown is recommended because the NSS context inited must be destroyed at the native code level, in oder to not interfere with the main program's nss instance.

If one wants to explicitly specify the Initialization  values, the following call can be used as well:

InitializationValues vals =
                    new InitializationValues(args[0],
                            "", "", "secmod.db");

CryptoManager.initializeWithContext(vals);
CryptoManager manager = CryptoManager.getIntance();

...
...

manager.shutdwown();

Note: That as of this patch , there is no way to call for a context init when using the JCA provider interface. That can be a future improvement.

### DIFF
--- a/src/main/java/org/mozilla/jss/CryptoManager.c
+++ b/src/main/java/org/mozilla/jss/CryptoManager.c
@@ -58,7 +58,8 @@ static const char* VARIABLE_MAY_NOT_BE_USED DLL_NSS_VERSION     = "NSS_VERSION =
 /* NSPR_version from mozilla/nsprpub/pr/include/prinit.h */
 static const char* VARIABLE_MAY_NOT_BE_USED DLL_NSPR_VERSION    = "NSPR_VERSION = " PR_VERSION;
 
-
+/* One single nss context if caller chooses to intialize jss / nss with an nss context */
+static NSSInitContext *nssInitContext=NULL;
 
 
 static jobject
@@ -186,6 +187,21 @@ finish:
     JSS_DerefJString(env, ocspResponderCertNickname, ocspResponderCertNickname_string);
 
     return result;
+
+}
+
+/*Common shutdown code
+ */
+void ShutdownNSSOrContext()
+{
+    if (NSS_IsInitialized()) {
+        if (nssInitContext!=NULL) {
+            NSS_ShutdownContext(nssInitContext);
+            nssInitContext = NULL;
+        } else {
+            NSS_Shutdown();
+        }
+    }
 
 }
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/mozilla/jss/CryptoManager.c
```diff
@@ -517,6 +533,251 @@ Java_org_mozilla_jss_CryptoManager_initializeAllNative2
     return;
 }
 
+/* Create a complete version of the initialization method that
+ * launches NSS as a context instead of a full NSS_Init.
+ * this allows the app to assure they don't inherit all the properties
+ * of a previous full init done by the main process.
+ * Allow the caller of JSS to EITHER do a full nit OR do a context init.
+ */
```
ladycfu, line 541: Might be a good idea to state something like the following:
-----------
Note: This method is modified from a copy of an existing method: "Java_org_mozilla_jss_CryptoManager_initializeAllNative2";
The changes involved are the following:
1. It calls NSS_InitContext(...) instead of NSS_Initialize(...)
2.  The original method has some code to detect readOnly and calls special read only nss init functions, which don't make sense with a context. That logic was therefore removed.
The reason why copy/modify is done instead of making changes to the existing code was because XYZ.
We need to be aware in case there's any bug fixes in the original code that will need to be applied to this method as well

src/main/java/org/mozilla/jss/CryptoManager.c
```diff
@@ -517,6 +533,251 @@ Java_org_mozilla_jss_CryptoManager_initializeAllNative2
...
+    }
+    if( cooperate ) {
+        initFlags |= NSS_INIT_COOPERATE;
+    }
+
+    /*
+    * Initialize NSS Context.
+    */
+    nssInitContext = NSS_InitContext(szConfigDir, szCertPrefix, szKeyPrefix,
+            szSecmodName,NULL, initFlags);
```
ladycfu, line 716: looks like in the python-nss code, the second to last parameter depends on the init params
(see https://bugzilla.redhat.com/show_bug.cgi?id=2013674#c5)
                                        py_init_params ? &py_init_params->params : NULL,
Would you also need to provide that or is NULL good enough for your case?

src/main/java/org/mozilla/jss/CryptoManager.java
```diff
@@ -528,8 +549,40 @@ public static synchronized void initialize( InitializationValues values )
                     "Must set ocspResponderCertNickname");
             }
         }
+        //Right now, allow only the choice of a full NSS init or a NSS context init, usefule for libraries,
```
ladycfu, line 552: please fix the tab and the spelling of "useful" ;-)

src/test/java/org/mozilla/jss/tests/JSSContextInitTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+public class JSSContextInitTest {
+
+    public static void usage() {
+        System.out.println(
+                "Usage: java org.mozilla.jss.crypto.JSSContextInitTest <dbdir> <dbpw>" +
+                " [tokenname]");
+    }
+
+    //The purpose of this test is to simply attemt a represantitive JSS task such as Signing
```
ladycfu, line 36: representative

ladycfu (commented):
a couple simple suggestions, and one question to respond.

src/main/java/org/mozilla/jss/CryptoManager.c
jmagne: Taken care of in most recent push.

src/main/java/org/mozilla/jss/CryptoManager.c
jmagne: This has been resolved by actually providing an argument of NSSInitParams instead of making the PK11 call done before.

src/main/java/org/mozilla/jss/CryptoManager.java
jmagne: Done

src/test/java/org/mozilla/jss/tests/JSSContextInitTest.java
jmagne: Done

src/main/java/org/mozilla/jss/CryptoManager.c
```diff
@@ -517,6 +533,267 @@ Java_org_mozilla_jss_CryptoManager_initializeAllNative2
...
+    initparams.passwordRequired = PR_FALSE;
+	
+    szConfigDir = JSS_RefJString(env, configDir);
+    /*
+    * Set up arguments to NSS_InitContext
+    */
+    szCertPrefix = JSS_RefJString(env, certPrefix);
+    szKeyPrefix = JSS_RefJString(env, keyPrefix);
+    szSecmodName = JSS_RefJString(env, secmodName);
+
```
ladycfu, line 695: Shouldn't the above three szXXX param assignments still need to be under
"    if( certPrefix != NULL || keyPrefix != NULL || secmodName != NULL "
like the original code?

ladycfu (commented):
Looks better.  One additional comment to address. Thanks.

src/main/java/org/mozilla/jss/CryptoManager.c
jmagne: This ok for the following reason:

The routine JSS_RefJString simply returns NULL if the input is NULL. That check in the original function was done to see if we can use one of the simpler NSS_Init variants. In this case we have only one NSS_InitContext so we simply will pass NULL for those 3 values of interest.

ladycfu (approved):

PR #862, opened 2022-06-06
Update OpenLDAP clients to use -H option

edewata:
The latest OpenLDAP clients no longer have the `-h` option so the the tests have been updated to use the `-H` option instead.

https://github.com/dogtagpki/pki/wiki/DS-Installation

### DIFF
--- a/tests/bin/ds-create.sh
+++ b/tests/bin/ds-create.sh
@@ -13,7 +13,7 @@ sed -i \
 
 dscreate from-file ds.inf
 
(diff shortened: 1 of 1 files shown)


edewata:
There seems to be an existing problem with the build tests.

ckelleyRH (approved):
LGTM, we need to find time to fix the new NSS provider problem.

edewata:
@ckelleyRH Thanks!

Issue #861, opened 2022-05-23
SSLEngine test failures on F36.


ckelleyRH:
On `F36` the following tests fail on-build for `JSS`:
```
SSLEngine_RSA (Failed)
SSLEngine_ECDSA (Failed)
SSLEngine_RSA_FIPSMODE (Failed)
SSLEngine_ECDSA_FIPSMODE (Failed)
```
They all passed on `F35`.

To reproduce simply do `./build.sh --with-tests` on `F36`, the same will succeed on F35.

An example failure is [here](https://github.com/dogtagpki/jss/runs/6554625304?check_suite_focus=true).

Stacktrace (it is the same for all failures):
```
73/77 Test #73: SSLEngine_ECDSA_FIPSMODE ..........................***Failed    0.21 sec
Initializing CryptoManager...
May 23, 2022 11:55:45 AM org.mozilla.jss.CryptoManager initialize
INFO: JSS CryptoManager: successfully initialized from NSS database at /github/home/build/jss/results/fips
Exception in thread "main" java.security.NoSuchProviderException: no such provider: SunJSSE
	at java.base/sun.security.jca.GetInstance.getService(GetInstance.java:83)
	at java.base/sun.security.jca.GetInstance.getInstance(GetInstance.java:206)
	at java.base/javax.net.ssl.SSLContext.getInstance(SSLContext.java:230)
	at org.mozilla.jss.tests.TestSSLEngine.sizeBuffers(TestSSLEngine.java:210)
	at org.mozilla.jss.tests.TestSSLEngine.initialize(TestSSLEngine.java:47)
	at org.mozilla.jss.tests.TestSSLEngine.main(TestSSLEngine.java:1014)
```
A quick read of the [source](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/javax/net/ssl/SSLContext.java#L190-L233) suggests that, for whatever reason, the `SunJSSE` provider is no longer being added to the list of registered providers whereas it was previously.

edewata:
It's reported on Debian too:
https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1855971.html

ckelleyRH:
In #864 I discovered that the issue only arises with OpenJDK 17. By migrating the Azure pipelines to use Java 17 I made them susceptible to this bug.

PR #864, opened 2022-06-10
Tidy up X509CRLImpl

ckelleyRH:
* Reorder modifiers to match the JLS
* Rename static constant to match JLS
* Remove commented out code
* Use pattern matching with instanceof
* Remove unnecessary else clauses
* Use ternary operator where appropriate
* Put array designator on the type not the variable
* Remove unnecessary negation in logic of parse() method

### DIFF
--- a/.github/workflows/maven-tests.yml
+++ b/.github/workflows/maven-tests.yml
@@ -6,6 +6,8 @@ jobs:
   build:
     name: Building JSS
     runs-on: ubuntu-latest
+    env:
+      JAVA_HOME: /usr/lib/jvm/java-17-openjdk-amd64
 
     steps:
     - name: Clone repository
@@ -14,7 +16,7 @@ jobs:
     - name: Set up Java
       uses: actions/setup-java@v2
       with:
-        java-version: '11'
+        java-version: '17'
         distribution: 'adopt'
 
     - name: Build JSS
--- a/.github/workflows/build-tests.yml
+++ b/.github/workflows/build-tests.yml
@@ -6,6 +6,8 @@ jobs:
   build-test:
     name: Build Test
     runs-on: ubuntu-latest
+    env:
+      JAVA_HOME: /usr/lib/jvm/java-17-openjdk-amd64
     strategy:
       matrix:
         os:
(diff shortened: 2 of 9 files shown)


fmarco76 (approved):
LGTM.

BTW,  CMake seems to build with java 17 sources but maven build with java 11. They should use the same version

ckelleyRH:
> BTW, CMake seems to build with java 17 sources but maven build with java 11. They should use the same version
Yeah I thought I had updated this when we moved on to OpenJDK 17 but clearly not, I must be thinking of another repo or mis-remembering.

ckelleyRH:
Now it works locally but doesn't work in the CI for some reason. Will look again tomorrow.

.github/workflows/maven-tests.yml
```diff
@@ -14,9 +14,12 @@ jobs:
     - name: Set up Java
       uses: actions/setup-java@v2
       with:
-        java-version: '11'
+        java-version: '17'
         distribution: 'adopt'
 
+    - name: Set JAVA_HOME
+      run: export JAVA_HOME=/usr/lib/jvm/jre-17-openjdk
```
edewata, line 21: I tried installing `openjdk-17-jdk` on a Debian container, the files got installed under `/usr/lib/jvm/java-17-openjdk-amd64` so maybe try that instead.

edewata: Can we actually set an environment variable this way in GH workflow? Or do we need to use `env:`?

fmarco76: The SonarCloud now is working but the env with the correct jvm should go in all tests with ubuntu/debian.

ckelleyRH:
@edewata @fmarco76 - now it fails because of #861. So I think we're "good".

edewata (approved):
It's reported on Debian too:
https://www.mail-archive.com/debian-bugs-dist@lists.debian.org/msg1855971.html
Regardless, the PR looks good. Thanks for the update!

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #866, opened 2022-06-23
Sonarcloud token

fmarco76:
Sonarcloud plug-in for maven allows to publish analyses without token if the project accept anonymous users. This allows to perform analysis on pull request from fork repository but have a side effect. Actions run in the fork can update analysis on Sonarcloud. As an example, the list of branch in Sonarcloud projects includes some branches from fork repository.

User could update also the analysis on master branch working in the fork.

Configuration on Sonarcloud is updated to avoid this behaviour and the change in this PR will work allows to work with PR from fork repositories.

### DIFF
--- a/.github/workflows/code-analysis-pull.yml
+++ b/.github/workflows/code-analysis-pull.yml
@@ -1,0 +1,102 @@
+name: Code Analysis for Pull Fork
+on:
+  workflow_run:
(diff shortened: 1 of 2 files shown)


edewata (approved):
Someday I need to learn how this works, but it LGTM :)

PR #868, opened 2022-07-01
Update Maven project

edewata:
The `pom.xml` files have been updated as follows:

- The JSS version has been updated to match RPM spec version
- The dependency versions have been updated to match the packages available on Fedora

The Maven build test in `maven-tests.yml` has been merged into `build-tests.yml`. It has also been updated to compare CMake build artifacts and Maven build artifacts. The comparison excludes `META-INF/maven/` since that folder is only available in Maven build artifacts.


fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM.

PR #869, opened 2022-07-06
Build native library in Maven

edewata:
This is a rebase of @cipherboy's PR #636 with some changes:
- the folder for Java code was changed to `base`
- the group ID was changed to `org.dogtagpki`
- the examples were added as a submodule

At least for now the native library can be built in Maven. Later the CI should be extended to test this library.


ckelleyRH (approved):
LGTM, although there is a lot of a Maven compilation stuff happening I have never seen before. If it works for you then OK :-)

fmarco76 (commented):
The overall structure looks good but I saw that the sonarcloud goal does not work so I have tested this repository locally but it got errror.

I tried running the command `mvn package` I exit with the following error:


```[INFO] /bin/sh -c cd '/home/mfargetta/Projects/tmp/jss/native' && 'gcc' '-shared' '-Wl,-z,defs' '-Wl,-soname' '-Wl,libjss.so' '-I/usr/include/nspr' '-I/usr/include/nspr4' '-I/usr/include/nss' '-I/usr/include/nss3' '-fPIC' '-I/usr/lib/jvm/java-17-openjdk/include' '-I/usr/lib/jvm/java-17-openjdk/include/linux' '-I/home/mfargetta/Projects/tmp/jss/native/src/main/native/org/mozilla/jss' '-I/home/mfargetta/Projects/tmp/jss/native/src/main/native/org/mozilla/jss/crypto' '-I/home/mfargetta/Projects/tmp/jss/native/src/main/native/org/mozilla/jss/nss' '-I/home/mfargetta/Projects/tmp/jss/native/src/main/native/org/mozilla/jss/pkcs11' '-I/home/mfargetta/Projects/tmp/jss/native/src/main/native/org/mozilla/jss/pkcs11/attrs' '-I/home/mfargetta/Projects/tmp/jss/native/src/main/native/org/mozilla/jss/ssl' '-I/home/mfargetta/Projects/tmp/jss/native/src/main/native/org/mozilla/jss/ssl/javax' '-I/home/mfargetta/Projects/tmp/jss/native/src/main/native/org/mozilla/jss/util' '-I/home/mfargetta/Projects/tmp/jss/native/../base/target/include' '-I/home/mfargetta/Projects/tmp/jss/native/src/main/native' '-o/home/mfargetta/Projects/tmp/jss/native/target/objs/CryptoManager.o' '-c' '/home/mfargetta/Projects/tmp/jss/native/src/main/native/org/mozilla/jss/CryptoManager.c'
/home/mfargetta/Projects/tmp/jss/native/src/main/native/org/mozilla/jss/CryptoManager.c:22:10: fatal error: jssver.h: No such file or directory
   22 | #include <jssver.h>
      |          ^~~~~~~~~~
compilation terminated.
```

Actually, the fils does not exist in the repository but there is the 
```
$ find . -name jssver.h*
./native/src/main/native/org/mozilla/jss/util/jssver.h.in
```
It works only if executed after the `build.sh`, is this intentional? You have to run both or something is missed in the `pom.xml` for  native code?

edewata:
Yeah, for now `build.sh` has to be run before `mvn package` because those header files are generated by CMake. In the future we might be able to update `build.sh` to use Maven instead of CMake, but ultimately we want to be able to run `mvn package` without `build.sh`.

cipherboy:
Just for some context dump :-)

Java has had versioning of (Crypto) Providers for a while so that (and our CryptoManager) are the two main places where the version is used within the code base.

Previously (back 4.4.x and earlier?) this was a manually maintained header file that needed to be updated on every release manually. CMake provided a templating functionality to let us pull it (ultimately) from the spec I believe (plus in-tree defaults for consumers from git/source tarball).

If there's an easy way to do that in Maven, all for it. :-) Perhaps it could be converted from a build-time process to an automated (fail the build if the version mismatch or overwrite with a newer version) checked-in variant? 

My 2c.

fmarco76 (approved):
@edewata and @cipherboy  thanks for the clarification. We can merge the code for me.
I will update the sonarcloud check so it will build correctly and then we will implements these last steps with maven.

fmarco76 (commented):
The version other values repeated in all the pom files should be properties so it will be easier to manage. Look at comments on https://github.com/dogtagpki/pki/pull/4050

ckelleyRH (approved):
LGTM

edewata:
@cipherboy @ckelleyRH @fmarco76 Woohoo! Thanks guys! \o/

PR #871, opened 2022-07-11
Fix issues in symkey library

edewata:
The `JSS_PK11_wrapSymKey()` in `SessionKey.cpp` has been modified to pass the symmetric key's nickname into `PK11SymKey`'s constructor just like the `JSS_PK11_wrapSymKey()` in `PK11SymKey.c`. Right now these methods belong to different libraries (`libjss.so` and `libjss-symkey.so`), but in the future they should be merged.

The `SessionKey.ListSymmetricKeys()` has been modified to use the internal token by default if the `tokenName` is `NULL`.

This is required by PKI PR [#4053](https://github.com/dogtagpki/pki/pull/4053).

### DIFF
--- a/symkey/src/main/java/org/mozilla/jss/symkey/SessionKey.cpp
+++ b/symkey/src/main/java/org/mozilla/jss/symkey/SessionKey.cpp
@@ -42,9 +42,12 @@ extern "C"
 #include <cstdarg>
 #include <string>
 
+#define PLAIN_CONSTRUCTOR "<init>"
+
 // DRM_PROTO begins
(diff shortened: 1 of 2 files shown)


ckelleyRH (commented):
Seems fine but my C++ is too limited to provide approval here.

fmarco76 (approved):
LGTM

The code is very C like, although these classes are compiled with a C++ compiler. Maybe the code could be cleaned using more C++ features but not in this PR for sure.

fmarco76:
I was wondering why some classes were in C and other in C++. Thanks @edewata for the clarification.

PR #872, opened 2022-07-13
Fix the compilation error on Sonarclod with maven

fmarco76:
After the PR #869 maven build does not work if the CMake build is not
execute before. This is due to properly generate the header for the native
code.

Since Sonarcloud is configured with maven build and it does not check c
code for the moment, the native module has been disabled for the code
analysis.


fmarco76:
To work properly again this has to be merged in main branch first because the action for sonarcloud has to be there to work,

edewata (approved):
LGTM

PR #873, opened 2022-07-13
Add doctype and missing html tags

fmarco76:
Missing elements identified by Sonarcloud in html files.

Actually, these files are manipulated by javadoc so it should not be needed!


ckelleyRH (approved):
LGTM!

PR #874, opened 2022-07-13
Sonarclout project name changes

fmarco76:
According to sonarcloud documentation the project id should not go
anymore in the pom.xml but it is derived or can be provided in the
command line.

The definition in the pom.xml generate an error


fmarco76:
@ckelleyRH and @edewata my previous PR does not fix all problem so I did this one and a direct commit (it is not the correct approach but was only a fix for this PR).
Now it works. Thanks!

PR #875, opened 2022-07-13
Fix promotion bugs

fmarco76:
Promotion bugs appen when a signed byte is converted to a signed int if
the upper bits are not masked.

See
https://wiki.sei.cmu.edu/confluence/display/java/NUM52-J.+Be+aware+of+numeric+promotion+behavior

and
http://findbugs.sourceforge.net/bugDescriptions.html#BIT_IOR_OF_SIGNED_BYTE

### DIFF
--- a/base/src/main/java/org/mozilla/jss/asn1/BIT_STRING.java
+++ b/base/src/main/java/org/mozilla/jss/asn1/BIT_STRING.java
@@ -122,7 +122,7 @@ public class BIT_STRING implements ASN1Value {
         BitSet bs = new BitSet();
         int numBits = (bits.length * 8) - padCount;
         for (int i = 0; i < numBits; i++) {
-            if ((bits[i / 8] & (0x80 >>> (i % 8))) != 0) {
+            if ((bits[i / 8] & (0x80 >>> (i % 8))) != 0) { //NOSONAR -  int promotion bug is prevented by the applied mask
                 bs.set(i);
             } else {
                 bs.clear(i);
@@ -143,7 +143,7 @@ public class BIT_STRING implements ASN1Value {
         // all elements are set to false by default
 
         for (int i = 0; i < array.length; i++) {
-            if ((bits[i / 8] & (0x80 >>> (i % 8))) != 0) {
+            if ((bits[i / 8] & (0x80 >>> (i % 8))) != 0) { //NOSONAR -  int promotion bug is prevented by the applied mask
                 array[i] = true;
             }
         }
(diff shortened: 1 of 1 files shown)


base/src/main/java/org/mozilla/jss/asn1/BIT_STRING.java
```diff
@@ -115,14 +115,14 @@ public byte[] getBits() {
...
-     * 
+     *
      * @return BIT STRING as BitSet.
      */
     public BitSet toBitSet() {
         BitSet bs = new BitSet();
         int numBits = (bits.length * 8) - padCount;
         for (int i = 0; i < numBits; i++) {
-            if ((bits[i / 8] & (0x80 >>> (i % 8))) != 0) {
+            if (((bits[i / 8] | 0xff) & (0x80 >>> (i % 8))) != 0) {
```
edewata, line 125: Hmm.. shouldn't we use `& 0xff` instead of `| 0xff` to mask off the bits? Also, IIUC the code is already masking off the bits using `& (0x80 ...)` so the additional mask might not be necessary.

fmarco76: @edewata yes, you are right about the mask, my mistake. It has to be with `&`.

I think the code analysis highlight the problem when you do bitwise operations between bytes and integers. In this particular case we have an additional mask so the promotion should not be a problem. If you and @ckelleyRH  think it is better to maintain the code we can disable this bug on Sonarcloud. Alternatively, I could try to modify the code in order to make the second mask identifiable by the code analysis but I am not sure about it.

edewata: Is it possible to disable SonarCloud for a specific line like `# pylint: disable=...`? I'm not sure it's worth tweaking the code just to prevent false positives in SonarCloud.

ckelleyRH: > Is it possible to disable SonarCloud for a specific line

[Yes](https://github.com/dogtagpki/tomcatjss/blob/master/core/src/main/java/org/apache/tomcat/util/net/jss/TomcatJSS.java#L519)! Just add `//NOSONAR` to a line to disable it, and I agree we should.

fmarco76: OK. I have reverted the code and disabled the checks in the lines.

The web interface allows to set a bug as a false positive but it is not clear what happen when the code get  updated and I did not find documentation.

edewata (approved):
LGTM. Thanks for the update!

PR #876, opened 2022-07-15
Fix UTF-8 conversion

fmarco76:
An if condition was never accessed because equal of a previous one.


ckelleyRH (approved):
LGTM!

One small improvement would be to correct the comment on line 139, as it is missing a zero.

I thought that there must be a library for this conversion somewhere, but can only find UCS-2 versions.

PR #877, opened 2022-07-21
Run shellcheck in JSS CI

ckelleyRH:
Currently the job fails because of a few warnings, but should be quick to fix in a separate PR.


fmarco76 (approved):
LGTM

ckelleyRH:
Thanks @fmarco76 !

PR #880, opened 2022-07-25
Make the base branch generic

fmarco76:
In case of commits to different branches the rebase has to be performed against the base branch of the PR. However, the code analysis execute in a `workflow_run` event and it has no information of the PR base, as in the [docuemtnation](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-example-51).

The available options are:
- to store the base branch during the original PR event and retrieve this information during the workflow run
- to retrieve the base branch using some javascript code interacting with GitHab rest API.

The former is followed here.
Of course, in order to work Sonarcloud in all the branch these have to include the relative workflows but this is required in any case to perform the analysis of them.

### DIFF
--- a/.github/workflows/code-analysis-pull.yml
+++ b/.github/workflows/code-analysis-pull.yml
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       pr-number: ${{ steps.pr-artifact-script.outputs.result }}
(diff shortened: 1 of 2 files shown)


edewata (approved):
There are issues with NSS 3.81 (https://github.com/dogtagpki/jss/issues/882) but this PR LGTM.

ckelleyRH (approved):
LGTM!

PR #881, opened 2022-07-25
Fix test failure introduced by efforts to fix shellcheck warnings

ckelleyRH:
A partial reversion which preserves the test while still allowing `shellcheck` to not issue warnings.


edewata (approved):
Looks like there are issues with NSS 3.81 (#882) but this PR LGTM. Thanks!

ckelleyRH:
Thanks @edewata !

PR #885, opened 2022-07-27
Revert workaround for NSS 3.81 due to COPR issues

edewata:
The workaround has been reverted since it was causing build issues in COPR:

```
Error:
 Problem 1: cannot install both nss-tools-3.81.0-1.fc36.x86_64 and nss-tools-3.77.0-1.fc36.x86_64
  - cannot install the best candidate for the job
  - conflicting requests
 Problem 2: cannot install both nss-devel-3.81.0-1.fc36.x86_64 and nss-devel-3.77.0-1.fc36.x86_64
  - cannot install the best candidate for the job
```

### DIFF
--- a/jss.spec
+++ b/jss.spec
@@ -80,12 +80,8 @@ BuildRequires:  zip
 BuildRequires:  unzip
 
 BuildRequires:  gcc-c++
-
-# TODO: Remove version upper bound once the issues with NSS 3.81 are resolved:
-# https://github.com/dogtagpki/jss/issues/882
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
Reverting the requirement it is possible to build JSS and deploy with PKI so at least we can test everything.

ckelleyRH (approved):

edewata:
@fmarco76 @ckelleyRH Thanks!

As expected the build tests are failing again. Just a note, Ubuntu/Debian is still using NSS 3.68/3.79 so they are not affected by this issue yet.

PR #886, opened 2022-07-28
Replace exclusion of i686 with new %{java_arches} macro

ckelleyRH:
This way we only build on archs that Java builds on.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2104064


edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #887, opened 2022-07-29
Fix synchronized overload

fmarco76:
Some extended class overload synchronized methods with not synchronized methods. This in theory could generate error although it is not the case as [discussed here](https://wiki.sei.cmu.edu/confluence/display/java/TSM00-J.+Do+not+override+thread-safe+methods+with+methods+that+are+not+thread-safe).

Actually, some of these methods overload methods of the class `Vector` and it is not clear to me if the synchronized is really required. If it is required then we have to add, otherwise it could be better to replace with  `ArrayList` as suggested in the java documentation.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/asn1/CountingStream.java
+++ b/base/src/main/java/org/mozilla/jss/asn1/CountingStream.java
@@ -31,7 +31,7 @@ class CountingStream extends InputStream {
     }
 
     @Override
-    public void mark(int readlimit) {
+    public synchronized void mark(int readlimit) {
         source.mark(readlimit);
         markpos = count;
         if (DEBUG) {
@@ -81,7 +81,7 @@ class CountingStream extends InputStream {
     }
 
     @Override
-    public void reset() throws IOException {
+    public synchronized void reset() throws IOException {
         source.reset();
         if (DEBUG) {
             System.out.println("reset from " + count + " to " + markpos);
--- a/base/src/main/java/org/mozilla/jss/ssl/SSLServerSocket.java
+++ b/base/src/main/java/org/mozilla/jss/ssl/SSLServerSocket.java
@@ -188,7 +188,7 @@ public class SSLServerSocket extends java.net.ServerSocket {
      * @param timeout The timeout time in milliseconds.
      */
     @Override
-    public void setSoTimeout(int timeout) {
+    public synchronized void setSoTimeout(int timeout) {
         base.setTimeout(timeout);
     }
 
@@ -198,7 +198,7 @@ public class SSLServerSocket extends java.net.ServerSocket {
      * @return The timeout time in milliseconds.
      */
     @Override
-    public int getSoTimeout() {
+    public synchronized int getSoTimeout() {
         return base.getTimeout();
     }
 


base/src/main/java/org/mozilla/jss/ssl/SSLSocket.java
```diff
@@ -761,7 +761,7 @@ public synchronized int getSoTimeout() throws SocketException {
      * Closes this socket.
      */
     @Override
-    public void close() throws IOException {
+    public synchronized void close() throws IOException {
         synchronized (this) {
```
ckelleyRH, line 765: I am sceptical of the thread safety paradigm used in this class, and I think there is deadlock potential.

We have a `readLock` object and a `writeLock` object, as well as obtaining a lock on the instance itself and the `close` method code itself is locked. In the `read` method, we obtain the lock on `readLock`, then we immediately take the lock on the instance, so what is the point of `readLock`? If we have the instance lock then no reading or writing can happen, so the granularity is unnecessary?

The `close` method takes the instance lock, then does "stuff", then potentially takes the `readLock` then the `writeLock`. The `write` and `read` methods do it the other way around, they take the `read/writeLock`, THEN they try to take the instance lock. If a `read` call came in just after a `close` call, then `close` holds the instance lock, waiting for `readLock`, and `read` has `readLock`, waiting for the instance lock. Deadlock.

What do you think @edewata?

ckelleyRH (approved):
I think the change itself LGTM, but I have a concern more generally about the class.

base/src/main/java/org/mozilla/jss/ssl/SSLSocket.java
fmarco76: @ckelleyRH I have review again all the changes and I agree in this case to synchronize the method without change the logic could create a deadlock problem. I would remove the `synchronized` from the method and add the exception in sonarcloud for the moment. I'll keep only for the others if you agree.

edewata: I agree with the possibility of deadlock here. In general I'd prefer we use only `synchronized` methods instead of `synchronized(object)` because the former is easier to understand. Maybe we should convert all `synchronized(this)` into new `synchronized` methods?

For the other classes, does it mean we need to synchronize all `hashCode()` and `equals()`? I think in general we assume that the classes are not thread-safe by default, so usually we only add `synchronized` if it's necessary.

edewata:
I think `Vector` is used in many places because the code was written before Java Collections existed, but we should replaced it if it can be done safely. I'd rather we do that instead of adding more `synchronized` keywords.

fmarco76:
> we should replaced it if it can be done safely

I'll try this change to see if there are not supported usage.

fmarco76:
I have updated the code to replace the `Vector` with `ArrayList` and revert the synchronization for the `SSLSocket`.

The code should be OK because I have replaced the `Vector` methods with other equivalents and the code build properly but the GitHub tests failk for the nss version problem.

ckelleyRH (changes requested):

base/src/main/java/org/mozilla/jss/netscape/security/x509/Extensions.java
```diff
@@ -40,7 +40,7 @@
  * @version 1.11
  * @see CertAttrSet
  */
-public class Extensions extends Vector<Extension>
+public class Extensions extends ArrayList<Extension>
```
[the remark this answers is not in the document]
edewata: I'd suggest we complete the `synchronized` changes in this PR first, then do the `Vector` replacement in a separate PR because that's possibly going to be much more complicated.

I'm actually hoping we could replace `Vector` variables/fields with `ArrayList` first. Replacing `Vector` in class inheritance can be done too, but as @ckelleyRH pointed above, it's likely to break backward compatibility. If we're OK with that, I think we should actually drop it from the class inheritance and make it a field instead, e.g.:

```
public class Extensions {
    ArrayList<Extension> extensions = new ArrayList<>();
}
```

This way other applications will be isolated from future changes in the internal JSS implementation. This is something that we can discuss in the new PR.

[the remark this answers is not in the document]
fmarco76: OK, I will create a new commit with only the few `synchronized` methods which make sense to fix now and move the discussion to new PRs which I would consider for all the collections.

fmarco76:
I am merging this PR with the synchronized method in the stream and socket where there were no comments or doubts.
The vector synchronization and the other more problematic synchronization will be managed in separate PR.

Thanks @edewata and @ckelleyRH for the comments.

Issue #796, opened 2021-08-04
Unable to validate PKCS #12 files generated by pk12util


edewata:
JSS can no longer validate PKCS #12 files generated by `pk12util` since recently `pk12util` was changed to use SHA-256 hash algorithm by default whereas JSS is still hard-coded to use SHA-1.

Steps to reproduce:
1. Install NSS 3.65 or later
2. Install PKI 10.11 or later
3. Create an NSS database:
```
$ pki client-init --force
```
4. Generate a CSR:
```
$ pki nss-cert-request \
    --subject "CN=Certificate Authority" \
    --ext /usr/share/pki/server/certs/ca_signing.conf \
    --csr ca_signing.csr
```
5. Issue a certificate:
```
$ pki nss-cert-issue \
    --csr ca_signing.csr \
    --ext /usr/share/pki/server/certs/ca_signing.conf \
    --cert ca_signing.crt
```
6. Import the certificate:
```
$ pki nss-cert-import --cert ca_signing.crt --trust CT,C,C ca_signing
```
7. Export the certificate and key into a PKCS12 file:
```
$ pk12util -d ~/.dogtag/nssdb -o test.p12 -W Secret.123 -n ca_signing
```
8. Inspect the content of PKCS12 file:
```
$ pki --debug pkcs12-cert-find --pkcs12-file test.p12 --pkcs12-password Secret.123
...
java.lang.Exception: Unable to validate PKCS #12 file: Digests do not match
	at org.mozilla.jss.netscape.security.pkcs.PKCS12Util.loadFromByteArray(PKCS12Util.java:839)
	at org.mozilla.jss.netscape.security.pkcs.PKCS12Util.loadFromFile(PKCS12Util.java:825)
	at com.netscape.cmstools.pkcs12.PKCS12CertFindCLI.execute(PKCS12CertFindCLI.java:128)
	at org.dogtagpki.cli.CommandCLI.execute(CommandCLI.java:58)
	at org.dogtagpki.cli.CLI.execute(CLI.java:357)
	at org.dogtagpki.cli.CLI.execute(CLI.java:357)
	at org.dogtagpki.cli.CLI.execute(CLI.java:357)
	at com.netscape.cmstools.cli.MainCLI.execute(MainCLI.java:656)
	at com.netscape.cmstools.cli.MainCLI.main(MainCLI.java:694)
...
```

[...]

PR #888, opened 2022-08-03
Fix import .p12 files with sha256 hash

fmarco76:
Pkcs12 files generated by nss and openssl cannot be imported by JSS bnecause the sha-256 signature is not currently supported in the digest verification. This problem is reported in the issue #796.

This PR will integrate the support to validate p12 file using sha256.


fmarco76:
Currently the sha256 digesst is validated but there are problem to read the keys.

fmarco76:
This PR has been merged with the PR #889 and the work will continue on that.

PR #889, opened 2022-08-05
Add ASN1 decoding of encrypted data using PBES2 algorithm

fmarco76:
The validation of pk12 file with certificate in `:pkcs7-encryptedData` fails because the ASN1 decode work only for PBES1. 

See https://bugzilla.redhat.com/show_bug.cgi?id=2115765.

With this PR the decode does not work but there are problem with the key generation but can be managed in a separate PR/task.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/crypto/HMACAlgorithm.java
+++ b/base/src/main/java/org/mozilla/jss/crypto/HMACAlgorithm.java
@@ -55,19 +55,19 @@ public class HMACAlgorithm extends DigestAlgorithm {
      */
     @Deprecated(since="5.0.1", forRemoval=true)
     public static final HMACAlgorithm SHA1 = new HMACAlgorithm
-        (CKM_SHA_1_HMAC, "SHA-1-HMAC",
-             OBJECT_IDENTIFIER.ALGORITHM.subBranch(26), 20);
+        (SEC_OID_HMAC_SHA1, "SHA-1-HMAC",
+                OBJECT_IDENTIFIER.RSA_DIGEST.subBranch(7), 20);
 
     public static final HMACAlgorithm SHA256 = new HMACAlgorithm
-        (CKM_SHA256_HMAC, "SHA-256-HMAC",
+        (SEC_OID_HMAC_SHA256, "SHA-256-HMAC",
              OBJECT_IDENTIFIER.RSA_DIGEST.subBranch(9), 32);
 
     public static final HMACAlgorithm SHA384 = new HMACAlgorithm
-        (CKM_SHA384_HMAC, "SHA-384-HMAC",
+        (SEC_OID_HMAC_SHA384, "SHA-384-HMAC",
              OBJECT_IDENTIFIER.RSA_DIGEST.subBranch(10), 48);
 
     public static final HMACAlgorithm SHA512 = new HMACAlgorithm
-        (CKM_SHA512_HMAC, "SHA-512-HMAC",
+        (SEC_OID_HMAC_SHA512, "SHA-512-HMAC",
              OBJECT_IDENTIFIER.RSA_DIGEST.subBranch(11), 64);
 
 }
--- a/base/src/main/java/org/mozilla/jss/crypto/PBEKeyGenParams.java
+++ b/base/src/main/java/org/mozilla/jss/crypto/PBEKeyGenParams.java
@@ -15,6 +15,7 @@ public class PBEKeyGenParams implements AlgorithmParameterSpec, KeySpec {
     private byte[] salt;
     private int iterations;
     private EncryptionAlgorithm encryptionAlgorithm = EncryptionAlgorithm.DES3_CBC;
+    private HMACAlgorithm hashAlgorithm = null;
 
     static private final int DEFAULT_SALT_LENGTH = 8;
     static private final int DEFAULT_ITERATIONS = 1;
@@ -52,12 +53,8 @@ public class PBEKeyGenParams implements AlgorithmParameterSpec, KeySpec {
      * @param iterations The iteration count for the PBE algorithm.
      */
     public PBEKeyGenParams(char[] pass, byte[] salt, int iterations) {
-        if (pass == null || salt == null) {
-            throw new NullPointerException();
-        }
-        this.pass = new Password(pass.clone());
-        this.salt = salt;
-        this.iterations = iterations;
+        this(pass, salt, iterations, null, null);
+
     }
 
     /**
@@ -77,6 +74,28 @@ public class PBEKeyGenParams implements AlgorithmParameterSpec, KeySpec {
     public PBEKeyGenParams(
             char[] pass, byte[] salt, int iterations,
             EncryptionAlgorithm encAlg) {
+        this(pass, salt, iterations, encAlg, null);
+    }
+
+    /**
+     * Creates PBE parameters using default encryption algorithm
+     * (DES3_EDE3_CBC).
+     *
+     * @param pass The password. It will be cloned, so the
+     *            caller is still responsible for clearing it. It must not be null.
+     * @param salt The salt for the PBE algorithm. Will <b>not</b> be cloned.
+     *            Must not be null. It is the responsibility of the caller to
+     *            use the right salt length for the algorithm. Most algorithms
+     *            use 8 bytes of salt.
+     * @param iterations The iteration count for the PBE algorithm.
+     * @param encAlg The encryption algorithm. This is used with SOME
+     *            PBE algorithms for determining the KDF output length.
+     * @param hashAlg The hash algorithm. This is used with PBEv2 algorithms
+     * because it cannot be derived from the key generation algorithm.
+     */
+    public PBEKeyGenParams(
+            char[] pass, byte[] salt, int iterations,
+            EncryptionAlgorithm encAlg, HMACAlgorithm hashAlg) {
         if (pass == null || salt == null) {
             throw new NullPointerException();
         }
@@ -85,8 +104,8 @@ public class PBEKeyGenParams implements AlgorithmParameterSpec, KeySpec {
         this.iterations = iterations;
         if (encAlg != null)
             this.encryptionAlgorithm = encAlg;
+        this.hashAlgorithm = hashAlg;
     }
-
     /**
      * Returns a <b>reference</b> to the password, not a copy.
      */
@@ -114,6 +133,14 @@ public class PBEKeyGenParams implements AlgorithmParameterSpec, KeySpec {
      */
     public EncryptionAlgorithm getEncryptionAlgorithm() {
         return encryptionAlgorithm;
+    }
+
+    /**
+     * The hash algorithm is used with PBEv2 algorithms because it cannot be
+     * derived from the key generation algorithm.
+     */
+    public HMACAlgorithm getHashAlgorithm() {
+        return hashAlgorithm;
     }
 
     /**
--- a/base/src/main/java/org/mozilla/jss/pkcs12/MacData.java
+++ b/base/src/main/java/org/mozilla/jss/pkcs12/MacData.java
@@ -83,22 +83,48 @@ public class MacData implements ASN1Value {
 
     /**
      * Creates a MacData by computing a HMAC on the given bytes. An HMAC
-	 *	is a message authentication code, which is a keyed digest. It proves
-	 *	not only that data has not been tampered with, but also that the
-	 *	entity that created the HMAC possessed the symmetric key.
-	 *
-	 * @param password The password used to generate a key using a
-	 *		PBE mechanism.
-	 * @param macSalt The salt used as input to the PBE key generation
-	 *		mechanism. If null is passed in, new random salt will be created.
+     * is a message authentication code, which is a keyed digest. It proves
+     * not only that data has not been tampered with, but also that the
+     * entity that created the HMAC possessed the symmetric key.
+     *
+     * @param password The password used to generate a key using a PBE mechanism.
+     * @param macSalt The salt used as input to the PBE key generation mechanism.
+     *      If null is passed in, new random salt will be created.
      * @param iterations The iteration count for creating the PBE key.
-	 * @param toBeMACed The data on which the HMAC will be computed.
+     * @param toBeMACed The data on which the HMAC will be computed.
      * @exception NotInitializedException If the crypto subsystem
      *      has not been initialized yet.
      * @exception TokenException If an error occurs on a crypto token.
-	 */
+     */
     public MacData( Password password, byte[] macSalt,
                     int iterations, byte[] toBeMACed )
+        throws NotInitializedException,
+            DigestException, TokenException, CharConversionException
+    {
+        this(password, macSalt, iterations, toBeMACed, null);
+    }
+
+
+    /**
+     * Creates a MacData by computing a HMAC on the given bytes. An HMAC
+     * is a message authentication code, which is a keyed digest. It proves
+     * not only that data has not been tampered with, but also that the
+     * entity that created the HMAC possessed the symmetric key.
+     *
+     * @param password The password used to generate a key using a PBE mechanism.
+     * @param macSalt The salt used as input to the PBE key generation mechanism.
+     *      If null is passed in, new random salt will be created.
+     * @param iterations The iteration count for creating the PBE key.
+     * @param toBeMACed The data on which the HMAC will be computed.
+     * @param algID The algorithm used to compute the HMAC, If null the
+     *      SHA1 will be used.
+     * @exception NotInitializedException If the crypto subsystem
+     *      has not been initialized yet.
+     * @exception TokenException If an error occurs on a crypto token.
+     */
+    public MacData( Password password, byte[] macSalt,
+                    int iterations, byte[] toBeMACed,
+                    AlgorithmIdentifier algID)
         throws NotInitializedException,
             DigestException, TokenException, CharConversionException
     {
@@ -115,24 +141,45 @@ public class MacData implements ASN1Value {
 
         try {
             // generate key from password and salt
-            KeyGenerator kg = token.getKeyGenerator(KeyGenAlgorithm.PBA_SHA1_HMAC);
+            if(algID == null) {
+                algID = new AlgorithmIdentifier(DigestAlgorithm.SHA1.toOID());
+            }
+            KeyGenerator kg = null;
+            JSSMessageDigest digest = null;
+            if(DigestAlgorithm.SHA1.toOID().equals(algID.getOID())){
+                kg = token.getKeyGenerator(KeyGenAlgorithm.PBA_SHA1_HMAC);
+                digest = token.getDigestContext(HMACAlgorithm.SHA1);
+            }
+            if(DigestAlgorithm.SHA256.toOID().equals(algID.getOID())){
+                kg = token.getKeyGenerator(KeyGenAlgorithm.PBE_SHA256_HMAC);
+                digest = token.getDigestContext(HMACAlgorithm.SHA256);
+            }
+            if(DigestAlgorithm.SHA384.toOID().equals(algID.getOID())){
+                kg = token.getKeyGenerator(KeyGenAlgorithm.PBE_SHA384_HMAC);
+                digest = token.getDigestContext(HMACAlgorithm.SHA384);
+            }
+            if(DigestAlgorithm.SHA512.toOID().equals(algID.getOID())){
+                kg = token.getKeyGenerator(KeyGenAlgorithm.PBE_SHA512_HMAC);
+                digest = token.getDigestContext(HMACAlgorithm.SHA512);
+            }
+            if(kg == null || digest == null) {
+                throw new NoSuchAlgorithmException("Algorithm (oid:" + algID.getOID().toDottedString() + ") not managed for digest");
+            }
             kg.setCharToByteConverter(new PasswordConverter());
             kg.initialize(params);
             SymmetricKey key = kg.generate();
 
             // perform the digesting
-            JSSMessageDigest digest = token.getDigestContext(HMACAlgorithm.SHA1);
             digest.initHMAC(key);
             byte[] digestBytes = digest.digest(toBeMACed);
 
             // put everything into a DigestInfo
-            AlgorithmIdentifier algID = new AlgorithmIdentifier(DigestAlgorithm.SHA1.toOID());
             this.mac = new DigestInfo(algID, new OCTET_STRING(digestBytes));
             this.macSalt = new OCTET_STRING(macSalt);
             this.macIterationCount = new INTEGER(iterations);
 
         } catch (NoSuchAlgorithmException e) {
-            throw new RuntimeException("SHA-1 HMAC algorithm not found on internal " +
+            throw new RuntimeException("HMAC algorithm not found on internal " +
                     "token: " + e.getMessage(), e);
 
         } catch (InvalidAlgorithmParameterException e) {
--- a/base/src/main/java/org/mozilla/jss/pkcs7/EncryptedContentInfo.java
+++ b/base/src/main/java/org/mozilla/jss/pkcs7/EncryptedContentInfo.java
@@ -30,6 +30,7 @@ import org.mozilla.jss.asn1.Tag;
 import org.mozilla.jss.crypto.Cipher;
 import org.mozilla.jss.crypto.CryptoToken;
 import org.mozilla.jss.crypto.EncryptionAlgorithm;
+import org.mozilla.jss.crypto.HMACAlgorithm;
 import org.mozilla.jss.crypto.IVParameterSpec;
 import org.mozilla.jss.crypto.IllegalBlockSizeException;
 import org.mozilla.jss.crypto.KeyGenAlgorithm;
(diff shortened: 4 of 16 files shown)


edewata:
I will need to defer to @jmagne on this one.

edewata:
SonarCloud complains about duplicated code blocks. Is there a way to make it less sensitive? If not we probably should just disable it because the complaints don't really make sense to me. @ckelleyRH What do you think?

ckelleyRH:
> SonarCloud complains about duplicated code blocks. Is there a way to make it less sensitive? If not we probably should just disable it because the complaints don't really make sense to me. @ckelleyRH What do you think?

It is complaining that there are multiple identical methods, e.g.:
```
private static final Tag TAG = SEQUENCE.TAG;
@Override
public Tag getTag() {
    return TAG;
 }
  
@Override
public void encode(OutputStream ostream) throws IOException {
    sequence.encode(ostream);
}
```
...is duplicated across many classes - implying that a common base class to implement this method may be useful. Or a default method in the interface. Anyway, I think this is actually useful as it is telling us that we can reduce boilerplate code by tweaking the classes/interfaces.

edewata:
Sometimes the classes are not inheriting from the same base class, and sometimes the code depends on class attributes (i.e. not static), so it could be difficult to consolidate the duplicate code without a major redesign. This could be a major barrier if we require that to be fixed before merging the PR.

ckelleyRH:
Difficult, and potentially dangerous, as something could go wrong easily in the refactor.

Currently the duplication metric doesn't affect the gate, so it is an advisory. I guess it has had the desired effect, we have discussed it, but there is no need to fix it for a PR as it won't make the CI fail.

edewata:
So code smells are just advisory? I thought I saw SonarCloud did not pass due to code smells (maybe there were too many).

ckelleyRH:
> So code smells are just advisory? I thought I saw SonarCloud did not pass due to code smells (maybe there were too many).

Supposedly, they could make the "maintainability" gate fail. I conclude this metric must be a simple weight of numbers, as there are supposed "blocker" and "critical" severity issues listed there and we have an A rating. This contrasts to the "reliability" metric, which does seem to be tied to the severity of the issues more.

fmarco76:
This PR incluide the #888. They fix two different problem so we keep them separated but both need to be merged in order to manage p12 files with SHA256 with jss

base/src/main/java/org/mozilla/jss/crypto/PBEKeyGenParams.java
```diff
@@ -31,7 +32,7 @@ public class PBEKeyGenParams implements AlgorithmParameterSpec, KeySpec {
      * @param iterations The iteration count for the PBE algorithm.
      */
     public PBEKeyGenParams(Password pass, byte[] salt, int iterations) {
-        if (pass == null || salt == null) {
+        if (pass == null) {
```
jmagne, line 35: So are we now accounting for a case where the salt is null here?

base/src/main/java/org/mozilla/jss/pkcs12/MacData.java
```diff
@@ -115,18 +141,31 @@ public MacData( Password password, byte[] macSalt,
...
+            if(algID == null) {
+                algID = new AlgorithmIdentifier(DigestAlgorithm.SHA1.toOID());
+            }
+            KeyGenerator kg = null;
+            JSSMessageDigest digest = null;
+            if(DigestAlgorithm.SHA1.toOID().equals(algID.getOID())){
+                kg = token.getKeyGenerator(KeyGenAlgorithm.PBA_SHA1_HMAC);
+                digest = token.getDigestContext(HMACAlgorithm.SHA1);
+            }
+            if(DigestAlgorithm.SHA256.toOID().equals(algID.getOID())){
```
jmagne, line 153: Are we going to want to add one for 384 as well?

base/src/main/java/org/mozilla/jss/pkcs12/MacData.java
```diff
@@ -115,18 +141,31 @@ public MacData( Password password, byte[] macSalt,
...
+                kg = token.getKeyGenerator(KeyGenAlgorithm.PBE_SHA256_HMAC);
+                digest = token.getDigestContext(HMACAlgorithm.SHA256);
+            }
+            if(kg == null) {
```
jmagne, line 157: Is it possible for the digest to be null as well?

jmagne (commented):
This is just to show I got started looking at this. Will continue.. thx.

base/src/main/java/org/mozilla/jss/pkcs12/MacData.java
fmarco76: I am not sure about the adoption but the plan should be to add all the option to 512 in future update.

base/src/main/java/org/mozilla/jss/pkcs12/MacData.java
fmarco76: Actually, both `kg` and `digest` should be null or both should have a value because if one of them is not possible than an exception is raised and this code is not reached. So for the check one of them  should be enough but  I am going to include both in the `if` for more safety.

base/src/main/java/org/mozilla/jss/crypto/PBEKeyGenParams.java
fmarco76: Sorry, this is a mistake. I am going to fix it.

native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyGenerator.c
```diff
@@ -282,53 +283,61 @@ Java_org_mozilla_jss_pkcs11_PK11KeyGenerator_generatePBE(
...
-            goto finish;
-        }
-
-        /* generate the key */
-        skey = PK11_PBEKeyGen(slot, algid, pwitem, PR_FALSE /*faulty3DES*/,
-                        NULL /*wincx*/);
-        if( skey == NULL ) {
-            JSS_throwMsg(env, TOKEN_EXCEPTION, "Failed to generate PBE key");
-            goto finish;
+        break;
```
jmagne, line 294: OK, after looking at the original file, I think I see what this is doing. Previously, the code was just an (if) for SHA1 amd (else) for everything else, which results in using a PBEV2 algid. You have just made this a switch where you've added a case for SEC_OID_SHA256, which takes it's own action.

My question here, just for my own understanding, is what case are we satisfying here? Is this something you ran into when experimenting with various p12 files and found out you needed this? Also should there be 384 and 512 variants.   Trying to refresh my memory, I think the case we were trying to debug was ending up in the (default) clause. I'm just trying to get an understanding where the added case comes into play and if you saw this code being hit? Thanks.

native/src/main/native/org/mozilla/jss/crypto/Algorithm.c
```diff
@@ -124,7 +124,7 @@ JSS_AlgInfo JSS_AlgTable[NUM_ALGS] = {
 /* 35 */    {CKM_AES_CBC_PAD, PK11_MECH},
 /* 36 */    {CKM_RC2_CBC_PAD, PK11_MECH},
 /* 37 */    {CKM_RC2_KEY_GEN, PK11_MECH},
-/* 38 */    {SEC_OID_SHA256, SEC_OID_TAG},
+/* 38 */    {SEC_OID_SHA256, PK11_MECH},
 /* 39 */    {SEC_OID_SHA384, SEC_OID_TAG},
 /* 40 */    {SEC_OID_SHA512, SEC_OID_TAG},
 /* 41 */    {SEC_OID_PKCS1_SHA256_WITH_RSA_ENCRYPTION, SEC_OID_TAG},
```
jmagne, line 130: I take it we will need entries here for 384 and 512?

jmagne (commented):
Just added a few more comments / questions. I guess we need to decide if for this round 256 is good enough or we try to get 384 and 512 working. Next round I should approve once we've decided.

fmarco76:
> I guess we need to decide if for this round 256 is good enough or we try to get 384 and 512 working

The original ticket generating this work is #796 where only the case for 256 is mentioned. I started working for that with the idea  to verify the other cases (>256) after this. We can try to add SHA-384 and SHA-512 in this PR if we agree since it should be straight forward (to be verified). I will do some tests.

base/src/main/java/org/mozilla/jss/pkcs7/EncryptedContentInfo.java
```diff
@@ -236,28 +238,59 @@ public static EncryptedContentInfo createCRSCompatibleEncryptedContentInfo(OBJEC
...
+            }
+            salt = pbeParams.getSalt();
+            iterations = pbeParams.getIterations();
+            encAlg = ((PBEAlgorithm)kgAlg).getEncryptionAlg();
+        }
+        else {
             byte[] encodedParams = ASN1Util.encode(params);
-            pbeParams = (PBEParameter)
-                ASN1Util.decode( PBEParameter.getTemplate(), encodedParams );
+            PBES2Params pbe2Params = (PBES2Params)
```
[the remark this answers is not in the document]
fmarco76: According to the specification `params` should be present and if not we cannot go ahead. If they are not present the pk12 is not valid. We could have a custom exception for p12 not valid in this case but I am not sure if it is something to consider at this level.

base/src/main/java/org/mozilla/jss/pkix/primitive/PBES2Params.java
```diff
@@ -0,0 +1,114 @@
...
+    private SEQUENCE sequence;
+
+    public AlgorithmIdentifier getKeyDerivationFunc() {
+        return keyDerivationFunc;
+    }
+
+    public AlgorithmIdentifier getEncryptionScheme() {
+        return encryptionScheme;
+    }
+
```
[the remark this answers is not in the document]
fmarco76: Yes, this is common with other `ASN1Value` implementation. All the values are setted by the constructor.

native/src/main/native/org/mozilla/jss/crypto/Algorithm.c
fmarco76: Yes

base/src/main/java/org/mozilla/jss/pkcs7/EncryptedContentInfo.java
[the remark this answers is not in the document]
fmarco76: Testing with OpenSSL p12 files there could be problem. I am investigating them.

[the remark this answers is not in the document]
fmarco76: The problem is only with `KeyLength` which is optional for the specification and `OpenSSL` was not adding. The overall `params` cannot be null because the `encode` or `decode` method would throw an exception.

native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyGenerator.c
fmarco76: Finally, I have reverted the code to use the `if`. However, I have add a new `if`  for  SHA-[256, 384, 512]. In case the symmetric key is not generated a last attempt is done with the default case. To note that the `skey` is null only if the other `if` are not used because if they are and the key is `NULL` the execution go the `finish` label at the end of the method.

Additionally, I have add a new method for the new case to avoid code repetition.

fmarco76:
The new version also works with SHA-384 and SHA-512. As an example, this is with a pkcs12 generated by OpenSSL:


```
$ /usr/lib/jvm/java-17-openjdk-17.0.4.1.1-1.fc36.x86_64/bin/java -cp "/usr/share/pki/lib/*" -Dcom.redhat.fips=false -Djava.util.logging.config.file=/usr/share/pki/etc/logging.properties com.netscape.cmstools.cli.MainCLI pkcs12-cert-find --pkcs12-file test-open-sha384.p12 --pkcs12-password Secret.123
WARNING: The SHA-1 algorithm used in org.mozilla.jss.pkcs12.SafeBag::getLocalKeyIDFromCert:264 is deprecated. Use a more secure algorithm.
---------------
1 entries found
---------------
  Certificate ID: 0xf8361681d6c99ac7b3cc01345bc7b6eb6acd224a
  Serial Number: 0xa0c0761e2dd098b4e030b3f7ecd15c10
  Friendly Name: Certificate Authority
  Subject DN: CN=Certificate Authority
  Issuer DN: CN=Certificate Authority
  Has Key: true
  Key ID: 0xf8361681d6c99ac7b3cc01345bc7b6eb6acd224a

$ openssl pkcs12 -in test-open-sha384.p12 -noout -info
Enter Import Password:
MAC: sha384, Iteration 2048
MAC length: 48, salt length: 8
PKCS7 Encrypted data: PBES2, PBKDF2, AES-256-CBC, Iteration 2048, PRF hmacWithSHA256
Certificate bag
PKCS7 Data
Shrouded Keybag: PBES2, PBKDF2, AES-256-CBC, Iteration 2048, PRF hmacWithSHA256
```

base/src/main/java/org/mozilla/jss/crypto/PBEKeyGenParams.java
jmagne: Looks good.

base/src/main/java/org/mozilla/jss/pkcs12/MacData.java
jmagne: Makes sense.

base/src/main/java/org/mozilla/jss/pkcs12/MacData.java
jmagne: Looks good.

base/src/main/java/org/mozilla/jss/pkcs7/EncryptedContentInfo.java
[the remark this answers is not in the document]
jmagne: OK.

base/src/main/java/org/mozilla/jss/pkix/primitive/PBES2Params.java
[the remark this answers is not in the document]
jmagne: Good.

jmagne (commented):
OK, I've gotten back to this one. Resolved a few conversations. I still need to get back to this next week and look over some of the final more complex changes such as the pbe key gneration changes. Thx.

base/src/main/java/org/mozilla/jss/crypto/HMACAlgorithm.java
```diff
@@ -58,6 +58,14 @@ public static HMACAlgorithm fromOID(OBJECT_IDENTIFIER oid)
         (CKM_SHA_1_HMAC, "SHA-1-HMAC",
              OBJECT_IDENTIFIER.ALGORITHM.subBranch(26), 20);
 
+    public static final HMACAlgorithm hSHA1 = new HMACAlgorithm
```
jmagne, line 61: My apologies if this has been discussed or asked about before.... I see that we've created some new types with names such as hSHA1 and sSHA256 to kind of add to the ones that were there before.  Question is, what is the purpose of these, since I've searched through the PR and not found any references. Perhaps they are referenced in a companion PR.

base/src/main/java/org/mozilla/jss/crypto/HMACAlgorithm.java
```diff
@@ -58,6 +58,14 @@ public static HMACAlgorithm fromOID(OBJECT_IDENTIFIER oid)
...
+            (SEC_OID_HMAC_SHA1, "SHA-1-HMAC",
```
jmagne, line 62: Apologies again since I didn't notice this the first trip through. Since the PR has been adding some new algorithms such as these PBE HMAC algorithms, I failed to mention that we usually put versions of these JSS algorithms in the cyrpto provider class we support to allow the various algorithms to be accessed through the cyrpto provider interface. The main file that establishes these interfaces is base/src/main/java/org/mozilla/jss/JSSProvider.java.  The following is an example I did for the PSS SHA256 signature algorithm within JSSProvider.java:

 put("Signature.RSASSA-PSS",
            "org.mozilla.jss.provider.java.security.JSSSignatureSpi$RSAPSSSignature");

        put("Alg.Alias.Signature.1.2.840.113549.1.1.10",     "RSASSA-PSS");
        put("Alg.Alias.Signature.OID.1.2.840.113549.1.1.10", "RSASSA-PSS");

        put("Signature.SHA-256/RSA/PSS",
            "org.mozilla.jss.provider.java.security.JSSSignatureSpi$SHA256RSAPSS");

        put("Alg.Alias.Signature.SHA256withRSA/PSS","SHA-256/RSA/PSS");

Note that the alias items refer to different names that can be used to call up the various alg with the cyrpto provider interface.

Examples of tests for both styles can be found in:   tests/SigTest.java and tests/JSASigTest.java. I mention the test files because this shows examples on how the JCA interface is used and also to state that we usually put some simple tests when we add new algs such as this, and the tests will run during the build time tests that usually take place. If you need further info on this, just ask, and that will be my opportunity to refresh my own memory of this since it's been awhile since I did the PSS feature.

native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyGenerator.c
```diff
@@ -183,6 +183,73 @@ print_secitem(SECItem *item) {
...
+    if( salt == NULL ) {
+        JSS_throwMsg(env, TOKEN_EXCEPTION,
+            "constructSHAPBEKey:"
+            " salt NULL");
+        goto finish;
+    }
+
+    params = PK11_CreatePBEParams(salt, pwitem,
+        iterationCount);
+    switch(mech){
```
jmagne, line 222: Looks like maybe we have some tab / space formatting issues here with respect to the switch clause and the lines after. Minor if you get around to it.

native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyGenerator.c
```diff
@@ -282,53 +349,64 @@ Java_org_mozilla_jss_pkcs11_PK11KeyGenerator_generatePBE(
...
+			JSS_throwMsg(env, TOKEN_EXCEPTION, "Failed to generate PBE key");
+			goto finish;
+		}
     }
 
     /* wrap the key. This sets skey to NULL. */
-    keyObj = JSS_PK11_wrapSymKey(env, &skey);
+    if( skey!=NULL ) {
+        keyObj = JSS_PK11_wrapSymKey(env, &skey);
+    }
```
jmagne, line 409: Also one more quick thing I'm just putting at the end for convenience. Since we've done some significant work in the native code, I usually take it upon myself to run some memory analyzer such as Valgrind on a few of the major manual test cases to see if any memory leaks have escaped the review. Often it's easy to miss simple instances of failing to free memory or resources. If you want to use valgrind, just google valgrind quick start and just run your major pk12 test files through and see if it reports anything inside any of the code that was added. This is of course if you haven't done this sort of thing already.

jmagne (commented):
Left a few more comments but it continues to look better and better, thx.

base/src/main/java/org/mozilla/jss/crypto/HMACAlgorithm.java
fmarco76: These are used in this PR. I need to register these algorithm because they are used to decrypt the PKCS Authenticated Safe. As an example is I remove them I get:

```
java.security.NoSuchAlgorithmException
	at org.mozilla.jss.crypto.HMACAlgorithm.fromOID(HMACAlgorithm.java:45)
	at org.mozilla.jss.pkcs7.EncryptedContentInfo.decrypt(EncryptedContentInfo.java:286)
	at org.mozilla.jss.pkcs12.AuthenticatedSafes.getSafeContentsAt(AuthenticatedSafes.java:169)
...
```

This is for a SHA1 based pkcs12 like:

```
$ openssl pkcs12 -in test-nss-sha1.p12 -info -noout
Enter Import Password:
MAC: sha1, Iteration 600000
MAC length: 20, salt length: 16
PKCS7 Data
Shrouded Keybag: PBES2, PBKDF2, AES-256-CBC, Iteration 600000, PRF hmacWithSHA1
PKCS7 Encrypted data: PBES2, PBKDF2, AES-128-CBC, Iteration 600000, PRF hmacWithSHA1
Certificate bag
```

If I tried with a pkcs12 generated by openssl with SHA-384 like:

``
$ openssl pkcs12 -in test-open-sha384.p12 -info -noout
Enter Import Password:
MAC: sha384, Iteration 2048
MAC length: 48, salt length: 8
PKCS7 Encrypted data: PBES2, PBKDF2, AES-256-CBC, Iteration 2048, PRF hmacWithSHA256
Certificate bag
PKCS7 Data
Shrouded Keybag: PBES2, PBKDF2, AES-256-CBC, Iteration 2048, PRF hmacWithSHA256
```

The error is:

```
org.mozilla.jss.crypto.TokenException: Cipher context finalization failed: (-8190) security library: received bad data.
	at org.mozilla.jss.pkcs11.PK11Cipher.finalizeContext(Native Method)
	at org.mozilla.jss.pkcs11.PK11Cipher.doFinal(PK11Cipher.java:186)
	at org.mozilla.jss.pkcs7.EncryptedContentInfo.decrypt(EncryptedContentInfo.java:325)
	at org.mozilla.jss.pkcs12.AuthenticatedSafes.getSafeContentsAt(AuthenticatedSafes.java:169)
```
In this second case the HMAC is found but it is not correct.

File: native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyGenerator.c (hunk 10 above)
Comment (fmarco76) [reply native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyGenerator.c:222]: I cannot find a good way to correctly format java and C in eclipse. Fixed using vim.

File: native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyGenerator.c (hunk 11 above)
Comment (fmarco76) [reply native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyGenerator.c:409]: This is my first valgrind test. I have build the code with debug opetion and run a test with valgrind.
I am attaching the output for the memory check. It seems there are no memory leak with the c code modified here but let me know if I run it in a wrong way.

[valgrind.txt](https://github.com/dogtagpki/jss/files/10086815/valgrind.txt)

Comment (jmagne) [reply native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyGenerator.c:409]: Looks like you did it correctly. For convenience, instead of using > you can specify everything to go to a log file with this invocation:  valgrind --log-file="filename"

I didn't find any reports in the native files you modified. Anyway maybe try one more run with the --log-file switch to make sure we are seeing everything possible. If you see pretty much the same output, then we can call it good.

Comment (fmarco76) [reply native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyGenerator.c:409]: This is definitely better because I can see the output of the program much easier. No problem with the modified file. This is the log file:
[valgrind.log](https://github.com/dogtagpki/jss/files/10122641/valgrind.log)

File: base/src/main/java/org/mozilla/jss/crypto/HMACAlgorithm.java [hunk 12]
```diff
@@ -56,18 +56,18 @@ public static HMACAlgorithm fromOID(OBJECT_IDENTIFIER oid)
     @Deprecated(since="5.0.1", forRemoval=true)
     public static final HMACAlgorithm SHA1 = new HMACAlgorithm
         (CKM_SHA_1_HMAC, "SHA-1-HMAC",
-             OBJECT_IDENTIFIER.ALGORITHM.subBranch(26), 20);
```
jmagne, line 59: OK, this looks interesting. In the chat you asked about the SEC_OIDS vs CKMs. My thing about converting to CKMs was only a question at the time, not a concrete suggestion :) After looking at the nss code, it seems like these algs are are declared in seoid.c something like this.:

OD(hmac_sha1, SEC_OID_HMAC_SHA1, "HMAC SHA-1",
       CKM_SHA_1_HMAC, INVALID_CERT_EXTENSION)

and

OD(hmac_sha256, SEC_OID_HMAC_SHA256, "HMAC SHA-256",
       CKM_SHA256_HMAC, INVALID_CERT_EXTENSION)

This leads me to the conclusion that nss must have a way to convert between the two when someone is trying to use say a SEC_OID. The CKMs are mechanisms that the pkcs#11 modules need to do the work. Therefore I think your idea to go all SEC_OIDs seems quite the way to go. Interesting that the SHA1 version we are still keeping as a CKM. Since the 3.67 nss code I was looking at shows that there is a corresponding SECOID for that one as well, SEC_OID_HMAC_SHA1, I wonder if it would make sense to just try that out to see if all is kosher. Maybe there has been some historical reason why this was done.

jmagne: I just wanted to get this straight since the hsm's use the pkcs#11 interface. If nss is correctly converting these for us it should be all good there, that is if the hsm actually supports said algs.

fmarco76: SEC_OID_HMAC_SHA1 is not defined I was adding it but reverting the code after the integration with JSSProvider I have removed it. I can try to reintroduce so we will use only SEC_OID and let nss to convert to the  correct pk11 mechanism.

fmarco76: > his leads me to the conclusion that nss must have a way to convert between the two when someone is trying to use say a SEC_OID

I think this should invoke the conversion: https://github.com/dogtagpki/jss/blob/master/native/src/main/native/org/mozilla/jss/crypto/Algorithm.c#L227-L240

jmagne (approved):
OK, I think we are at the point where we have to approve, which I'm doing. If any corner cases have been introduced hopefully and their test suite will uncover. Thanks for the assistance on this one.

fmarco76:
Last commit was just to squash all commits in just two and run the CI workflow which was broken during the previous commits.
No code changes

fmarco76:
@jmagne Thanks for the support and review!

PR #890, opened 2022-08-08
Tidyup SSLClient and SSLServer

fmarco76:
Clean some of the SonarCloud complaints on SSLClient and SSLServer classes.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/ssl/SSLClient.java
+++ b/base/src/main/java/org/mozilla/jss/ssl/SSLClient.java
@@ -13,7 +13,8 @@ import java.io.OutputStream;
 import java.io.PrintStream;
 import java.net.InetAddress;
 import java.net.Socket;
-import java.util.Hashtable;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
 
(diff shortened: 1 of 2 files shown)


ckelleyRH (approved):
LGTM!

edewata (approved):
LGTM.

SonarCloud complains about Cognitive Complexity. It might be useful for new code, but we have a lot of complex legacy code. Should we just disable this warning? Regardless, feel free to merge.

ckelleyRH:
> LGTM.
> 
> SonarCloud complains about Cognitive Complexity. It might be useful for new code, but we have a lot of complex legacy code. Should we just disable this warning? Regardless, feel free to merge.

I think it's a good idea, there are probably a lot of demons hiding in these very long, old methods.

fmarco76:
I think it is good to remove the checks if you think the warning is not true. There are many cases where this happen and I will disable(e.g. we have warning asking to use try with resources in factory methods where the resources is the return value)

In this specific  case it is a good suggestion to simplify the method so we can leave in case we should change the method in the future.

fmarco76:
Thanks @ckelleyRH and @edewata

PR #891, opened 2022-08-08
Consolidate ContentInfo classes

ckelleyRH:
There were two ContentInfo classes, which differed only by a single
whitespace character!

One copy was raised up the hierarchy to a more common location and the
other was deleted.

There is a 3rd `ContentInfo` class, which looks very similar but doesn't have an identical implementation. I left that one alone, for now at least!


ckelleyRH:
Might be some more work to do, as `ContentInfo` uses some classes from the package(s) it was in originally, so this might need more thought (or abandoning).

fmarco76:
The majority of classes in the package cms are similar to the classes in pkcs7, not only `ContentInfo`. I have not done a detailed comparison but it seems they could be reorganised and merged.

ckelleyRH:
> The majority of classes in the package cms are similar to the classes in pkcs7, not only `ContentInfo`. I have not done a detailed comparison but it seems they could be reorganised and merged.

Yeah this was supposed to be step 1 along the path of doing as you describe! This will require corresponding changes in PKI as moving the classes between packages breaks the PKI build.

fmarco76:
It is OK to move the class one-by-one and verify the compatibility but I was thinking that if we plan to have only a copy of each than for me make sense to have in pkcs7 package because they represent pkcs7 data, but I do not know what it is `pkix/cms`.

edewata:
One possible solution is to make one of them the base class, then deprecate the others and make them inherit from the base class.

ckelleyRH:
I won't get to finish this, so closing and leaving here for future reference.

PR #892, opened 2022-08-09
Tidyup ExtensionsRequested

fmarco76:


edewata (approved):
LGTM. I have some comments below but feel free to merge.

Currently the `DerInputStream.getDerValue()` will never return a `null` value, so the loop will only terminate if there's an exception. Since the old code also terminates using an exception, the behavior is unchanged, so this is fine. I think ideally the `DerInputStream.getDerValue()` should be modified to return a `null` if it reaches the end of the stream. That way we would be able to distinguish it more easily from real problems.

SonarCloud complains about empty catch block. Maybe we can put something like this?
```
} catch (IOException e) {
    // ignore
}
```
If we modify `DerInputStream.getDerValue()` to return `null` at the end of stream, we can actually log the error there and/or rethrow the exception so it can be handled properly:
```
} catch (IOException e) {
    logger.error("Unable to parse extension: " + e.getMessage(), e);
    throw e;
}
```

base/src/main/java/org/mozilla/jss/netscape/security/x509/ExtensionsRequested.java
```diff
@@ -160,17 +160,13 @@ private void construct(DerValue dv) throws IOException {
...
 
-        while (true) {
-            DerValue ext_dv = null;
-            try {
-                ext_dv = stream.getDerValue();
-            } catch (IOException ex) {
-                break;
+        DerValue extDv = null;
+        try {
+            while ((extDv = stream.getDerValue()) != null) {
```
ckelleyRH, line 165: Can we do `stream.available() > 0` instead, to check if there is anything in the buffer? This would allow you to drop the `try` with the empty `catch` (I'm not a fan of empty catches), as you break out of the loop before you get the `IOException`. I am sure I tried to do that once and @edewata suggested a reason why not to do it, but I can't remember for the life of me what the reason was. Can you please remind me ^_^

ckelleyRH: Ah, or maybe I tried to do the `null` check in the `while` loop... that sounds familiar. Go team concurrent review, like London buses you wait all day for a review and two come at once :business_suit_levitating: :busstop: :bus: :bus:

edewata: Sorry, I can't remember what we discussed... :)
I don't have any specific objection for using `stream.available()` here, but it might potentially change the behavior. If we're confident enough it won't cause any problems, then sure.

fmarco76: My original idea was to keep the exact same behaviour but I agree it was not the best.

Check the null value requires to modify multiple classes but the check with available  should work as expected. The check is done with 1 because a DER Value require at least 2 byte ([documentation](https://letsencrypt.org/docs/a-warm-welcome-to-asn1-and-der/)) and I have verified in the code. So if the DER value is not read it is for some exception that it is propagated.

ckelleyRH (approved):
LGTM, thanks for verifying the new check!

base/src/main/java/org/mozilla/jss/netscape/security/x509/ExtensionsRequested.java
```diff
@@ -160,18 +160,12 @@ private void construct(DerValue dv) throws IOException {
...
-                break;
-            }
-
-            Extension ext = new Extension(ext_dv);
+        DerValue extDv = null;
+        while (stream.available()>1) {
+            extDv = stream.getDerValue();
+            Extension ext = new Extension(extDv);
             exts.addElement(ext);
         }
```
edewata, line 168: Hmm.. maybe this is the issue that I discussed with @ckelleyRH before. According to `InputStream.available()` [Javadoc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/InputStream.html#available()):
> Returns an **estimate** of the number of bytes that can be read (or skipped over) from this input stream **without blocking**, which may be 0, or 0 when end of stream is detected.

To my understanding this means that `available() <= 1` doesn't imply that it has reached the end of the stream. It's possible that there will be more data coming later. So terminating the loop based on this condition might not be correct.

fmarco76: Yes, that is in the definition for the super class `InputStream`. However, in this case it is a ByteArrayInputStream which seems not have this behaviour, according to the [javadoc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/ByteArrayInputStream.html#available()). In this case the `read()` is not blocking and the `available()` should return the length of all you can read, if I get correctly.

@edewata if you think my interpretation could be not correct let me know and I will follow the other approach with the `null` condition.

edewata: Alternatively, we can move that logic into `getDerValue()` to hide the details from `DerInputStream` users. Something like this:
```
public DerValue getDerValue() throws IOException {

    // buffer.available() will return the remaining bytes in the stream precisely
    // (since it's a ByteArrayInputStream). DER values are at least 2 bytes long:
    // https://letsencrypt.org/docs/a-warm-welcome-to-asn1-and-der/

    return buffer.available() >= 2 ? new DerValue(buffer) : null;
}
```

fmarco76: Comment added in the while loop.
The method `getDerValue()` is called from 91 different places and in all these cases the logic is based on the exception. If we modify the method to return `null` instead of the exception we need to change all the calling code and it is a big effort. I think this effort is valuable if we plan to reorganise these classes to optimise/modernise the code  but this will be in a separate PR.

fmarco76:
Thanks @ckelleyRH and @edewata

PR #896, opened 2022-08-30
Use ExcludeArch on older platforms

edewata:
`java_arches` macro is only available on newer platforms, so the spec file has been modified to use `ExcludeArch` on older platforms.

This will fix COPR builds on older Fedora versions and RHEL.


fmarco76 (approved):
It is good and you can merge although I would prefer a conditional approach as suggested in dogtagpki/pki#4131

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #897, opened 2022-09-20
Allow other build tests to continue if one fails

ckelleyRH:
Currently if a build job fails on e.g. Fedora, the Ubuntu job will cancel. This makes it hard to see whether the other jobs are afflicted with a similar issue, so add continue-on-error so all jobs run to completion.


edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata !

PR #899, opened 2022-09-27
Replace deprecated getPackage() call with call to System classloader

ckelleyRH:


fmarco76 (approved):
LGTM, although it does not fix the nss problem I think we should merge.

edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata @fmarco76 !

PR #901, opened 2022-09-29
Remove unnecessary else clauses from various classes

ckelleyRH:
* Where appropriate check-and-cast and ternary operators were used
* Some logic was also simplified by removing Boolean literals

### DIFF
--- a/base/src/main/java/org/mozilla/jss/pkcs12/AuthenticatedSafes.java
+++ b/base/src/main/java/org/mozilla/jss/pkcs12/AuthenticatedSafes.java
@@ -174,20 +174,18 @@ public class AuthenticatedSafes implements ASN1Value {
                                                 SafeBag.getTemplate() );
                 return (SEQUENCE) ASN1Util.decode(seqt, decrypted);
             } catch(InvalidBERException e) {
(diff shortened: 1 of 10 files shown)


fmarco76 (approved):
LGTM

base/src/main/java/org/mozilla/jss/pkcs12/AuthenticatedSafes.java
```diff
@@ -185,9 +185,8 @@ public SEQUENCE getSafeContentsAt(Password password, int index)
                 SEQUENCE.OF_Template seqt = new SEQUENCE.OF_Template(
                                                     SafeBag.getTemplate() );
                 return (SEQUENCE) ASN1Util.decode(seqt, os.toByteArray());
-              } else {
-                throw e;
               }
+            throw e;
```
edewata, line 189: This code should be indented with 2 spaces to match the original code block. Ideally, the code block could be indented with 4 spaces, but it's up to you if you want to fix it or just leave it as is.

edewata (approved):
Just one minor comment, but otherwise it LGTM.

ckelleyRH:
Thanks @edewata @fmarco76 !

Issue #882, opened 2022-07-25
Incompatibilities with NSS 3.81


edewata:
Multiple tests have failed since NSS 3.81 was released in Fedora. [Downgrading to NSS 3.77](https://github.com/dogtagpki/jss/pull/883) seems to fix the problems.

https://github.com/dogtagpki/jss/runs/7506244588?check_suite_focus=true#step:7:2239
```
35/77 Test #35: SSLClientAuth .....................................***Failed    6.81 sec
Jul 25, 2022 6:27:06 PM org.mozilla.jss.CryptoManager initialize
INFO: JSS CryptoManager: successfully initialized from NSS database at /github/home/build/jss/results/nssdb
***FilePasswordCallback returns m1oZilla
using port:2876
Server about to create socket
Server created socket
Server specified cert directly
Server about to accept
client about to connect
Client specified cert directly
client connected
Server accepted
Server about to read
Error: Exception in thread "main" org.mozilla.jss.ssl.SSLSocketException: SSL_ForceHandshake failed: (-8010) Unknown error
	at org.mozilla.jss.ssl.SSLSocket.forceHandshake(Native Method)
	at org.mozilla.jss.tests.SSLClientAuth.testConnection(SSLClientAuth.java:372)
	at org.mozilla.jss.tests.SSLClientAuth.doIt(SSLClientAuth.java:183)
	at org.mozilla.jss.tests.SSLClientAuth.main(SSLClientAuth.java:117)
java.io.IOException: SocketException cannot read on socket: Error reading from socket: (-12271) SSL peer cannot verify your certificate.
	at org.mozilla.jss.ssl.SSLSocket.read(SSLSocket.java:1511)
	at org.mozilla.jss.ssl.SSLInputStream.read(SSLInputStream.java:43)
	at org.mozilla.jss.ssl.SSLInputStream.read(SSLInputStream.java:28)
	at org.mozilla.jss.tests.SSLClientAuth.run(SSLClientAuth.java:464)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: org.mozilla.jss.ssl.SSLSocketException: Error reading from socket: (-12271) SSL peer cannot verify your certificate.
	at org.mozilla.jss.ssl.SSLSocket.socketRead(Native Method)
	at org.mozilla.jss.ssl.SSLSocket.read(SSLSocket.java:1505)
	... 4 more
Server exiting
```
[...]

PR #902, opened 2022-10-01
Fix Azure pipelines

edewata:
Due to recent changes in Azure or Fedora it's no longer possible to run tests in Azure with the standard Fedora image.

To fix the problem, the `Dockerfile` has been updated to define a new Fedora/CentOS image that provides a `systemd` service. The test job in Azure has been split into two jobs: one for Fedora/CentOS and one for Debian/Ubuntu. The Fedora/CentOS job will build the image, create a container with the image, then run the tests with it, whereas the Debian/Ubuntu job is basically unchanged.

**Note:** The test is now running, but it's failing due to a issue https://github.com/dogtagpki/jss/issues/882:
https://dev.azure.com/dogtagpki/jss/_build/results?buildId=814&view=logs&jobId=185660ed-fcb9-5db1-85f2-144090af2c88

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,16 @@ ARG BASE_IMAGE="registry.fedoraproject.org/fedora:latest"
 ARG COPR_REPO="@pki/master"
 
 ################################################################################
-FROM $BASE_IMAGE AS jss-builder
+FROM $BASE_IMAGE AS jss-base
+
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
LGTM

ckelleyRH (approved):
LGTM!

PR #903, opened 2022-10-28
Introduce new scripts to automate the build process

ckelleyRH:
A new script update_version.sh is introduced, to start to reduce some of the manual burden of building packages. Usage is as:

./update_version.sh <major> <minor> <update> <phase>

...where phase is optional.

In this first iteration, it does the following:

* Updates the spec version to the new version provided
* Commits that change
* Creates a tag based on the new version provided
* Creates a source tarball based on the new version provided

A companion script is also included to revert the version update, if required.

### DIFF
--- a/revert_update_version.sh
+++ b/revert_update_version.sh
@@ -1,0 +1,19 @@
+#!/bin/bash -e
+
+# Use this script to revert the commit and delete the tag created using the update_version.sh script.
+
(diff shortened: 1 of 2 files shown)


edewata (approved):
LGTM

fmarco76 (approved):
LGTM

ckelleyRH:
@fmarco76 @edewata Thanks!

PR #906, opened 2022-11-11
Run RPMInspect in JSS CLI

ckelleyRH:


edewata (approved):
The `build-test` job was originally meant to be a generic build test (i.e. using `.jar` files instead of `.rpm`) so it doesn't use the `if` condition except for the platform-specific dependency installations. Since `rpminspect` is only available on Fedora it might make more sense to put the new test into a separate job, but either way is fine, I'll let you decide. Feel free to merge.

fmarco76 (approved):
LGTM

tests/bin/rpminspect.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
+
+# Don't run metadata check as we can't know the build host subdomain
+# of CI runners in advance to add to an allow list
+
+echo "Running RPMInspect on SRPM"
+rpminspect-fedora -E metadata /github/home/build/jss/SRPMS/*.rpm
```
fmarco76, line 7: After looking tomcatjss PR I agree this (and line 10) could be relative path for a more usability of the script

ckelleyRH:
> The `build-test` job was originally meant to be a generic build test (i.e. using `.jar` files instead of `.rpm`) so it doesn't use the `if` condition except for the platform-specific dependency installations. Since `rpminspect` is only available on Fedora it might make more sense to put the new test into a separate job, but either way is fine, I'll let you decide. Feel free to merge.

Yeah wasn't sure which way to go with this one. I tried to choose the path where I could re-use the build but I can't actually do that so a new test is probably better - I shall refactor and merge in, thanks!

tests/bin/rpminspect.sh
ckelleyRH: Ack, will update, thanks!

PR #907, opened 2022-12-02
Update workflows to use not deeprecated actions

fmarco76:
Some action are using [Node.js v12 and this has been deprecated by GitHub](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).

Additionally, the use of [`set-state` and `set-output` has been deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/).

Therefore the workflows are updated to the new version of the actions.

### DIFF
--- a/.github/workflows/build-tests.yml
+++ b/.github/workflows/build-tests.yml
@@ -19,7 +19,7 @@ jobs:
     container: ${{ matrix.os }}
     steps:
     - name: Clone repository
-      uses: actions/checkout@v2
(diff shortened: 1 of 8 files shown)


ckelleyRH (approved):
LGTM, hopefully we can figure out the CI issues soon!

fmarco76:
Thanks, @ckelleyRH

PR #908, opened 2022-12-02
Modify container cgroup

fmarco76:
Systemd has some problem to run in docker container if the host uses only cgroups2.

This could make the run container failing with the error:

Failed to create /init.scope control group: Read-only file system Failed to allocate manager object: Read-only file system [!!!!!!] Failed to allocate manager object.
Exiting PID 1...

A possible solution is fo use docker.slice as proposed here: https://serverfault.com/questions/1053187/systemd-fails-to-run-in-a-docker-container-when-using-cgroupv2-cgroupns-priva

### DIFF
--- a/tests/bin/runner-init.sh
+++ b/tests/bin/runner-init.sh
@@ -7,7 +7,6 @@ docker run \
     --privileged \
     --tmpfs /tmp \
     --tmpfs /run \
-    -v /sys/fs/cgroup:/sys/fs/cgroup:ro \
     -v ${GITHUB_WORKSPACE}:${SHARED} \
     -i \
     ${IMAGE}


edewata:
I'm not an expert in this area, but it looks like the containers run just fine in PKI with this script:
https://github.com/dogtagpki/pki/blob/master/tests/bin/runner-init.sh

Note that we don't touch cgroups at all there, so maybe we should try to do the same here?

fmarco76:
> we should try to do the same here?

@edewata yes, you are right. Removing the cgroups  should be the best approach. Thanks!

fmarco76:
@edewata Thanks!

PR #910, opened 2022-12-07
Fix pkcs7 serialisation when no signatures present

fmarco76:
According the documentation [1]:

  In the case of external signatures, the
   content being signed is absent from the EncapsulatedContentInfo value
   included

The current implementation always include the content but with length equela to zero. This is modified to be consistent with the specification.

1. https://www.rfc-editor.org/rfc/rfc5652#section-5.2


frasertweedale:
Tested, works as expected.  `content` tagged OCTET_STRING is now omitted when empty.

PR #911, opened 2022-12-07
[v4.9.x] Fix pkcs7 serialisation when no signatures present

fmarco76:
According the documentation [1]:

  In the case of external signatures, the
   content being signed is absent from the EncapsulatedContentInfo value
   included

The current implementation always include the content but with length equela to zero. This is modified to be consistent with the specification.

1. https://www.rfc-editor.org/rfc/rfc5652#section-5.2

### DIFF
--- a/src/main/java/org/mozilla/jss/netscape/security/pkcs/ContentInfo.java
+++ b/src/main/java/org/mozilla/jss/netscape/security/pkcs/ContentInfo.java
@@ -75,9 +75,11 @@ public class ContentInfo {
      * Make a contentInfo of type data.
      */
     public ContentInfo(byte[] bytes) {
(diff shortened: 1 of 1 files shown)


edewata:
Per discussion on gchat this change might be too risky to be done in the `v4.9.x` branch, and the CI is no longer working, and also there has not been any real problems caused by the existing code, so we probably should not change it. However, it should be fine to change the `master` branch since QE will run a full regression test in the future. I'd suggest closing this PR and keeping the #910 instead.

frasertweedale:
I agree with @edewata.  Although the behaviour is not conformant to RFC, there are no *known* issues caused by the the current behaviour.  In this case we should leave the releases alone to avoid unnecessary risk.

I merged #910 and I will close this PR.

PR #913, opened 2022-12-12
[v5.3] fix pkcs12

fmarco76:
Solve the validation error on pkcs12 files identified in the bug: 2017098


ckelleyRH (approved):
Looks fine to me and reasoning for backport seems sound but will leave it to @jmagne to offer his thoughts as the original reviewer.

jmagne (approved):
Looks exactly like the last one. Therefore if all tests etc look good approving.

PR #915, opened 2022-12-14
Consolidate build jobs

edewata:
The build jobs in test workflows have been consolidated into `build.yml` such that the build will be created just once by the build workflow, and the test workflows will use the same build once it's completed.

https://github.com/lewagon/wait-on-check-action


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #918, opened 2022-12-21
Drop default COPR repo

edewata:
The GitHub workflows have been modified to no longer use a COPR repo by default. Instead, it will install packages from GitHub Packages. The Azure Pipelines still have a dependency on COPR. It will be removed separately later.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #919, opened 2022-12-22
Add jss-dist image

edewata:
The CI has been modified to store the RPMs in an Alpine-based image and publish it to GH Packages to reduce the size of the distribution.


fmarco76 (approved):
LGTM

PR #920, opened 2022-12-22
Remove ot modify expression which always evaluate the same

fmarco76:
Following SonarCloud indications some expressions always verify the same so they have been removed of modified


edewata (approved):
LGTM

PR #922, opened 2022-12-23
Fix several issues identified by SonarCloud

fmarco76:
This commit will solve some of the most critical problem identified by SonarCloud. In more details they are:

- buried exceptions: some `finally` clause where thrown exception hiding the exceptions thrown in the try block. The code has been modified so the finally exception is thrown only if the try block has not exception, preserving the order of appearance;
- some conditions where always (or never) verified;
- possible NullPointerException in several points.

This whould also allow the code quality to go from "D" to "C" level.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/netscape/security/util/Cert.java
+++ b/base/src/main/java/org/mozilla/jss/netscape/security/util/Cert.java
@@ -126,7 +126,7 @@ public class Cert {
         mime64 = stripCertBrackets(mime64.trim());
         String newval = normalizeCertStr(mime64);
         // byte rawPub[] = mDecoder.decodeBuffer(newval);
-        byte rawPub[] = Utils.base64decode(newval);
+        byte[] rawPub = Utils.base64decode(newval);
         X509CertImpl cert = null;
 
         try {
@@ -141,7 +141,7 @@ public class Cert {
         mime64 = stripCertBrackets(mime64.trim());
         String newval = normalizeCertStr(mime64);
         // byte rawPub[] = mDecoder.decodeBuffer(newval);
-        byte rawPub[] = Utils.base64decode(newval);
+        byte[] rawPub = Utils.base64decode(newval);
         PKCS7 p7 = null;
 
         try {
@@ -157,7 +157,7 @@ public class Cert {
         mime64 = stripCRLBrackets(mime64.trim());
         String newval = normalizeCertStr(mime64);
         // byte rawPub[] = mDecoder.decodeBuffer(newval);
-        byte rawPub[] = Utils.base64decode(newval);
+        byte[] rawPub = Utils.base64decode(newval);
         X509CRL crl = null;
 
         try {
@@ -171,7 +171,7 @@ public class Cert {
             throws IOException {
         mime64 = stripCRLBrackets(mime64.trim());
 
-        byte rawPub[] = Utils.base64decode(mime64);
+        byte[] rawPub = Utils.base64decode(mime64);
         X509CRL crl = null;
 
         try {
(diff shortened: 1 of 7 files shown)


base/src/main/java/org/mozilla/jss/pkcs11/PK11KeyWrapper.java
```diff
@@ -291,29 +291,37 @@ private AlgorithmParameterSpec checkParams(AlgorithmParameterSpec params)
...
+        } catch (TokenException e) {
+            te = e;
+        }
+        if (parameters != null) {
+            try {
+                ((NativeEnclosure) parameters).close();
+            } catch (Exception e) {
+                if (te == null) {
+                    te =  new TokenException(e.getMessage(), e);
                 }
```
edewata, line 318: If `te` is not  `null`, the `e` will be buried instead, and we will never know if something went wrong in `NativeEnclosure.close()` which might cause a memory leak. I'd suggest to log an error message with the stack trace since the `NativeEnclosure.close()` is not supposed to fail.

base/src/main/java/org/mozilla/jss/pkcs11/PK11KeyWrapper.java
```diff
@@ -622,19 +630,28 @@ token, symKey, wrapped, algorithm, algFromType(type),
...
+        } catch (TokenException e) {
+            te = e;
+        }
+        if (parameters != null) {
+            try {
+                ((NativeEnclosure) parameters).close();
+            } catch (Exception e) {
+                if (te == null) {
+                    te = new TokenException(e.getMessage(), e);
                 }
```
edewata, line 648: Please see the above comment.

edewata (approved):
I have some comments, but the changes generally look good. Feel free to update as needed and merge. Thanks!

base/src/main/java/org/mozilla/jss/netscape/security/util/Cert.java
```diff
@@ -183,32 +183,28 @@ public static X509CRL mapCRL1(String mime64)
...
-            val.append(s.charAt(i));
         }
         return val.toString();
     }
 
     public static String normalizeCertStrAndReq(String s) {
-        StringBuffer val = new StringBuffer();
+        StringBuilder val = new StringBuilder();
 
         for (int i = 0; i < s.length(); i++) {
```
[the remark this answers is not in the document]
fmarco76: Fixed, The two methods are almost identical, the code could be improved but this is for a separate PR.

fmarco76:
There is a [CI error with RPMInspect](https://github.com/dogtagpki/jss/actions/runs/3874965528/jobs/6607446235#step:8:57) but it seems not related to this PR.

@ckelleyRH do you know if a a fix is needed or the problem is on fedora package?

ckelleyRH (approved):
LGTM

ckelleyRH:
> There is a [CI error with RPMInspect](https://github.com/dogtagpki/jss/actions/runs/3874965528/jobs/6607446235#step:8:57) but it seems not related to this PR.
> 
> @ckelleyRH do you know if a a fix is needed or the problem is on fedora package?



> There is a [CI error with RPMInspect](https://github.com/dogtagpki/jss/actions/runs/3874965528/jobs/6607446235#step:8:57) but it seems not related to this PR.
> 
> @ckelleyRH do you know if a a fix is needed or the problem is on fedora package?

I've not looked in detail but the same issue is reported in other CI runs so this is OK to merge. Recently I converted the licence fields of the packages we maintain in Fedora to SPDX (as that is the direction Fedora is going). It is likely that there is now some new `RPMInspect` check that validates that licence and we are failing it for some reason (the `RPMInspect` job only runs in the Fedora VM, with Fedora packages). I will raise an issue/ticket to investigate it.

fmarco76:
@ckelleyRH @edewata Thanks!

PR #923, opened 2023-01-03
Replace BASE64_MATRIX with BASE64_OS

edewata:
Previously the `BASE64_MATRIX` parameter provided a mechanism to test against multiple Fedora versions at once. However, since the test resources are limited and only one of the versions is eventually published, the parameter has been replaced with a new `BASE64_OS` parameter which only supports a single Fedora version.

https://github.com/dogtagpki/pki/wiki/Configuring-Test-OS

### DIFF
--- a/tests/bin/init-workflow.sh
+++ b/tests/bin/init-workflow.sh
@@ -1,14 +1,15 @@
 #!/bin/bash -e
 
-if [ "$BASE64_MATRIX" == "" ]
+if [ "$BASE64_OS" != "" ]
 then
-    MATRIX="{\"os\":[\"latest\"]}"
+    OS_VERSION=$(echo "$BASE64_OS" | base64 -d)
 else
-    MATRIX=$(echo "$BASE64_MATRIX" | base64 -d)
+    OS_VERSION=latest
 fi
 
-echo "MATRIX: $MATRIX"
-echo "matrix=$MATRIX" >> $GITHUB_OUTPUT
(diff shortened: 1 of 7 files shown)


tests/bin/init-workflow.sh
```diff
@@ -1,14 +1,20 @@
 #!/bin/bash -e
 
-if [ "$BASE64_MATRIX" == "" ]
+if [ "$BASE64_OS" != "" ]
 then
-    MATRIX="{\"os\":[\"latest\"]}"
+    OS_VERSION=$(echo "$BASE64_OS" | base64 -d)
+elif [ "$BASE64_MATRIX" != "" ]
```
ckelleyRH, line 6: Do we need this deprecated option still? Can we not simply replace the matrix with os?

edewata: Not really. I just thought it a common practice to deprecate first if possible before removing something. But since these params are meant to be temporary, it might be OK to just replace it. Should I go ahead and replace it?

edewata: OK, I went ahead and dropped `BASE64_MATRIX` completely.

ckelleyRH: Yeah it is good practice in general for sure. I think in this case though it is fine to drop it as it is only us working with this parameter and we are certain that we have accounted for every use of it.

ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH Thanks!

PR #925, opened 2023-01-05
git Make update_version script update jss_config_version

ckelleyRH:
In JSS there is an extra step required for updating where the value of jss_config_version must be bumped in sync with the version and phase. Currently this does not happen, which will lead to irregularities in testing and potentially incompatible config.


edewata (approved):
LGTM

PR #930, opened 2023-01-11
Fix potential problem identifid by Covscan

fmarco76:
For the error see:
https://cov01.lab.eng.brq2.redhat.com/covscanhub/task/269948//log/added.err


ckelleyRH (approved):
Thanks for investigating @fmarco76 !

fmarco76:
@ckelleyRH Thanks!

Issue #610, opened 2020-07-28
Unable to create SSL connection for PostgreSQL


edewata:
JSS fails to establish an SSL connection for PostgreSQL. This may be required by PKI ACME in certain environments. If the SSL mode is disabled, the PostgreSQL connection can be established properly without SSL.

Steps to reproduce:
1. Install PKI 10.9, JSS 4.7, and PostgreSQL JDBC 42.2.8.
2. Prepare a PostgreSQL database that supports SSL.
3. Install PKI ACME (https://github.com/dogtagpki/pki/blob/master/docs/installation/acme/Installing_ACME_Responder.md) with the following database.conf:
```
class=org.dogtagpki.acme.database.PostgreSQLDatabase
password=<password>
url=jdbc:postgresql://<server>:5432/<database>
user=<username>
```
4. Execute an operation that uses database connection, e.g.:
```
$ certbot certonly --standalone -d $HOSTNAME \
    --server http://$HOSTNAME:8080/acme/directory \
    --register-unsafely-without-email \
    --agree-tos
```
5. Disable SSL by adding the following parameter into database.conf:
```
sslmode=disable
```
6. Restart the server and rerun the same command.

Actual result: The operation in step 4 failed, but the same operation in step 6 succeeded.

Expected result: Both operations in step 4 and 6 should succeed.

Additional info: In step 4 the ACME debug log will have the following error messages:
```
2020-07-28 02:06:14 [http-nio-8080-exec-2] INFO: Connecting to jdbc:postgresql://<server>:5432/<database>
2020-07-28 02:06:15 [http-nio-8080-exec-2] SEVERE: Servlet.service() for servlet [ACME] in context with path [/acme] threw exception
org.jboss.resteasy.spi.UnhandledException: org.postgresql.util.PSQLException: SSL error: Unable to unwrap data using SSLEngine: Error duing SSL.ForceHandshake() :: SSL_ERROR_INAPPROPRIATE_FALLBACK_ALERT (-12157)
	at org.jboss.resteasy.core.ExceptionHandler.handleApplicationException(ExceptionHandler.java:78)
	at org.jboss.resteasy.core.ExceptionHandler.handleException(ExceptionHandler.java:222)
	at org.jboss.resteasy.core.SynchronousDispatcher.writeException(SynchronousDispatcher.java:179)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:422)
[...]

PR #931, opened 2023-01-12
Add test with postgresql

fmarco76:
Add a new workflow testing ACME with postgresql. The connection to the DB  uses SSL and it is managed through JSS so this verify it is working with a third party applications.

See issue #610


ckelleyRH (approved):
Seems to work, LGTM!

.github/workflows/external-application-connection-tests.yml
```diff
@@ -0,0 +1,203 @@
...
+          docker exec pki ln -s /usr/share/java/ongres-stringprep/stringprep.jar /usr/share/pki/server/common/lib/
+          docker exec pki pki-server restart --wait
+
+      - name: Install ACME in PKI container
+        run: |
+          docker exec pki pki-server acme-create
+          docker exec pki pki-server acme-database-mod \
+              --type postgresql \
+              -Dpassword=mysecretpassword \
+              -Durl=jdbc:postgresql://postgresql.exampla.com:5432/acme
```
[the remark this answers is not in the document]
edewata: I think we should add `ssl=true` and `sslmode=require` to the `url` to ensure that it's using an SSL connection. See also:
* https://jdbc.postgresql.org/documentation/use/
* https://jdbc.postgresql.org/documentation/ssl/

.github/workflows/external-application-connection-tests.yml
```diff
@@ -0,0 +1,203 @@
...
+          docker exec pki pki-server acme-issuer-mod --type pki
+          docker exec pki pki-server acme-realm-mod \
+              --type postgresql \
+              -Dpassword=mysecretpassword \
+              -Durl=jdbc:postgresql://postgresql.exampla.com:5432/acme
```
edewata, line 159: Same thing here about the `url`.

.github/workflows/external-application-connection-tests.yml
```diff
@@ -0,0 +1,203 @@
...
+          openssl pkcs12 -in sslserver.p12 -nocerts -out sslserver.key -noenc -password  pass:secret
+          openssl pkcs12 -in sslserver.p12 -nokeys -clcerts -out sslserver.crt  -password pass:secret
+          openssl pkcs12 -in sslserver.p12 -nokeys -cacerts -out ca.crt  -password pass:secret
+
+      - name: Build postgrsql image with certificates
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          tags: postgres-ssl
+          target: postgres-ssl
```
edewata, line 123: It looks like it's possible to deploy the standard PostgreSQL container with SSL (without building a new image):
* https://gist.github.com/mrw34/c97bb03ea1054afb551886ffc8b63c3b
* https://stackoverflow.com/questions/55072221/deploying-postgresql-docker-with-ssl-certificate-and-key-with-volumes

But I have not tried that myself so I'll let you decide.

.github/workflows/external-application-connection-tests.yml
```diff
@@ -0,0 +1,203 @@
...
+              postgres-ssl -c ssl=on \
+              -c ssl_cert_file=/var/lib/postgresql/server.crt \
+              -c ssl_key_file=/var/lib/postgresql/server.key
+
+      - name: Connect DB container to network
+        run: docker network connect example postgresql --alias postgresql.example.com
+
+      - name: Set up database drivers
+        run: |
+          docker exec pki dnf install -y postgresql-jdbc ongres-scram-client ongres-stringprep
```
edewata, line 140: To my understanding `ongres-scram-client` and `ongres-stringprep` will be installed automatically when we install `postgresql-jdbc`, so they don't need to be installed explicitly.

edewata (commented):
Thanks for the PR! Generally it looks good, but I think we should ensure that it's indeed creating an SSL connection. Please see my comments.

.github/workflows/external-application-connection-tests.yml
fmarco76: I have already tested this but I could not get it working without root access on my machine. If I try as normal user the files inside the container does not get the right ownership and postgrsql do not start.

edewata: That's OK, we probably can revisit this another time.

edewata (approved):
Thanks for the updates! Feel free to merge.

For future improvement we probably can copy this test into PKI's ACME tests, then simplify the JSS-PostgreSQL test to use a simple Java client connecting to PostgreSQL directly using JDBC over SSL. That way it will be easier to set up and troubleshoot if there's any problem.

fmarco76:
> copy this test into PKI's ACME tests

Yes, I totally agree. I started replicating the tests I did for the issue but after I had the feeling this was better to test for ACME. However, we need to deploy the client, I will think on that.

fmarco76:
@edewata @ckelleyRH Thanks!

PR #934, opened 2023-02-22
Add test for connection with postgresql

fmarco76:
The test to verify the communication with postgresql deploy the ACME subsystem with the DB as back-end and ssl enabled.
A new test has been written to perform a simple communication with postgresql using the only JSS.

The previous test could be deleted or moved to pki to test ACME with external DB.

### DIFF
--- a/.github/workflows/external-application-connection-tests.yml
+++ b/.github/workflows/external-application-connection-tests.yml
@@ -31,7 +31,147 @@ jobs:
           wait-interval: 30
         if: github.event_name == 'pull_request'
 
+  
   postgresql-test:
+    name: Testing connection to postrgresql
+    needs: [init, build]
+    runs-on: ubuntu-latest
+    env:
+      SHARED: /tmp/workdir/jss
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+
+      - name: Retrieve jss-builder image
+        uses: actions/cache@v3
+        with:
+          key: jss-builder-${{ github.sha }}
+          path: jss-builder.tar
+
+      - name: Load jss-builder image
+        run: docker load --input jss-builder.tar
+
+      - name: Create network
+        run: docker network create example
+
+      - name: Run container
+        run: |
+          tests/bin/runner-init.sh
+        env:
+          IMAGE: jss-builder
+          NAME: jss
+          HOSTNAME: jss.example.com
+  
+      - name: Connect JSS container to network
+        run: docker network connect example jss --alias jss.example.com
+
+      - name: Set up jss and  database drivers
+        run: |
+          docker exec jss dnf install -y postgresql-jdbc
+          docker exec -t jss sh -c 'dnf install -y /root/jss/build/RPMS/*.rpm'
+
+      - name: Create postgresql certificates
+        run: |
+          docker exec jss dnf install -y dogtag-pki
+          docker exec jss pki nss-cert-request --subject "CN=postgresql.example.com" \
+              --csr /root/sslserver.csr --ext /usr/share/pki/server/certs/sslserver.conf
+          docker exec jss pki nss-cert-issue --csr /root/sslserver.csr \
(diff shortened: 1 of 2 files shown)


fmarco76:
There is a new error with the *Sandox Test* but it is on the native code and I have not done any change on that so it should be not related to this. However, some investigation is needed.

.github/workflows/external-application-connection-tests.yml
```diff
@@ -31,7 +31,147 @@ jobs:
...
+          security.provider.3=sun.security.ssl.SunJSSE
+          security.provider.4=sun.security.rsa.SunRsaSign
+          security.provider.5=sun.security.ec.SunEC
+          security.provider.6=com.sun.net.ssl.internal.ssl.Provider
+          security.provider.7=com.sun.crypto.provider.SunJCE
+          security.provider.8=sun.security.jgss.SunProvider
+          security.provider.9=com.sun.security.sasl.Provider
+          security.provider.10=org.jcp.xml.dsig.internal.dom.XMLDSigRI
+          security.provider.11=sun.security.smartcardio.SunPCSC
+          EOF
```
edewata, line 147: Instead of creating a `java.security` file (which affects the whole JVM) we probably can use the `sslfactory` param with a custom `SSLSocketFactory` just for the PostgreSQL connection:
https://jdbc.postgresql.org/documentation/use/#connection-parameters
This way it might be easier to use and would avoid conflicts with everything else running in the JVM.

fmarco76: According to the documentation the `sslfactory` is string with the name of the class to instantiate as socket factory. Additionally:

> This class must have a zero argument constructor or a single argument constructor 

In **JSS** the only  factory I have seen is `org.mozilla.jss.ssl.javax.JSSSocketFactory` and this has a constructor with multiple parameters so cannot be directly used. If a new factory has to be implemented then I do not think make sense for this simple test but maybe I am missing something. 
@edewata, is there another approach without adding a new security provider, via `java.security` or API?

.github/workflows/external-application-connection-tests.yml
```diff
@@ -31,7 +31,147 @@ jobs:
...
+              -e POSTGRES_USER=jss \
+              postgres-ssl -c ssl=on \
+              -c ssl_cert_file=/var/lib/postgresql/server.crt \
+              -c ssl_key_file=/var/lib/postgresql/server.key
+
+      - name: Connect DB container to network
+        run: docker network connect example postgresql --alias postgresql.example.com
+
+      - name: Build the tests
+        run: docker exec jss ./build.sh --work-dir=./build
```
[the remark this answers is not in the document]
fmarco76: If it could be useful in other cases, yes!

.github/workflows/external-application-connection-tests.yml
edewata: Yeah, we can't do this with the existing code, but it's some to think about for future improvements, and hopefully it won't require major changes. To my understanding this is a pretty common way to provide a custom socket, so we should try to support this if possible.

.github/workflows/external-application-connection-tests.yml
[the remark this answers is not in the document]
edewata: We do have `pki-tests` package so, ideally, people can use it to validate that PKI works well on their system. IPA provides `python3-ipatests` too. So same idea for `jss-tests`.

.github/workflows/external-application-connection-tests.yml
fmarco76: Yes, I do not think this is a peculiarity of _Postgresql_ so in the future we should work on that. Do we should create an issue for that as a remind?

I am going to merge for now.

For the _ACME_ based test, I move it to __pki__ or just drop it?

edewata: Yes, please create a ticket for the socket factory and move the ACME test to PKI. That would be a good addition to PKI CI. Thanks!

PR #935, opened 2023-02-23
Fix warning for accessing `-Werror=maybe-uninitialized`

fmarco76:
The error stop the building only with debug enabled and all warning enabled. This is not a problem but it blocks the `sandbox-test` which exit with error.

The error is arise only in fedora rawhide and not in fedora 37.

The current logic of the application does not change.


ckelleyRH (approved):
LGTM, thanks for spotting this!

fmarco76:
@ckelleyRH Thanks!

PR #937, opened 2023-02-27
Remove Postgrsql test using ACME

fmarco76:
The test will be moved to **pki** repository because it is more an **pki ACME** subsystem test with different back-end than a **JSS** test.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #938, opened 2023-03-10
Fix some Sonar Cloud issues

fmarco76:
The method `toString()` does never return `null` and the subclasses should maintain its original behaviour. Similarly, some syncronized methods were overrided without the `syncronize` modifier.

Finally, some conditions were not possible.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/provider/java/security/KeyFactorySpi1_2.java
+++ b/base/src/main/java/org/mozilla/jss/provider/java/security/KeyFactorySpi1_2.java
@@ -185,7 +185,7 @@ public class KeyFactorySpi1_2 extends java.security.KeyFactorySpi
             // we need to chop off a leading zero byte
             if( y.bitLength() % 8 == 0 ) {
                 byte[] newBA = new byte[yBA.length-1];
-                assert(newBA.length >= 0);
+                assert(newBA.length > 0);
                 System.arraycopy(yBA, 1, newBA, 0, newBA.length);
                 yBA = newBA;
             }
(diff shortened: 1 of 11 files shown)


base/src/main/java/org/mozilla/jss/provider/java/security/KeyFactorySpi1_2.java
```diff
@@ -185,7 +185,7 @@ protected java.security.PrivateKey engineGeneratePrivate(KeySpec keySpec)
             // we need to chop off a leading zero byte
             if( y.bitLength() % 8 == 0 ) {
                 byte[] newBA = new byte[yBA.length-1];
-                assert(newBA.length >= 0);
+                assert(newBA.length > 0);
```
ckelleyRH, line 188: This `assert` does nothing, as assertions only apply with debugging enabled in the JVM which is disabled by default. This should be replaced with proper `if/else` control flow.

ckelleyRH (approved):
One comment but otherwise LGTM.

base/src/main/java/org/mozilla/jss/provider/java/security/KeyFactorySpi1_2.java
fmarco76: If the length is 0 the code here should work properly and an eventual exception will be generated later. I do not think it is necessary to modify the control flow at this level.

ckelleyRH: OK, then the `assert` should be dropped completely because it does nothing but provide false assurance.

edewata: Yeah, this particular `assert` is not very useful since it's already checked by Java, so I think we can just drop it.

edewata (approved):
I added a comment, but it LGTM regardless.

base/src/main/java/org/mozilla/jss/provider/java/security/KeyFactorySpi1_2.java
fmarco76: SonarCloud identify many issues with asserts overlapping the java checks. I think the idea was to stop the execution where the problem arise if executed in debug mode. The jave code and the assert are not doing exactly the same test. I would not drop this now but I am planning to evaluate these asserts again later.

fmarco76:
@ckelleyRH @edewata Thanks!

PR #939, opened 2023-03-17
Fix decrypt for PBKDF1

fmarco76:
Passwords for PBKDF1 and PBKDF2 in JSS have to be in different format because they are handled by different nss libraries which work with different data encoding.

Fix Bug 2115765


ckelleyRH (approved):
LGTM but defer to @jmagne 's expertise in this area.

fmarco76:
@jmagne @ckelleyRH Thanks!

PR #940, opened 2023-03-23
Fix some SonarCloud major bugs

fmarco76:
Removed a block of code performing no operations.
Add back the interrupted flag in case of a sleep operation get interrupted so the JVM behaviour is preserved.


base/src/main/java/org/mozilla/jss/pkix/crmf/CertReqMsg.java
```diff
@@ -138,26 +137,13 @@ public void verify(CryptoToken token) throws SignatureException,
...
-				//System.out.println("worked!!");
-				return; // success
-			} else {
+			if( !sig.verify(sig_from.getBits()) ) {
 				throw new SignatureException(
 					"Signed request information does not "+
 					"match signature in POP");
 			}
 
 		} else if (type == ProofOfPossession.KEY_ENCIPHERMENT) {
```
ckelleyRH, line 146: This `else if` clause is empty now, can we drop it completely?

ckelleyRH (approved):
LGTM, one suggestion.

base/src/main/java/org/mozilla/jss/pkix/crmf/CertReqMsg.java
```diff
@@ -138,26 +137,13 @@ public void verify(CryptoToken token) throws SignatureException,
...
-			POPOPrivKey keyEnc = pop.getKeyEncipherment();
-			POPOPrivKey.Type ptype = keyEnc.getType();
-			if (ptype == POPOPrivKey.THIS_MESSAGE) {
-				//BIT_STRING thisMessage = keyEnc.getThisMessage();
-				//This should be the same as from the archive control
-				//It's verified by DRM.
-
-			} else if (ptype == POPOPrivKey.SUBSEQUENT_MESSAGE) {
-				new ChallengeResponseException("requested");
-			}
```
edewata, line 160: @ladycfu @jmagne This code doesn't seem to be doing anything useful. Can we remove it? Just making sure that it's not a bug (e.g. missing a `throw` statement).

fmarco76: For reference, I have verified and the code as not been modified from the first commit:
https://github.com/dogtagpki/jss/commit/75922fe01d6bb333fc562306f8d47016da09df28#diff-c1595095f2733278b696e30b654dcc612c0b66823a33816c8e138b8489590de4R177

Additionally, the exception is not declared in the method signature.

jmagne: I think if it's been like this forever and the code doesn't do anything, it can go. I can't remember the last time I"ve looked in this area, if ever. But I like the reasoning. Leaving the comment seems good as a possible reminder if we need it in the future.

jmagne (approved):
LGTM, approving.

base/src/main/java/org/mozilla/jss/pkix/crmf/CertReqMsg.java
ladycfu: @fmarco76  sorry for the late reply. So, according to pkix/crmf/ChallengeResponseException.java, "An exception thrown when challenge response pop is requested", at least that seems to be the design.
Seems like it was used here before:
ca/src/main/java/com/netscape/cms/servlet/cert/HashEnrollServlet.java
but then was commented out for some reason?
You might want to investigate.  If you must do something about it, I'd suggest commenting it out instead of removal.

fmarco76: @ladycfu The code in `HashEnrollServlet.java` has been commented out before the project move to git. This is the file in the oldest commit in github:
https://github.com/dogtagpki/pki/blob/d0f2e4efbd3eb0f1d7f5a28e7f97c1fb4ec027bb/pki/base/common/src/com/netscape/cms/servlet/cert/HashEnrollServlet.java#L901-L913

It is impossible to know the reason why it was removed without identifying when. Anyway,  since this is not in use for so long I will go ahead removing it but I will add more information as comment just in case we could need to revert it.

fmarco76:
@ckelleyRH @edewata @jmagne  @ladycfu Thanks!

PR #941, opened 2023-03-27
Removde printStackTrace()

fmarco76:
The `printStackTrace()` output should not be used in production because it can leak information and cannot be disabled.

The method is deleted or converted to log.

Note: some tests and client code still use the method


edewata (approved):
LGTM

PR #942, opened 2023-03-28
Deprecate RSAParameterSpec.getKeySize

fmarco76:
The class RSAParameterSpec has been modified to extend the standard `java.security.spec.RSAKeyGenParameterSpec` but the method `getKeySize()` has a different capitalisation in the superclass so this was kept for compatibility.

However, all the *JSS* code has been updated to use the superclass method and there is no use in other *PKI* tools so the old method is not used and can be removed in the future.

The commit modifying the RSAParameterSpec extending the standard class is:

https://github.com/dogtagpki/jss/commit/b995847206e21b696c697f1551b0ee61e17f9497


fmarco76:
@edewata If this class is in use only inside JSS (I did a quick `grep` in the other project and cannot find it), we could deprecate the entire class and remove it in the future. Is there a reason to keep it?

edewata:
I'm not too familiar with this class, but since it was modified to extend `RSAKeyGenParameterSpec` in JSS 4.6 I think we can deprecate the entire class (instead of just the method) and remove it in the future. @jmagne Do you have any objection?

I found some leftover references to this class, so we need to replace it with `RSAKeyGenParameterSpec` too:
https://github.com/dogtagpki/jss/blob/master/base/src/main/java/org/mozilla/jss/crypto/KeyPairGenerator.java#L69
https://github.com/dogtagpki/jss/blob/master/base/src/main/java/org/mozilla/jss/crypto/KeyPairGenerator.java#L85
https://github.com/dogtagpki/jss/blob/master/base/src/main/java/org/mozilla/jss/pkcs11/PK11KeyPairGenerator.java#L351
https://github.com/dogtagpki/jss/blob/master/base/src/test/java/org/mozilla/jss/tests/TestKeyGen.java#L68
https://github.com/dogtagpki/jss/blob/master/base/src/test/java/org/mozilla/jss/tests/TestKeyGen.java#L117

jmagne (approved):
Looks simple approving and seems quite safe.

fmarco76:
@edewata @jmagne Thanks!

PR #943, opened 2023-03-28
Fix switch issue

fmarco76:
Some `switch` cases are not closed by a break.
Actually, it is not a problem because following cases are safe or not possible but `break` is added and exception in case of not reachable case.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/netscape/security/pkcs/PKCS9Attribute.java
+++ b/base/src/main/java/org/mozilla/jss/netscape/security/pkcs/PKCS9Attribute.java
@@ -744,10 +744,10 @@ public class PKCS9Attribute implements DerEncoder {
         case 14: // ExtensionRequest attribute
             value =
                     new CertificateExtensions(elems[0].toDerInputStream());
+            break;
 
-            // break unnecessary
-
-        default: // can't happen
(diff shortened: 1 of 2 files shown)


base/src/main/java/org/mozilla/jss/netscape/security/pkcs/PKCS9Attribute.java
```diff
@@ -859,9 +859,9 @@ public void derEncode(OutputStream out) throws IOException {
                 } catch (CertificateException e) {
                     throw new IOException("PKCS9 extension attributes not encoded");
                 }
-
-                // break unnecessary
+                break;
             default: // can't happen
+                throw new IOException("PKCS9 attribute index error");
```
edewata, line 864: Same thing here.

edewata (approved):
I have some minor comments, but it LGTM. Feel free to update/merge.

I generally wouldn't trust the "can't happen" comments in the code. In ideal world that case might never happen, but there could be bugs that allow it to happen, so we still need to provide the code to handle it (e.g. throwing exception) instead of just ignoring it. So you might want to remove the "can't happen" comments as well.

fmarco76:
@edewata Thanks!

PR #944, opened 2023-03-31
Tidy up some classes in provider crypto package

fmarco76:
The change are
- remove some assertions which are already verified in conditions;
- move instance variable to local;
- clean up if else conditions;
- using instanceof with assignment;
- fix signature to java recommendation.


fmarco76:
@edewata Thanks!

PR #945, opened 2023-04-07
Clean ASN1 package

fmarco76:
Removed some bugs/code smells from the `org.mozilla.jss.asn1` package. These include:
- convert `Vector` to `ArrayList`
- remove or convert to real check some `assert` conditions

There are several layout changes made by eclipse but they are in the javadoc comments.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/asn1/SEQUENCE.java
+++ b/base/src/main/java/org/mozilla/jss/asn1/SEQUENCE.java
@@ -8,7 +8,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.util.Vector;
+import java.util.ArrayList;
 
 /**
  * An ASN.1 SEQUENCE. This class is an ordered collection of ASN.1 values.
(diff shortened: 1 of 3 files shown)


base/src/main/java/org/mozilla/jss/asn1/SEQUENCE.java
```diff
@@ -403,14 +403,13 @@ public ASN1Value decode(Tag tag, InputStream istream)
...
                     // optional or have a default.  Otherwise, it's an error.
-                    Element e = elements.elementAt(index);
+                    Element e = elements.get(index);
                     if ((lookAhead == null) || lookAhead.isEOC() ||
                             !e.tagMatch(lookAhead.getTag())) {
                         if (e.isRepeatable()) {
                             repeatableElement = true;
                         } else if (e.isOptional()) {
                             // put an empty entry into the SEQUENCE
-                            SEQUENCE.Element se = new SEQUENCE.Element(null, null);
```
ckelleyRH, line 413: SonarCloud is telling us that with the removal of `se` the private inner class `SEQUENCE.ELEMENT` is now unused, so you can also remove that.

ckelleyRH (approved):
LGTM. There is one further tidy up suggested by SonarCloud if you want to add it as part of this change.

fmarco76:
@ckelleyRH Thanks!

PR #949, opened 2023-04-12
Add Maven module for symkey

edewata:
A `pom.xml` has been added to define Maven module for symkey. The build test has been added to compare the `jss-symkey.jar` created by CMake and by Maven to ensure that they contain the same files.

For now Maven will only build symkey Java code and SonarCloud will ignore symkey native code.

The `jss.spec` has been updated to use Maven dependencies.

### DIFF
--- a/.github/workflows/build-tests.yml
+++ b/.github/workflows/build-tests.yml
@@ -51,8 +51,24 @@ jobs:
 
     - name: Compare jss.jar
       run: |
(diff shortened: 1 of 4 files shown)


ckelleyRH (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #951, opened 2023-04-12
Fix native filename

fmarco76:
Running `mvn install` generate error for messing package name. Investigating and checking the code the artificatId has to be the same of the library generated.

https://stackoverflow.com/questions/25138413/java-jni-maven-native-maven-plugin-how-to-set-shared-library-final-name https://github.com/mojohaus/maven-native/blob/master/native-maven-plugin/src/main/java/org/codehaus/mojo/natives/plugin/NativeInitializeMojo.java

The install phase is relevant to deploy in the artificat in the local database and use the artifact as dependency

### DIFF
--- a/native/pom.xml
+++ b/native/pom.xml
@@ -11,7 +11,7 @@
         <version>${revision}</version>
     </parent>
 
-    <artifactId>jss-native</artifactId>
+    <artifactId>libjss</artifactId>
     <packaging>so</packaging>
 
     <dependencies>


ckelleyRH (approved):
Works on my machine, thanks!

ckelleyRH:
Do we need to update PKI now with the new name?

edewata (approved):
LGTM. It's weird that the answer that suggested modifying `<artifactId>` got a 0 points and the answer that suggested modifying `<linkerFinalName>` got 4 points.

fmarco76:
> Do we need to update PKI now with the new name?

Actually, PKI links only of jss and jss-symkey but I cannot  getting it to work. I am testing.

edewata:
> Do we need to update PKI now with the new name?

I think we have not defined JSS dependency in PKI using Maven yet, so there's nothing to change in PKI right now.

fmarco76:
@edewata @ckelleyRH Thanks!

PR #953, opened 2023-04-13
Make maven artifcat usabale

fmarco76:
When parent version is managed with a `revision` property in the submodules the generated artificats will have the `version` field not updated do they cannot be consumed by other projects. The solution is to add a plugin transforming the `pom.xml` for the artifact.

References:

- https://maven.apache.org/maven-ci-friendly.html#install-deploy
- https://jeanchristophegay.com/en/posts/maven-unique-version-multi-modules-build/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,6 +19,36 @@
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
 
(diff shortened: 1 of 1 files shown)


fmarco76:
@edewata Thanks!

PR #954, opened 2023-04-14
Add build.sh --without-java and --without-native

edewata:
New options have been added into `build.sh` to support building Java and native binaries separately.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,8 @@ if (DEFINED ENV{WITH_INTERNET})
 endif()
 option(TEST_WITH_INTERNET "When enabled, runs various tests which require an internet connection. " ${TEST_WITH_INTERNET_ENV})
 
+option(WITH_JAVA "Build Java binaries." TRUE)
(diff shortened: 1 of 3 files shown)


ckelleyRH (approved):
LGTM!

fmarco76 (approved):
LGTM but it does not work as in the description. If the build script is executed with the option `--without-java` I am expecting that classes are not compiled and javadoc is not generated but they are created. However, they are not installed.

edewata:
@ckelleyRH @fmarco76 Thanks!

Yes, some of the native code currently depends on C header files that are generated by `javac` while compiling the Java classes. It's possible to break this dependency by storing the pre-generated C header files into the source repository. We can discuss this separately later. As for Javadoc, you should be able to skip that with `--without-javadoc`.

So with this change the build process will happen in multiple steps:
```
make java
make native
make javadoc
```
When we're ready we will convert at least the first step to Maven:
```
mvn package
make native
make javadoc
```
We can decide later whether we want to convert the other steps as well.

PR #955, opened 2023-04-17
Fix some bugs in org.mozilla.jss.pkix

fmarco76:
The main issuees fixed are:

- override of equals and and hash code, (they should be done both)
- check byde read from stream
- removed not possible exceptions
- syntax compliant to Java best practice


ckelleyRH (approved):
LGTM!

fmarco76:
@ckelleyRH Thanks!

PR #959, opened 2023-05-16
Convert tests to JUnit5 (PrintableConverterTest)

fmarco76:
This is the first junit test converted to junit 5.

In order to run the test also a new runner class is introduced. 

Here the migration documentation: https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4

The remaining tests make use of junit 4 or makes use of a custom  approach. All of them could be converted to junit 5.

At compile time there is a warning with `org.apiguardian.api.API$Status` which seems related to a library problem in junit 5 but it can be ignored according to many post (e.g. https://stackoverflow.com/questions/46702273/warning-unknown-enum-constant-status-stable).

### DIFF
--- a/.classpath
+++ b/.classpath
@@ -4,14 +4,28 @@
 	<classpathentry kind="src" path="base/src/test/java"/>
 	<classpathentry kind="src" path="base/src/broken_test/java"/>
 	<classpathentry kind="src" path="examples/src/main/java"/>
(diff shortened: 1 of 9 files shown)


edewata (approved):
LGTM

ckelleyRH (approved):
LGTM!

fmarco76:
@ckelleyRH @edewata Thanks!

PR #960, opened 2023-05-16
Suppress warnings for deprecated algorithms

edewata:
Algorithms such as SHA-1 and MD5 are deprecated but they still have valid uses and in certain cases cannot be replaced easily. The problem is currently code that uses these algorithms generates warnings that can undermine user's confidence in the code. To avoid the problem, the warnings have been converted into regular debug messages which will only appear in debug mode.


jmagne (approved):
Approving :)

ckelleyRH (approved):
LGTM

fmarco76 (approved):
LGTM

edewata:
@jmagne @ckelleyRH @fmarco76 Thanks!

I'll merge this PR to master branch, but if you think we need to apply the same changes to existing releases just let me know.

edewata:
OK, I cherry-picked the change to `v5.4` branch. It looks like there is a fix for another bug in that branch already, so we can piggy-back the same bug (i.e. whenever the package is rebased to pick up the fix this change will be include as well). Thanks!

PR #961, opened 2023-05-17
Upgdate all junit tests to JUnit5

fmarco76:
All JUnit tests use  `junit5` and `junit4` removed.


ckelleyRH (approved):
LGTM, thanks! It is interesting that Sonar highlighted the public methods here but not in my change for PKI. I thought the test rules were identical for all repos (at least, that is how I tried to set it up).

fmarco76:
@ckelleyRH Thanks!

PR #962, opened 2023-05-22
Replace ECCurve_Code.ordinal() with code()

edewata:
Previously the `Enum.ordinal()` was used to determine the EC curve code which is not reliable since the position of the constants in the enumeration is not guaranteed. A new field has been added into `ECCure_Code` to store a permanent curve code regardless of constant position.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/pkcs11/PK11KeyPairGenerator.java
+++ b/base/src/main/java/org/mozilla/jss/pkcs11/PK11KeyPairGenerator.java
@@ -36,88 +36,108 @@ public final class PK11KeyPairGenerator
 {
     public static Logger logger = LoggerFactory.getLogger(PK11KeyPairGenerator.class);
 
-    // curve code for getting the actual EC curve
+    /**
+     * The ECCurve_Code enum defines a code for each EC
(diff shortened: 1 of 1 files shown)


ckelleyRH (approved):
This is slightly more clear that the order of enum entries matters, but maybe a comment explicitly stating that the order is being depended on not to change would be good too?

Better still would be to refactor this so it doesn't require the ordinal (or replacement) at all. I took a look and it seems like this code has been extended to do multiple jobs over the years and I wouldn't want to modify it further without a better reason, so approving.

edewata:
@ckelleyRH Thanks! I added a comment for the `ECCurve_Code` enum, could you take a look?

I'm not sure how the `initialize()` is used, so I don't want to change how the `code` is used for now, but I have several more PRs to simplify / remove duplicate code related to EC curves.

ckelleyRH:
> @ckelleyRH Thanks! I added a comment for the `ECCurve_Code` enum, could you take a look?

LGTM!

> I'm not sure how the `initialize()` is used, so I don't want to change how the `code` is used for now, but I have several more PRs to simplify / remove duplicate code related to EC curves.

Looking forward to reviewing them!

PR #963, opened 2023-05-22
Refactor PK11KeyPairGenerator.getCurveCodeByName()

edewata:
The `PK11KeyPairGenerator.getCurveCodeByName()` has been modified to use `ECCurve_Code.valueOf()` to translate the curve name into the curve code instead of using a separate `Hashtable`. If the curve name is not defined in `ECCurve_Code`, the `valueOf()` will throw an `IllegalArgumentException`.


ckelleyRH (approved):
LGTM!

ckelleyRH (approved):
LGTM!

ckelleyRH:
So good I approved it twice :rofl:

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #964, opened 2023-05-23
Code clean up in CertTemplate

ckelleyRH:
* Use ternary operator where appropriate
* Document empty constructor
* Put array designator on the type
* Remove unused indent variable, and subsequently unused vars/args
* Remove exception that cannot be thrown from print()
* Use pattern-matching instanceof


fmarco76 (approved):
LGTM

ckelleyRH:
Thanks @fmarco76 !

PR #965, opened 2023-05-23
Code clean up in ObjectIdentifier

ckelleyRH:
* Flatten logic in getObjectIdentifier()
* Use pattern-matching instanceof and ternary operators
* Remove unnecessary Boolean literals
* Reorder modifiers to match the JLS
* Rename variables to match the JLS naming conventions
* Move array designator to the type
* Declare each var on its own line
* Catch Exception instead of Throwable
* Declare mOIDs with a generic type, use HashMap, lower visibility
* Prefer StringBuilder to StringBuffer
* Remove exceptions that cannot be thrown


fmarco76 (approved):
LGTM. I have a concern with the `Throwable` <-> `Exception` replacement. They are not the same and it is not clear to me if the first was used because needed or just over reacting to the problems. Feel free to merge.

PR #969, opened 2023-05-24
Drop legacy CSR header/footer

edewata:
The `PK11Token.generateCertRequest()` has been modified to generate a CSR with the header and footer described in RFC 7468.

See also:

* https://datatracker.ietf.org/doc/html/rfc7468#section-7
* https://github.com/dogtagpki/pki/issues/3843


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #970, opened 2023-05-25
Fix OoM in JSSEngineReferenceImpl object

fmarco76:
If TLS connection is terminated by the server with a `close_notify` tomcat will call the `closeOutbound()` method but the `closeInbound()` is never called so the cleanup cannot be done at the end.

It is possible to test the problem with a tomcat instance. If the option `-H 'Connection: close'` is present only the `closeoutbound()` is called.


fmarco76:
The `finalize()` method is not needed because the `cleanup()` is always invoked.

As result the PR #928 should not be necessary.

edewata (approved):
I'll defer to others since I'm not familiar with the code, but if you don't get additional feedback feel free to merge it since all tests passed.

uplogix-mmcclain:
The system no longer leaks on every connection with this change; I can't tell for certain if this completely fixes the issue called out in #928.

I don't think removing the finalize() is a good idea.  There seem to be some objects that are being left behind by the garbage collection.  I don't currently have the code instrumented, but I don't see where you ensured cleanup() was called, so cleanupSSLFD() and session.close() may not be called.

fmarco76:
> The system no longer leaks on every connection with this change; I can't tell for certain if this completely fixes the issue called out in #928.

If there is no leak in every connection I think the #928 can be closed. Even if there is still a leak problem the solution should be different preserving the "X509ExtendedTrustManager". Is this OK?

> 
> I don't think removing the finalize() is a good idea. There seem to be some objects that are being left behind by the garbage collection. I don't currently have the code instrumented, but I don't see where you ensured cleanup() was called, so cleanupSSLFD() and session.close() may not be called.

Actually, during `wrap()` and `unwrap()` the `tryClanup()` is invoked if everything is correctly done and the `cleanup()` in case of exception. So objects should not been leaked.

Without, this patch there were cases where `trycleanup()` was not working because inbound connection was not closed by tomcat. As a result resources were leaked. This was evident with the custom trust manager leak identified in #928 but the `c` resources allocated were also leaked, with or without custom trust manager. Now, both of them should not be a problem.

Additionally, using `finalize()` has many drawbacks and it is currently deprecated from [Java API](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Object.html#finalize()). 

A temporary solution could be revert the `finalize()` method and add some warning to identify if there are situation were potential leak could still happen. What do you think @edewata, @ckelleyRH and @jmagne?

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1274,7 +1274,7 @@ public SSLEngineResult unwrap(ByteBuffer src, ByteBuffer[] dsts, int offset, int
             SSLException excpt = ssl_exception;
             ssl_exception = null;
             handshake_state = SSLEngineResult.HandshakeStatus.NEED_WRAP;
-            tryCleanup();
+            cleanup();
```
cipherboy, line 1277: Wow, this is a really interesting change.

I think, `finalize` made sense to me, as a catch-all for any failures of applications to conform with expectations. [If my memory serves correctly, `Cleaner`](https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Cleaner.html) is its replacement, and I thought we had looked into that when we were considering the JDK9+ port. Maybe @jmagne or @edewata can figure out where we had those discussions back when Dogtag was still Java 8-only.

I think for this change to work, you'd need to run the `wrap(...)` call and store the results of it _before_ you call cleanup.

A counter-test would be to present a bad CA chain as a client. This should trigger a server to verify the chain (during an unwrap) and subsequently need to send an outbound alert (`NEED_WRAP`).

What I expect to happen -- and again, disclaimers about memory aside, happy to be wrong here -- is that because you've cleaned up the SSLFD and underlying NSS sockets, you'd get no data to wrap. This means the client would "just" see a peer disconnect error and not see an actual verification error. 

Anyhow, some thoughts from the peanut gallery :D

fmarco76: > Wow, this is a really interesting change.
> 
> I think, `finalize` made sense to me, as a catch-all for any failures of applications to conform with expectations. [If my memory serves correctly, `Cleaner`](https://docs.oracle.com/javase/9/docs/api/java/lang/ref/Cleaner.html) is its replacement, and I thought we had looked into that when we were considering the JDK9+ port. Maybe @jmagne or @edewata can figure out where we had those discussions back when Dogtag was still Java 8-only.
> 

Yes, `Cleaner` is a possible alternative but I do not think it is needed for this case. Nevertheless, I am reverting the `finlize()` for the moment and if it is the case re-evaluate the approach in a future PR.


> I think for this change to work, you'd need to run the `wrap(...)` call and store the results of it _before_ you call cleanup.
> 
> A counter-test would be to present a bad CA chain as a client. This should trigger a server to verify the chain (during an unwrap) and subsequently need to send an outbound alert (`NEED_WRAP`).
> 
> What I expect to happen -- and again, disclaimers about memory aside, happy to be wrong here -- is that because you've cleaned up the SSLFD and underlying NSS sockets, you'd get no data to wrap. This means the client would "just" see a peer disconnect error and not see an actual verification error.
> 

Yes, I think for this you are totally right, so I have reverted the change.

> Anyhow, some thoughts from the peanut gallery :D

Really appreciate, thanks!

fmarco76:
@edewata @cipherboy @uplogix-mmcclain Thanks!

edewata:
@fmarco76 @cipherboy @uplogix-mmcclain Thanks a lot for you help! Sorry I couldn't really provide much help here since this is very low-level and I'm not an expert in this area. The only thing I can suggest is to add memory-related tests so we can catch these types of issues in the future.

PR #971, opened 2023-06-08
Fix sonar bugs

fmarco76:
Change in `AVA` class:
    - remove not possible `throws` exception
    - check number of byte read from InputStream
    - Use UTF8 standard charset

Change in `PK11Token` class:
    - Override `hashCode()` since `equals()` is override
    - Remove useless assignment
    - Refactor method `generateRequest(String, int, String, byte[], byte[], byte[])`
    - Removed `catch` without logic
    - Remove `throws` exeception of subclasses when the parent is thrown


base/src/main/java/org/mozilla/jss/netscape/security/x509/AVA.java
```diff
@@ -131,17 +131,16 @@ public AVA(InputStream in) throws IOException {
         try {
             // convert from UTF8 bytes to java string then parse it.
             byte[] buffer = new byte[in.available()];
-            in.read(buffer);
-
-            Charset charset = Charset.forName("UTF-8");
-            CharsetDecoder decoder = charset.newDecoder();
-
-            CharBuffer charBuffer = decoder.decode(ByteBuffer.wrap(buffer));
+            if (in.read(buffer) > 0) {
```
edewata, line 134: If the stream has reached EOF should we throw an exception instead of letting to `AVA` object be created with empty `oid` and `value`?

edewata (approved):
I have one question but feel free to merge since it seems to be an existing issue.

base/src/main/java/org/mozilla/jss/netscape/security/x509/AVA.java
fmarco76: I am not sure if an empty AVA should exist. If it shouldn't then we should verify EOF is not reached but also the bytes contained in the buffer.
However, in this PR I would not change the logic, if it is the case we can open an issue and fix in a separate PR.

edewata: Yeah, I think an empty `AVA` should not exist (especially if it's caused by EOF) since it's probably unusable anyway. It might be better to convert this constructor into a factory method. We can address this separately.

fmarco76:
@edewata Thanks!

PR #972, opened 2023-06-23
Add IPs address to the session

fmarco76:
The SSLEngine session "JSSSession" has been extended to container the IP addresses of the client and the server. These are used for the audit and have not other use in the protocol. By design the SSLEngine should be unaware of the underlying communication but this is need to keep the original audit format required for the certification.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -1262,9 +1262,10 @@ public class JSSEngineReferenceImpl extends JSSEngine {
             }
         } while (this_src_write != 0 || this_dst_write != 0);
 
-        if (seen_exception == false && ssl_exception == null) {
-            ssl_exception = checkSSLAlerts();
-            seen_exception = (ssl_exception != null);
+        SSLException checkException = checkSSLAlerts();
+        if (checkException != null && !seen_exception) {
+            ssl_exception = checkException;
+            seen_exception = true;
         }
 
         // Before we return, check if an exception occurred and throw it if
--- a/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -968,7 +968,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
             }
 
             debug("JSSEngine: Got inbound alert: " + event);
-
+            event.setEngine(this);
             // Fire inbound alert prior to raising any exception.
             fireAlertReceived(event);
 
@@ -991,7 +991,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
             }
 
             debug("JSSEngine: Got outbound alert: " + event);
-
+            event.setEngine(this);
             // Fire outbound alert prior to raising any exception. Note that
             // this still triggers after this alert is written to the output
             // wire buffer.
--- a/base/src/main/java/org/mozilla/jss/ssl/javax/JSSSession.java
+++ b/base/src/main/java/org/mozilla/jss/ssl/javax/JSSSession.java
@@ -1,16 +1,23 @@
 package org.mozilla.jss.ssl.javax;
 
-import java.lang.AutoCloseable;
-import java.security.cert.Certificate;
-import javax.security.cert.X509Certificate;
 import java.security.Principal;
+import java.security.cert.Certificate;
 import java.util.HashMap;
 
-import javax.net.ssl.*;
+import javax.net.ssl.SSLPeerUnverifiedException;
+import javax.net.ssl.SSLSession;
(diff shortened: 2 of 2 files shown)


fmarco76:
@edewata there are several missing dependencies during the build but I have not done any change on that. Are they related to the update done for _Packit_?

ckelleyRH:
> @edewata there are several missing dependencies during the build but I have not done any change on that. Are they related to the update done for _Packit_?

Yes there is no `pki` build in the COPR at the minute (I'm working on it now). Interestingly, yours fails for the `tomcat` dependencies, and they are there :thinking:

fmarco76:
> yours fails for the `tomcat` dependencies,

This is strange. I have modified dependencies on `tomcatjss` but not here. It is better to verify what is the problem when the other problem are fixed

base/src/main/java/org/mozilla/jss/ssl/javax/JSSSession.java
```diff
@@ -315,4 +325,21 @@ public int getPeerPort() {
...
+        this.localAddr = localAddr;
+    }
+
+    public String getRemoteAddr() {
+        return remoteAddr;
+    }
+
+    public void setRemoteAddr(String remoteAddr) {
+        this.remoteAddr = remoteAddr;
+    }
```
edewata, line 343: The class already has a `peerHost`. Should we just use that instead of adding a `remoteAddr`?

fmarco76: > The class already has a `peerHost`. Should we just use that instead of adding a `remoteAddr`?

The peerHost contains the hostname of the client and is assigned by [JSSEngine](https://github.com/dogtagpki/jss/blob/60ef42cb76dc85d922a7677943f17b7afe302e6e/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngine.java#L241). The audit has the IP addresses of client and server.

fmarco76:
Additionally, the handshake failure would not fire the audit event because there was not invocation of `checkSSLAerts()` in case of an exception was caught during the unwrap. 
Therefore the invocation of the check has been moved in order to be verified independently of the exceptions.

ladycfu (approved):
Nice catch for the missing checkSSLAlerts() call!

base/src/main/java/org/mozilla/jss/ssl/javax/JSSSession.java
edewata: @ladycfu Do we need to capture IP addresses specifically, or are hostnames fine too?

ladycfu: the current version of pp says we are allowed to provide at least one of  [IP Address, FQDN, user FQDN, or cert DN], however it also added that in the future, they will all be required.

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1262,9 +1262,10 @@ public SSLEngineResult unwrap(ByteBuffer src, ByteBuffer[] dsts, int offset, int
...
         } while (this_src_write != 0 || this_dst_write != 0);
 
-        if (seen_exception == false && ssl_exception == null) {
-            ssl_exception = checkSSLAlerts();
-            seen_exception = (ssl_exception != null);
+        SSLException checkException = checkSSLAlerts();
+        if (checkException != null && !seen_exception) {
+            ssl_exception = checkException;
+            seen_exception = ssl_exception != null;
         }
```
edewata, line 1269: I'm not really familiar with this code, but it looks like the original code would check for SSL alerts then stop checking once it has seen an exception, and the new code will always check for SSL alerts but discard all exceptions after the first one. Is that the intended change?

I think the new code can be simplified like this:
```
SSLException checkException = checkSSLAlerts();
if (checkException != null && !seen_exception) {
    ssl_exception = checkException;
    seen_exception = true;
}
```

edewata (approved):
I have one question due to a possible redundancy, but no objections if that's required to fix the issue. I'll defer to @ladycfu and @jmagne for merging. I suppose the more critical changes are in [tomcatjss/pull/73](https://github.com/dogtagpki/tomcatjss/pull/73).

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
fmarco76: > I'm not really familiar with this code, but it looks like the original code would check for SSL alerts then stop checking once it has seen an exception, and the new code will always check for SSL alerts but discard all exceptions after the first one. Is that the intended change?

The problem with the previous code was that if an exception raise [here](https://github.com/dogtagpki/jss/blob/e22202201397a282a486a46454afc74ca6dc726d/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java#L1259) the method `checkSSLAlerts()` is never invoked and TLS alerts are not propagated (and audit are not written). With this change it is invoked in any case but the first error is reported.

PR #975, opened 2023-07-31
Import Tomcat JSS project

edewata:
To simplify package maintenance, the code from Tomcat JSS master branch has been imported into `tomcat` and `tomcat-9.0` modules which will be distributed as `jss-tomcat` RPM package:
https://github.com/edewata/jss/blob/tomcatjss/docs/changes/v5.5.0/Packaging-Changes.adoc

Later PKI will need to be updated to depend on `jss-tomcat` instead of `tomcatjss`, then `tomcatjss` can be deprecated.


ckelleyRH (approved):
LGTM!

edewata:
@ckelleyRH Thanks!

PR #977, opened 2023-08-02
Code cleanup in X509CertImpl

ckelleyRH:
* Override equals/hashCode as we add fields and the superclass overrides equals
* Extract constants to reduce string duplication
* Don't create vars just to return them
* Invert some negated logic for readability, remove unnecessary else clauses and use ternary operator where appropriate
* Remove declared exception that cannot be thrown
* Delete commented out code
* Put array designators on the type
* Use try-with-resources
* Combine identical catch clauses
* Remove unnecessary Boolean literal and semicolons


fmarco76 (approved):
LGTM

There is a failing test which seem uncorrelated with the changes. Maybe it is a temporary problem. We can verify with next PRs.

ckelleyRH:
> LGTM
> 
> There is a failing test which seem uncorrelated with the changes. Maybe it is a temporary problem. We can verify with next PRs.

Thanks @fmarco76 ! Yeah it looks like a problem downloading packages from Fedora COPR, I'll run it one more time and the merge.

PR #979, opened 2023-08-08
Update test for Tomcat connector with NSS database

edewata:
The test for Tomcat connector with NSS database has been updated to validate the SSL connection using `sslscan`.

### DIFF
--- a/.github/workflows/tomcat-tests.yml
+++ b/.github/workflows/tomcat-tests.yml
@@ -32,13 +32,18 @@ jobs:
         if: github.event_name == 'pull_request'
 
   # docs/admin/server/Configuring-HTTPS-Connector-with-NSS-Database.adoc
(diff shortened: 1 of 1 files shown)


ckelleyRH:
> @fmarco76 Thanks!
> 
> There seems to be a problem with Fedora repo which hopefully is just temporary, and there's a known problem with TPS which is under investigation right now.

Indeed, we are part way through the branching event for F39 so different services can have different ideas about what `fedora:latest` means for a day or so, while the F39 infra gets spun up, repos become available etc.

I've noticed in addition that at least one CI run failed because of a `dnf/dnf5` clash, and the `dnf5` as default for F39 change was meant to be reverted, so we need to keep an eye on that and report it to the Fedora devel list if it is not just transient as part of the release branching.

PR #980, opened 2023-08-08
Update RPM test

edewata:
The RPM test has been updated to compare JAR files from RPM packages (which were built using XMvn) against JAR files built directly using Maven.


fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #982, opened 2023-08-10
Fix the build tests

ckelleyRH:
They were failing because the `rawhide` image in Docker registry is from an old compose, and we are pulling shiny packages from the latest compose, which was causing `dnf`/`dnf5` compat issues. We can prevent this going forward by running images from quay.io, which is updated with every successful `rawhide` compose. 

I did some code clean up in my fork to trigger the CI, so I figure we may as well keep it. It was supposed to be minor but apparently Eclipse saw fit to tidy up some whitespace on-save so apologies about the diff pollution :smile:

### DIFF
--- a/.github/workflows/build-tests.yml
+++ b/.github/workflows/build-tests.yml
@@ -159,7 +159,8 @@ jobs:
   sandbox-test:
     name: Sandbox Test
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


edewata (approved):
LGTM

ckelleyRH:
Thanks @edewata !

PR #983, opened 2023-08-28
Push IPs address to SSLEngine session

fmarco76:
**SSLEngine** is by design unaware of the underlying communication channel. In tomcat the communication channel is started by the classes `NioEndpoint` and it is maintained in `SecureNioChannel` which will create the buffer used with the SSLEngine in order to wrap and unwrap the messages.

To allow the audit of TLS messages to include IP addresses of the client and server, the above classed have been extended in order to store the IPs in the SSLEngine session after its creation.

Replace the tomcatJSS PR#73
(https://github.com/dogtagpki/tomcatjss/pull/73)


fmarco76:
For the code coming from tomcat I have indicated the differences in the file and add the copyright disclaimer for each file following the indication from the [Free Software Law Center](https://softwarefreedom.org/resources/2007/gpl-non-gpl-collaboration.html), section 2.2. If I have understood correctly licence in spec file should not be modified because that is the distribution licence for the artefact and this will not change, we need to incorporate the copyright only in the file. Is this correct?

ladycfu (approved):
Hi @fmarco76 , I can see that comments from PR #73 are addressed.  Thanks!  So, this (PR #983) is to go with PR #972 ?  I see that both are for the master branch.  So after that, they will be ported to v10.13?  Anyway, as far as I can tell with my limited knowledge with Tomcat, this looks good to me, as long as it works.  I recall that you put in some test cases in one of the tickets before.  Please put those in the proper Bugilla bug for QE to follow.  Thanks!

fmarco76:
Apache-2.0 added in `jss.spec`.

fmarco76:
> So after that, they will be ported to v10.13?

Yes, I will port to the other branch after merge.

PR #984, opened 2023-09-04
Bug_2233740 Fix wrong port number in log and audit

fmarco76:
If the connection is created from a java socket the port number was reported incorrectly because a conversion from host to network format was missed.


ladycfu (approved):
Great catch!  Thanks.  ACK.

fmarco76:
@ladycfu Thanks!

PR #985, opened 2023-09-07
Fix null exception in RevocationReason

fmarco76:


edewata (approved):
OK, so at least this will avoid NPE if the input `string` is `null`, but the caller will need to handle a possible `null` return value. I think we can just return `UNSPECIFIED` instead of `null`, but I won't object this change. Feel free to merge, thanks!

fmarco76:
@edewata Thanks!

edewata:
Just a thought, not sure if we really need to worry about this since it might be a corner case. Let's say we call `valueOf()` with an invalid reason string, the method will also return a `null`, so we can't really distinguish between a `null` input string and an invalid one. Maybe we should throw an exception if `LABELS.get()` returns `null`?

fmarco76:
Looking at the code If `null` is returned we should get an NPE in the following operations. The server will return a 500 error.

Therefore the options are:

- handle the null and reporting to the user
- replace with `UNSPECIFIED` and log a warning.

Since it is a corner case I would go with the latter.

edewata:
Generally I'd avoid using warnings since I'm assuming the admin will not be looking at the logs very closely all the time, so the problem might not be noticeable until a worse problem happens. But again, this is probably a corner case, so it's a low priority.

PR #986, opened 2023-09-07
Update PostgreSQL test

edewata:
The PostgreSQL test has been updated to use LDAP SDK and PKI packages from Quay instead of COPR since it is generally more reliable and older packages will not expire.


ckelleyRH (approved):
LGTM, thanks!

fmarco76 (approved):
LGTM

edewata:
@ckelleyRH @fmarco76 Thanks!

PR #988, opened 2023-11-01
Add PK11Store.importCert()

edewata:
Currently none of the cert import methods provided by JSS works like `certutil -A` since they call different NSS functions so in some cases it's necessary to call this external command from Java which could be problematic and does not work well with HSM.

To address the problem the `PK11Store.importCert()` has been added to call the same NSS functions used by `certutil -A` so it's no longer necessary to call this external command from Java.

https://github.com/dogtagpki/nss/blob/master/cmd/certutil/certutil.c#L102-L182


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #990, opened 2024-02-06
Fix BASE_IMAGE param

edewata:
The tests in GitHub and Azure DevOps have been modified to run using the latest Fedora by default, but it can be changed using the `BASE_IMAGE` param.

The GitHub tests have also been updated to reuse the images created by the build workflow. The build workflow has also been modified to store the images in a single cache instead of separate caches.

The `runner-init.sh` has been modified to use a positional argument to specify the container name and use `jss-runner` image by default.

https://github.com/dogtagpki/pki/wiki/Configuring-Test-OS

### DIFF
--- a/.github/workflows/build-tests.yml
+++ b/.github/workflows/build-tests.yml
@@ -3,19 +3,36 @@ name: Build Tests
 on: [push, pull_request]
 
 jobs:
+  wait-for-build:
+    name: Waiting for build
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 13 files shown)


edewata:
I think the Rawhide build failure is caused by NSS changes on Rawhide (issue https://github.com/dogtagpki/jss/issues/991), not by this PR.

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

Issue #991, opened 2024-02-06
Build failure on Rawhide


edewata:
JSS has a copy of `kea_alg_defs` array definition from an older NSS, but the array size (i.e. `ssl_kea_size`) is still defined in NSS:
https://github.com/dogtagpki/jss/blob/master/native/src/main/native/org/mozilla/jss/ssl/SSLCipher.c#L26-L37
https://github.com/dogtagpki/nss/blob/NSS_3_96_BRANCH/lib/ssl/sslt.h#L86

Rawhide has a new NSS 3.97 which [added new elements into the array](https://github.com/dogtagpki/nss/commit/212f032fd7b9250276505b0b1f96d7939a195777), so the array size has changed as well:
https://github.com/dogtagpki/nss/blob/NSS_3_97_BRANCH/lib/ssl/ssl3con.c#L369-L381
https://github.com/dogtagpki/nss/blob/NSS_3_97_BRANCH/lib/ssl/sslt.h#L88

Building JSS on Rawhide is failing since the size of the array defined in the current JSS no longer matches the array size defined in the newer NSS:
```
PR_STATIC_ASSERT(PR_ARRAY_SIZE(kea_alg_defs) == ssl_kea_size);
```
Here are the logs:
https://github.com/edewata/jss/actions/runs/7805544892/job/21290171658#step:7:3443
```
#15 29.97 In file included from /usr/include/nspr4/pratom.h:14,
#15 29.97                  from /usr/include/nspr4/nspr.h:9,
#15 29.97                  from /root/jss/build/BUILD/jss-5.5.0-alpha3/native/src/main/native/org/mozilla/jss/ssl/SSLCipher.c:1:
#15 29.97 /usr/include/nspr4/prtypes.h:556:38: error: size of array ‘arg’ is negative
#15 29.97   556 |     extern void pr_static_assert(int arg[(condition) ? 1 : -1])
#15 29.97       |                                      ^~~
#15 29.97 /root/jss/build/BUILD/jss-5.5.0-alpha3/native/src/main/native/org/mozilla/jss/ssl/SSLCipher.c:37:1: note: in expansion of macro ‘PR_STATIC_ASSERT’
#15 29.97    37 | PR_STATIC_ASSERT(PR_ARRAY_SIZE(kea_alg_defs) == ssl_kea_size);
#15 29.97       | ^~~~~~~~~~~~~~~~
[...]

PR #992, opened 2024-02-07
Update kea_alg_defs to match NSS 3.97

edewata:
The `kea_alg_defs` array in `SSLCipher.c` has been updated to match the one defined in NSS 3.97. The assertion that compares the size of the array with `ssl_kea_size` has also been removed. These changes will allow JSS to work with the newer NSS 3.97 on Fedora Rawhide as well as the older NSS versions on other platforms.

Resolves: https://github.com/dogtagpki/jss/issues/991

Note: With these changes the build now works on Rawhide and most tests pass too. There are some failures, but they don't seem to be related so they need to be investigated separately:

* https://github.com/edewata/jss/actions/runs/7821129758/job/21337674464#step:11:559
* https://github.com/edewata/jss/actions/runs/7821129726/job/21337673840#step:7:2220


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

Issue #993, opened 2024-02-08
PKCS11Constants test failed on Rawhide


edewata:
The constants in `PKCS11Constants.java` were generated from an older NSS version. NSS 3.97 on Rawhide introduced some new constants, so the CI failed with these messages:

https://github.com/edewata/jss/actions/runs/7821129726/job/21337673840#step:7:2220

```
 6328,6334d6327
<     public static final long CKK_NSS_KYBER = 0xCE534355L;
< 
<     /**
<      * Content automatically generated; see NSS documentation for more information.
<      *
<      * Source file: /usr/include/nss3/pkcs11n.h
<      */
6580,6586d6572
<     public static final long CKA_NSS_PARAMETER_SET = 0xCE534378L;
< 
<     /**
<      * Content automatically generated; see NSS documentation for more information.
<      *
<      * Source file: /usr/include/nss3/pkcs11n.h
<      */
7049,7062d7034
<     public static final long CKM_NSS_KYBER_KEY_PAIR_GEN = 0xCE53437DL;
< 
<     /**
<      * Content automatically generated; see NSS documentation for more information.
<      *
<      * Source file: /usr/include/nss3/pkcs11n.h
<      */
<     public static final long CKM_NSS_KYBER = 0xCE53437EL;
< 
<     /**
<      * Content automatically generated; see NSS documentation for more information.
<      *
<      * Source file: /usr/include/nss3/pkcs11n.h
<      */
7134,7147d7105
< 
<     /**
<      * Content automatically generated; see NSS documentation for more information.
<      *
<      * Source file: /usr/include/nss3/pkcs11n.h
<      */
<     public static final long CKP_NSS = 0xCE534350L;
< 
<     /**
<      * Content automatically generated; see NSS documentation for more information.
<      *
<      * Source file: /usr/include/nss3/pkcs11n.h
<      */
<     public static final long CKP_NSS_KYBER_768_ROUND3 = 0xCE534351L;
```

If the new constants in NSS 3.97 are added into JSS, the test on Rawhide will work, but the test on other platforms that have older NSS version will fail since the constants in the new jSS and older NSS will no longer match.

PR #995, opened 2024-02-16
Fix PKCS #11 constants test

edewata:
The PKCS #11 constants test failed on Fedora 40 since it uses NSS 3.97 that provides some new PKCS #11 constants. As a workaround the new constants need to be excluded until NSS 3.97 becomes available on all supported platforms.

The test has also been modified to reuse the `jss-builder` image which already has the build dependencies.

Resolves: https://github.com/dogtagpki/jss/issues/993

### DIFF
--- a/.github/workflows/pkcs11-tests.yml
+++ b/.github/workflows/pkcs11-tests.yml
@@ -48,16 +48,19 @@ jobs:
         run: |
           tests/bin/runner-init.sh jss
         env:
+          IMAGE: jss-builder
           HOSTNAME: jss.example.com
 
-      - name: Install dependencies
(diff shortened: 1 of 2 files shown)


fmarco76 (approved):
LGTM.

I have some doubt on the utility of this test in general. If we do not declare all the constants but only the one we use/recognise, should this be a problem? We can think a better approach when we will re-introduce these constants in the file.

edewata:
@fmarco76 Thanks!

IIRC older JDKs used to ship Sun package which provided these constants. Newer JDKs don't ship those packages anymore, so now we're providing the constants. I'm not sure how these constants are actually used, but we can certainly discuss that, probably with @ladycfu and @jmagne as well.

PR #997, opened 2024-03-11
Fix missing OID mappings

fmarco76:
Pretty print some additional OID:

OCSP No Check - id-pkix-ocsp-nocheck { 1.3.6.1.5.5.7.48.1.5} Inhibit Any Policy {2.5.29.54}
Extended Key Usage:
- serverAuth {1 3 6 1 5 5 7 3 1}
- clientAuth {1 3 6 1 5 5 7 3 2}
- codeSigning {1 3 6 1 5 5 7 3 3}
- emailProtection {1 3 6 1 5 5 7 3 4}

### DIFF
--- a/base/src/main/java/org/mozilla/jss/netscape/security/x509/OIDMap.java
+++ b/base/src/main/java/org/mozilla/jss/netscape/security/x509/OIDMap.java
@@ -26,6 +26,8 @@ import java.util.Iterator;
 import java.util.Properties;
 
 import org.mozilla.jss.netscape.security.extensions.AuthInfoAccessExtension;
+import org.mozilla.jss.netscape.security.extensions.InhibitAnyPolicyExtension;
+import org.mozilla.jss.netscape.security.extensions.OCSPNoCheckExtension;
 import org.mozilla.jss.netscape.security.extensions.SubjectInfoAccessExtension;
 import org.mozilla.jss.netscape.security.util.ObjectIdentifier;
 import org.slf4j.Logger;
@@ -94,8 +96,11 @@ public class OIDMap {
 
     public static final String EXT_KEY_USAGE_NAME = "ExtendedKeyUsageExtension";
     public static final String EXT_INHIBIT_ANY_POLICY_NAME = "InhibitAnyPolicyExtension";
+    private static final String EXT_INHIBIT_ANY_POLICY = ROOT + "." + InhibitAnyPolicyExtension.NAME;
     private static final String EXT_KEY_USAGE = ROOT + "." +
             EXT_KEY_USAGE_NAME;
+
+    private static final String OCSP_NO_CHECK = ROOT + "." + OCSPNoCheckExtension.NAME;
 
     private static final String CRL_NUMBER = ROOT + "." +
                                           CRLNumberExtension.NAME;
@@ -132,6 +137,8 @@ public class OIDMap {
         props.put(AUTH_KEY_IDENTIFIER, "2.5.29.35");
         props.put(SUBJ_DIR_ATTR, "2.5.29.9");
         props.put(EXT_KEY_USAGE, "2.5.29.37");
+        props.put(EXT_INHIBIT_ANY_POLICY, "2.5.29.54");
+        props.put(OCSP_NO_CHECK, "1.3.6.1.5.5.7.48.1.5");
     }
 
     // Load the default name to class map (EXTENSIONS_CLASSES)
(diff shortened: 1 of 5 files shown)


base/src/main/java/org/mozilla/jss/netscape/security/x509/OIDMap.java
```diff
@@ -132,6 +137,8 @@ private static void loadNamesDefault(Properties props) {
         props.put(AUTH_KEY_IDENTIFIER, "2.5.29.35");
         props.put(SUBJ_DIR_ATTR, "2.5.29.9");
         props.put(EXT_KEY_USAGE, "2.5.29.37");
+        props.put(EXT_INHIBIT_ANY_POLICY, "2.5.29.54");
```
edewata, line 140: This OID is already defined in `InhibitAnyPolicyExtension.OID`. Should we use that instead?

base/src/main/java/org/mozilla/jss/netscape/security/x509/OIDMap.java
```diff
@@ -132,6 +137,8 @@ private static void loadNamesDefault(Properties props) {
...
+        props.put(OCSP_NO_CHECK, "1.3.6.1.5.5.7.48.1.5");
```
edewata, line 141: This is also already defined in `OCSPNoCheckExtension.OID`. Should we use that one?

base/src/main/java/org/mozilla/jss/netscape/security/x509/OIDMap.java
```diff
@@ -166,6 +173,10 @@ private static void loadClassDefault(Properties props) {
                   "org.mozilla.jss.netscape.security.x509.SubjectDirAttributesExtension");
         props.put(EXT_KEY_USAGE,
                   "org.mozilla.jss.netscape.security.extensions.ExtendedKeyUsageExtension");
+        props.put(EXT_INHIBIT_ANY_POLICY,
+                "org.mozilla.jss.netscape.security.extensions.InhibitAnyPolicyExtension");
+        props.put(OCSP_NO_CHECK,
+                "org.mozilla.jss.netscape.security.extensions.OCSPNoCheckExtension");
```
edewata, line 179: To ensure that the class names and packages are correct we probably should use `InhibitAnyPolicyExtension.class.getName()` and `OCSPNoCheckExtension.class.getName()`.

edewata (approved):
I have some comments, but we can address that in a separate PR. Feel free to merge the PR as is.

We can certainly fix the constant names, but to avoid breaking existing code we should deprecate them first, then drop the old constants in a future release.

fmarco76:
@edewata I have just read in RHCS-4883 about other OIDs which are not printed. I'll incorporate these other OID and reorganise this code for better maintainability. Moving this PR to draft for the moment. Thanks

fmarco76:
@edewata I have reorganised the code and going deeply I found the real reason some of the extension were not shown. There was a static code which was never invoked when used with `PrettyPrintCert` because the extension class was never loaded. The reference to the class were resolved at compile time since they are `static final`. I have moved the loading code to a different place and it works with `PrettyPrintCert` and it should work with PKI since OIDMap is always loaded to access the maps.

There is a problem with postgresql but it is also in other PR. We can investigate in a separate PR.

base/src/main/java/org/mozilla/jss/netscape/security/x509/OIDMap.java
```diff
@@ -131,43 +143,28 @@ private static void loadNamesDefault(Properties props) {
         props.put(CERT_POLICIES, "2.5.29.32");
         props.put(AUTH_KEY_IDENTIFIER, "2.5.29.35");
         props.put(SUBJ_DIR_ATTR, "2.5.29.9");
-        props.put(EXT_KEY_USAGE, "2.5.29.37");
```
edewata, line 134: Just to confirm, so the `EXT_KEY_USAGE` doesn't need to be specified here anymore?

edewata (approved):
Thanks for the update! I have some more comments but the changes look good. Feel free to update/merge.

base/src/main/java/org/mozilla/jss/netscape/security/x509/OIDMap.java
fmarco76: The `ExtendedKeyUsageExtension` class was adding itself [here](https://github.com/dogtagpki/jss/blob/66bd995f79a5a05c78823821535f4fb4908eaea1/base/src/main/java/org/mozilla/jss/netscape/security/extensions/ExtendedKeyUsageExtension.java#L65-L71). I have not investigated there reason but it is like adding this extension whatever is the map. However, this code was not executed so I moved in the static block of the OIDMap with the `addClass()` method.

base/src/main/java/org/mozilla/jss/netscape/security/x509/OIDMap.java
```diff
@@ -131,43 +143,28 @@ private static void loadNamesDefault(Properties props) {
...
-        props.put(NAME_CONSTRAINTS,
-                  "org.mozilla.jss.netscape.security.x509.NameConstraintsExtension");
-        props.put(POLICY_CONSTRAINTS,
-                  "org.mozilla.jss.netscape.security.x509.PolicyConstraintsExtension");
-        props.put(CERT_POLICIES,
-                  "org.mozilla.jss.netscape.security.x509.CertificatePoliciesExtension");
-        props.put(SUBJ_DIR_ATTR,
-                  "org.mozilla.jss.netscape.security.x509.SubjectDirAttributesExtension");
-        props.put(EXT_KEY_USAGE,
-                  "org.mozilla.jss.netscape.security.extensions.ExtendedKeyUsageExtension");
```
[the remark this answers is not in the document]
fmarco76: Same anwer.

fmarco76:
@edewata Thanks! I have just remove the useless `toString()` before merge.

PR #1002, opened 2024-04-04
Update exception handling in PFX.verifyAuthSafes()

edewata:
Previously the `PFX.verifyAuthSafes()` discarded the original exception and simply returned a generic error message which was usually insufficient to determine the cause of the issue.

To help troubleshooting, the `PFX.verifyAuthSafes()` has been modified to throw the exception so the original error message and the stack trace are preserved.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/pkcs12/PFX.java
+++ b/base/src/main/java/org/mozilla/jss/pkcs12/PFX.java
@@ -128,52 +128,38 @@ public class PFX implements ASN1Value {
      *      this PFX does not contain a MacData, returns false.
      */
     public boolean verifyAuthSafes(Password password, StringBuffer reason)
-            throws NotInitializedException {
+            throws Exception {
 
-        try {
-            if (reason == null) {
-                // this is just so we don't get a null pointer exception
-                reason = new StringBuffer();
-            }
+        if (reason == null) {
+            // this is just so we don't get a null pointer exception
+            reason = new StringBuffer();
+        }
 
-            if (macData == null) {
-                reason.append("No MAC present in PFX");
-                return false;
-            }
-
-            if (encodedAuthSafes == null) {
-                // We weren't decoded from a template, we were constructed,
-                // so just verify the encoding of the AuthSafes provided to
-                // the constructor.
-                encodedAuthSafes = ASN1Util.encode(authSafes);
-            }
-
-            // create a new MacData based on the encoded Auth Safes
-            DigestInfo macDataMac = macData.getMac();
-            MacData testMac = new MacData(password,
-                    macData.getMacSalt().toByteArray(),
-                    macData.getMacIterationCount().intValue(),
-                    encodedAuthSafes,
-                    macData.getMac().getDigestAlgorithm());
-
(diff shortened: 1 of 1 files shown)


base/src/main/java/org/mozilla/jss/pkcs12/PFX.java
```diff
@@ -164,17 +164,8 @@ public boolean verifyAuthSafes(Password password, StringBuffer reason)
...
-
-        } catch (TokenException e) {
-            reason.append("A TokenException occurred");
-            return false;
-
-        } catch (CharConversionException e) {
-            reason.append("An exception occurred converting the password from chars to bytes");
-            return false;
+        } catch (Exception e) {
+            throw e;
```
fmarco76, line 168: Since we are not doing any action with the exception then try/catch could be totally removed.

fmarco76 (commented):
This code is invoked by the CLI so if there is an error the user get the stacktrace. I think the original approach with a reason message was better. Could we just add the exception message to the reason to make clearer what is the problem? Is it possible to add more information when the CLI run in debug mode?

base/src/main/java/org/mozilla/jss/pkcs12/PFX.java
edewata: Updated, the try-catch is removed now.

edewata:
@fmarco76 Please see the updated patch. I removed the try-catch statement as you suggested. As for the stack trace, I think it's better to let the application (i.e. CLI or UI) handle the exception. Currently by default the CLI will only show the exception message without the stack trace, for example:
```
$ pki pkcs12-cert-find --pkcs12 certs.p12 --password wrong
ERROR: Unable to validate PKCS #12 file: Digests do not match
```
However, for troubleshooting we can run the CLI in verbose mode to see the stack trace:
```
$ pki pkcs12-cert-find --pkcs12 certs.p12 --password wrong -v
INFO: Initializing NSS
INFO: Using internal token
java.lang.Exception: Unable to validate PKCS #12 file: Digests do not match
	at org.mozilla.jss.netscape.security.pkcs.PKCS12Util.loadFromByteArray(PKCS12Util.java:830)
	at org.mozilla.jss.netscape.security.pkcs.PKCS12Util.loadFromFile(PKCS12Util.java:816)
	at com.netscape.cmstools.pkcs12.PKCS12CertFindCLI.execute(PKCS12CertFindCLI.java:128)
	at org.dogtagpki.cli.CommandCLI.execute(CommandCLI.java:58)
	at org.dogtagpki.cli.CLI.execute(CLI.java:353)
	at org.dogtagpki.cli.CLI.execute(CLI.java:353)
	at org.dogtagpki.cli.CLI.execute(CLI.java:353)
	at com.netscape.cmstools.cli.MainCLI.execute(MainCLI.java:659)
	at com.netscape.cmstools.cli.MainCLI.main(MainCLI.java:698)
```
The original code removed the exception info completely so it's not possible to see the stack trace at all, and the original error messages were too generic to be useful.

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #1003, opened 2024-04-10
Enabling certificate verification using CRL-DP extension

fmarco76:
Currently certificates can be validated only using OCSP with a configured responder or using the AIA certificate extension. If the responder cannot be used verification is not possible. This is the case for the startup certificates of the responder.

The new policy add a verification using the CRL-DP extension defined in the certificate. If this extension is defined it has precedence over the OCSP and if it pass no other check are performed. If CRL cannot be retrieved then the OCSP responder is used. This new method takes place when OCSP is configured without a default responder and the PKIX verification method is adopted (with the policy OCSP_LEAF_AND_CHAIN_POLICY).

At least a verification method has to return success to accept the certificate.

Additionally, the connector parameter `enableOCSP` is renamed to `enableRevocationCheck` since this is used to enable CRL-DP and AIA based validation

### DIFF
--- a/docs/changes/v5.6.0/Parameter-Changes.adoc
+++ b/docs/changes/v5.6.0/Parameter-Changes.adoc
@@ -1,0 +1,5 @@
+= Paramater Changes =
+
+== Rename enableOCSP to enableRevocationCheck ==
+
+The `enableOCSP` param has been deprecated. Use `enableRevocationCheck` instead.
--- a/native/src/main/native/org/mozilla/jss/ssl/common.c
+++ b/native/src/main/native/org/mozilla/jss/ssl/common.c
@@ -936,16 +936,19 @@ JSSL_verifyCertPKIXInternal(CERTCertificate *cert,
     CERTVerifyLog *log, SECCertificateUsage *usage,
     CERTCertList *trustedCertList)
 {
-    /* Put the first set of possible flags internally here first. Later
-     * there could be a more complete list to choose from; for now we only
-     * support our hard core fetch AIA OCSP policy. Note that we disable
-     * CRL fetching as Dogtag doesn't support it. Additionally, enable OCSP
-     * checking on the chained CA certificates. Since NSS/PKIX's
-     * CERT_GetClassicOCSPEnabledHardFailurePolicy doesn't do what we want,
-     * we construct the policy ourselves. */
+    /* Since NSS/PKIX's CERT_GetClassicOCSPEnabledHardFailurePolicy doesn't
+     * do what we want, we construct the policy ourselves.
+     *
+     * When enabled the checking on the chained CA certificates.
+     * With this policy the verification process does:
+     * - if one between AIA and CRL-DP is present then it will be used;
+     * - if AIA and CRL-DP are both presents only AIA is used and in case
+     *   freshin formation cannot be retrieved it fails the validation;
+     * - it no AIA and CRL-DP are present no revocation check is performed.*/
     PRUint64 ocsp_Enabled_Hard_Policy_LeafFlags[2] = {
         /* crl */
-        CERT_REV_M_DO_NOT_TEST_USING_THIS_METHOD,
+        CERT_REV_M_TEST_USING_THIS_METHOD |
+            CERT_REV_M_FAIL_ON_MISSING_FRESH_INFO,
         /* ocsp */
         CERT_REV_M_TEST_USING_THIS_METHOD |
             CERT_REV_M_FAIL_ON_MISSING_FRESH_INFO
@@ -953,7 +956,8 @@ JSSL_verifyCertPKIXInternal(CERTCertificate *cert,
 
     PRUint64 ocsp_Enabled_Hard_Policy_ChainFlags[2] = {
         /* crl */
-        CERT_REV_M_DO_NOT_TEST_USING_THIS_METHOD,
+        CERT_REV_M_TEST_USING_THIS_METHOD |
+            CERT_REV_M_FAIL_ON_MISSING_FRESH_INFO,
         /* ocsp */
         CERT_REV_M_TEST_USING_THIS_METHOD |
             CERT_REV_M_FAIL_ON_MISSING_FRESH_INFO
--- a/native/src/main/native/org/mozilla/jss/ssl/common.c
+++ b/native/src/main/native/org/mozilla/jss/ssl/common.c
@@ -929,6 +929,21 @@ finish:
     return root; 
 }
 
+
+/* Verify if the CRLDP extension is defined in the certificate. */
+static PRBool JSSL_isCRLDPExtensionInCert(CERTCertificate *cert)
+{
+    SECStatus rv = CERT_FindCertExtension(cert,
+                                          SEC_OID_X509_CRL_DIST_POINTS,
+                                          NULL);
+    if (rv == SECSuccess) {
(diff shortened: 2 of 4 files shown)


ladycfu (commented):
looks good. my only comment is,, as stated in slack, to keep enableOCSP working as before just for backward compatibility purpose.  Thanks

tomcat/src/main/java/org/dogtagpki/jss/tomcat/Http11NioProtocol.java
```diff
@@ -73,12 +73,12 @@ public void setServerCertNickFile(String serverCertNickFile) {
...
-    public boolean getEnabledOCSP() {
-        return tomcatjss.getEnableOCSP();
+    public boolean getEnableRevocationCheck() {
+        return tomcatjss.getEnableRevocationCheck();
     }
 
-    public void setEnableOCSP(boolean enableOCSP) {
-        tomcatjss.setEnableOCSP(enableOCSP);
+    public void setEnableRevocationCheck(boolean enableRevocationCheck) {
+        tomcatjss.setEnableRevocationCheck(enableRevocationCheck);
```
edewata, line 81: As discussed on Slack, we should keep the `enabledOCSP` param for backward compatibility, but now the getter/setter for that param can call `TomcatJSS.enableRevocationCheck`'s getter/setter instead.

tomcat/src/main/java/org/dogtagpki/jss/tomcat/TomcatJSS.java
```diff
@@ -267,9 +267,9 @@ public void loadJSSConfig(Properties config) {
         if (passwordFileProp != null)
             setPasswordFile(passwordFileProp);
 
-        String enableOCSPProp = config.getProperty("enableOCSP");
-        if (enableOCSPProp != null)
-            setEnableOCSP(Boolean.parseBoolean(enableOCSPProp));
+        String enableRevocationCheckProp = config.getProperty("enableRevocationCheck");
+        if (enableRevocationCheckProp != null)
+            setEnableRevocationCheck(Boolean.parseBoolean(enableRevocationCheckProp));
```
edewata, line 272: Similarly, we should also check `enableOCSP` param for backward compatibility.

tomcat/src/main/java/org/dogtagpki/jss/tomcat/TomcatJSS.java
```diff
@@ -343,9 +343,9 @@ public void loadTomcatConfig(Document document) throws XPathExpressionException
         if (serverCertNickFileProp != null)
             setServerCertNickFile(serverCertNickFileProp);
 
-        String enableOCSPProp = connector.getAttribute("enableOCSP");
-        if (enableOCSPProp != null)
-            setEnableOCSP(Boolean.parseBoolean(enableOCSPProp));
+        String enableRevocationCheckProp = connector.getAttribute("enableRevocationCheck");
+        if (enableRevocationCheckProp != null)
+            setEnableRevocationCheck(Boolean.parseBoolean(enableRevocationCheckProp));
```
edewata, line 348: Similarly, we should also check `enableOCSP` param for backward compatibility.

docs/changes/v5.6.0/Parameter-Changes.adoc
```diff
@@ -0,0 +1,7 @@
+= Paramater Changes =
+
+== enableOCSP become enableRevocationCheck ==
+
+Since verification with CRL-DP is introduced and it is enabled using
+this parameter in `server.xml` connector, the name is modified to be not related only to OCSP.
```
edewata, line 6: Since we're going to keep `enableOCSP`, I'd suggest to modify this to be something like:

Rename enableOCSP to enableRevocationCheck

The enableOCSP param has been deprecated. Use enableRevocationCheck instead.

edewata (approved):
I have some comments about backward compatibility, but everything else looks good. Feel free to update/merge.

As discussed in Slack, in the future we might want to add more params to configure the validation policy (e.g. OCSP first then CRL).

native/src/main/native/org/mozilla/jss/ssl/common.c
```diff
@@ -936,24 +936,28 @@ JSSL_verifyCertPKIXInternal(CERTCertificate *cert,
...
-     * CRL fetching as Dogtag doesn't support it. Additionally, enable OCSP
-     * checking on the chained CA certificates. Since NSS/PKIX's
-     * CERT_GetClassicOCSPEnabledHardFailurePolicy doesn't do what we want,
-     * we construct the policy ourselves. */
+    /* Since NSS/PKIX's CERT_GetClassicOCSPEnabledHardFailurePolicy doesn't
+     * do what we want, we construct the policy ourselves.
+     *
+     * When enabled the checking on the chained CA certificates.
+     * With this policy the verification process does:
+     * - if one between AIA and CRL-DP is present then it will be used;
```
ladycfu, line 944: suggestion to rephrase  : "- if only an AIA or an CRL-DP is present, the respective validation method is used;"

native/src/main/native/org/mozilla/jss/ssl/common.c
```diff
@@ -929,36 +929,62 @@ CERTCertificate *getRoot(CERTCertificate *cert,
...
 
     PRUint64 ocsp_Enabled_Hard_Policy_ChainFlags[2] = {
         /* crl */
-        CERT_REV_M_DO_NOT_TEST_USING_THIS_METHOD,
+        CERT_REV_M_TEST_USING_THIS_METHOD |
+            CERT_REV_M_FAIL_ON_MISSING_FRESH_INFO,
         /* ocsp */
         CERT_REV_M_TEST_USING_THIS_METHOD |
             CERT_REV_M_FAIL_ON_MISSING_FRESH_INFO
     };
```
ladycfu, line 980: for the following setting, it might help the reader to add a comment that says something like "/* if CRL-dp is present in the cert, disable CERT_REV_M_FAIL_ON_MISSING_FRESH_INFO for ocsp */

native/src/main/native/org/mozilla/jss/ssl/common.c
```diff
@@ -929,36 +929,62 @@ CERTCertificate *getRoot(CERTCertificate *cert,
...
-     * we construct the policy ourselves. */
+    /* Since NSS/PKIX's CERT_GetClassicOCSPEnabledHardFailurePolicy doesn't
+     * do what we want, we construct the policy ourselves.
+     *
+     * When enabled the checking on the chained CA certificates.
+     * With this policy the verification process does:
+     * - if only an AIA or an CRL-DP is present, the respective validation method is used;
+     * - if AIA and CRL-DP are both presents the execution order is: AIA first followed by
+     *   CRL only in case OCSP endpoint does not provide fresh information;
+     * - it no AIA and CRL-DP are present no revocation check is performed.*/
```
ladycfu, line 962: typo. "if" instead of "it"

ladycfu (commented):
looks good.  Just a couple comments-related suggestions.

edewata:
I don't have any additional comments. Thanks for the updates!

native/src/main/native/org/mozilla/jss/ssl/common.c
fmarco76: Comment added to the if where the flag is modified

PR #1006, opened 2024-05-15
Fix CRL check when CA miss CRLSign key usage

fmarco76:
When CRLSign key usage is not present NSS crl check will consider the CRL source not present and skip the test. This is against the policy we want to implement.

This change force the need for the information source and the certificate will be marked as revoked if CRLSign is not present.


rjrelyea:
Just verifying, this patch looks good to me!,
bob

ladycfu (approved):
LGTM if tested to work.

fmarco76:
@ladycfu @rjrelyea Thanks for the review. I am merging these changes

PR #1007, opened 2024-05-17
Fix EKU decoding

fmarco76:
OIDMap was not correctly mapping EKU when decoding a certificate to an X509CertImpl object because the name should include the full path from the Root element.

This partially revert the pr #997  where OIDs were reorganised and resolve the error with third party when the certificate verification is not totally delegated to NSS code.


edewata (approved):
Sorry I missed this one. LGTM.

fmarco76:
@edewata Thanks!

PR #1008, opened 2024-05-21
Include certificate information in SSL session

fmarco76:
Certificates are included in the SSL session also in case of handshake failure. If certificate are not available there are no exception and or error reported beside the one creating the failure.

Certificate information are needed in case of event audits.


ladycfu (approved):
LGTM

fmarco76:
@ladycfu Thanks!

PR #1011, opened 2024-06-25
Fix test 17 error on build

### LINKED ISSUE #1010: Build fails on Fedora
https://copr.fedorainfracloud.org/coprs/g/pki/master/build/7662117/

Build fails because one of the unit tests is failing:

```
#15 62.58 17/77 Test #17: JSS_Test_Raw_SSL ..................................***Failed    0.10 sec
#15 62.58 Calling TestSSLImportFD()...
#15 62.58 Exception in thread "main" java.lang.NullPointerException: SSL_ImportFD failed:  (-8038)
#15 62.58 	at org.mozilla.jss.nss.SSL.ImportFD(SSL.java:169)
#15 62.58 	at org.mozilla.jss.tests.TestRawSSL.TestSSLImportFD(TestRawSSL.java:17)
#15 62.58 	at org.mozilla.jss.tests.TestRawSSL.main(TestRawSSL.java:165)
```

fmarco76:
With the nss update to >= 3.101 the nss context gets initialised after the CryptoManager instance is retrieved so this step is added before the test.

The original error was in this NSS line: https://github.com/nss-dev/nss/blob/4b0cca5ba45b3713e279f9d528dfa01dd886cea8/lib/nss/nssinit.c#L1002

All the other tests were working because the use of the CryptoManager before performing and operation involving NSS.

Additionally, the ssl option test for get and set has been modified to enable/disable ssl cache since the ssl3 option is policy dependent and do not always work. 
Fix #1010 

Note about the options: running the test with NSS released on fedora using SSL3 option was not working but using my build (I have included debug options) the test works. Looking at the code SSL3 cannot be enabled if it is not supported by the policy: https://github.com/nss-dev/nss/blob/4b0cca5ba45b3713e279f9d528dfa01dd886cea8/lib/ssl/sslsock.c#L629. Therefore I have just used a different option since the test was just to verify that option can be configured and it was not related to ssl3. We can investigate more if needed.

### DIFF
--- a/base/src/test/java/org/mozilla/jss/tests/TestRawSSL.java
+++ b/base/src/test/java/org/mozilla/jss/tests/TestRawSSL.java
@@ -1,5 +1,6 @@
 package org.mozilla.jss.tests;
 
+import org.mozilla.jss.CryptoManager;
 import org.mozilla.jss.nss.PR;
 import org.mozilla.jss.nss.PRFDProxy;
 import org.mozilla.jss.nss.SSL;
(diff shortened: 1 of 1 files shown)


edewata (approved):
LGTM. Thanks! This is a test for SSL options so this should be sufficient. I think SSL3 would be a separate issue.

fmarco76:
@edewata Thanks!

PR #1017, opened 2024-08-06
Update JSSTrustManager to support trusted peers

edewata:
`JSSTrustManager` has been updated to mimic NSS cert validation which supports trusted peers. The `checkCertChain()` has been modified to check whether the cert chain has `P,,` trust flags, and if that's the case the cert chain is considered trusted so it's not necessary to check the cert issuer anymore.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #1019, opened 2024-08-08
Add cert approval callback for JSSTrustManager

edewata:
Previously the `JSSTrustManager` would throw an exception to reject a cert if it found an issue in the cert. The code has been changed to collect all issues, then pass them into an optional callback object. If the callback approves it, the cert will be considered trusted. Otherwise, or if there is no callback, the code will throw an exception based on the first issue.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
+++ b/base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
@@ -20,18 +20,25 @@
 package org.mozilla.jss.provider.javax.crypto;
 
 import java.security.cert.CertificateException;
+import java.security.cert.CertificateExpiredException;
+import java.security.cert.CertificateNotYetValidException;
 import java.security.cert.X509Certificate;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Enumeration;
 import java.util.List;
 
 import javax.net.ssl.X509TrustManager;
+import javax.security.auth.x500.X500Principal;
 
 import org.mozilla.jss.CryptoManager;
 import org.mozilla.jss.NotInitializedException;
 import org.mozilla.jss.netscape.security.util.Cert;
 import org.mozilla.jss.pkcs11.PK11Cert;
+import org.mozilla.jss.ssl.SSLCertificateApprovalCallback;
+import org.mozilla.jss.ssl.SSLCertificateApprovalCallback.ValidityItem;
+import org.mozilla.jss.ssl.SSLCertificateApprovalCallback.ValidityStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
```diff
@@ -175,25 +246,15 @@ public void checkKeyUsage(X509Certificate[] certChain, String keyUsage) throws E
...
-                first = false;
-            }
-            msg += " }";
+            logger.debug("JSSTrustManager: Missing " + keyUsage + " key usage");
         }
 
-        msg += ".  class = " + cert.getClass();
-        throw new CertificateException(msg);
+        // TODO: fix depth param
+        status.addReason(ValidityStatus.INADEQUATE_KEY_USAGE, cert, 0);
```
fmarco76, line 257: This check is done only for the leaf certificate so the depth is correct. The TODO can be removed.

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
```diff
@@ -131,25 +182,45 @@ public void checkSignature(X509Certificate cert, X509Certificate[] caCerts) thro
...
 
-            cert.checkValidity();
+            try {
+                cert.checkValidity();
+
+            } catch (CertificateNotYetValidException e) {
+                logger.debug("JSSTrustManager: Cert not yet valid: " + cert.getSubjectX500Principal());
+
+                // NSS uses EXPIRED_CERTIFICATE for this case in CERT_CheckCertValidTimes()
+                // TODO: fix depth param
```
fmarco76, line 211: We are following the chain from the root so depth start from `certChain.length - 1` and go to zero. It seems an easy fix, is there a reason to move this to a future PR?

fmarco76 (changes requested):
The overall changes looks good but I have a couple of comments I would understand/fix before merge.

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
edewata: According to the doc the order of the certs in the original chain is not defined, and the chain could even be incomplete:
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/net/ssl/X509TrustManager.html

In the beginning of `checkCertChain()` we sort the certs, but when calling `addReason()` we probably should use the original index (which I'm not sure what we should use if the chain is incomplete).

I think ideally `addReason()` should not need to take a `depth` param since it already has the `cert`. Maybe we can fix that later, but for now let me see if I can add another method to find the cert's index in the original (unsorted) cert chain.

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
edewata: Please see my comment below.

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
fmarco76: OK, I was not considering incomplete chain. Thanks!

edewata: Actually, the `getDepth()` is defined like this:
https://github.com/dogtagpki/jss/blob/master/base/src/main/java/org/mozilla/jss/ssl/SSLCertificateApprovalCallback.java#L148-L152
so we can always calculate the `depth` as you described from the sorted chain regardless of the original chain.

fmarco76: Yes, I read this definition and thought it could be easily calculated but the definition is ambiguous because it is not clear if refer to the provided chain or the real chain

PR #1020, opened 2024-08-09
Add hostname validation for JSSTrustManager

edewata:
The `JSSTrustManager` has been modified to take an optional `hostname` attribute. If provided, it will be used to validate the SAN extension or the cert subject DN.


fmarco76 (approved):
LGTM. The hostname checks look good but it is not working in case of wildcard DNS name in the SAN. Should they be supported? In case, they could be managed in separate PR, just adding a TODO/note

edewata:
Good point. I've added TODO comments. Thanks!

PR #1021, opened 2024-08-19
Fix Red Hat Certificate SystemRHCS-5261:  Fix JSS self tests to not use DSA and generate < 2048 bit RSA keys

jmagne:
Fix JSS self tests to not use DSA and generate < 2048 bit RSA keys

Remove / modify JSS self test code that previously generated DSA key pairs and RSA key pairs of < 2048 bit sizes. Also remove tests generating RSA key pairs with smaller than the minimum exponent sizes.


edewata (approved):
LGTM

PR #1022, opened 2024-08-20
Workaround for SSL events in JSSEngine

edewata:
Currently `JSSEngine` does not generate SSL events although it seems to receive the alerts from NSS.

As a temporary workaround, the `SSLFDProxy` has been updated to keep a list of SSL socket listeners, then `JSSEngine` will add a listener into it. When NSS generates an SSL event, it will be passed to `SSLFDProxy` listeners directly, then eventually to `JSSEngine` listeners as well.

The code that creates the SSL event in `JSS_NSS_addSSLAlert()` has been moved into `JSS_NSS_createSSLAlert()` so it can be reused.

**Note:** This PR is needed by https://github.com/dogtagpki/pki/pull/4832

### DIFF
--- a/base/src/main/java/org/mozilla/jss/nss/SSLFDProxy.java
+++ b/base/src/main/java/org/mozilla/jss/nss/SSLFDProxy.java
@@ -1,21 +1,29 @@
 package org.mozilla.jss.nss;
 
 import java.util.ArrayList;
+import java.util.List;
 
 import org.mozilla.jss.crypto.X509Certificate;
 import org.mozilla.jss.pkcs11.PK11Cert;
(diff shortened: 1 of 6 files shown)


edewata:
@fmarco76 Thanks for the investigation. Please see PR #1023. I've updated the test to include self-signed user cert. It shows some inconsistencies between the two socket factories.

edewata:
@fmarco76 Please see PR #1024. I've moved some of the changes here into that PR to make it easier to investigate SSL alert issues later.

edewata:
The missing alerts are addressed in PR #1029. Closing this PR.

PR #1023, opened 2024-08-27
Add test for self-signed user cert validation

edewata:
The PKI CA test has been updated to run PKI CLI using blocking and non-blocking socket factories to perform cert validations for variety of cases including self-signed user cert.

Currently there are inconsistencies in the SSL alerts generated by these factories. They will be investigated separately later.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #1024, opened 2024-08-28
Update SSLFDProxy to implement SSLSocketListener

edewata:
Previously `SSLFDProxy.c` was accessing the fields in `SSLFDProxy` class directly using JNI to update `handshakeComplete` and add SSL alert events into `inboundAlerts` and `outboundAlerts`.

To make it easier to investigate SSL alert issues, `SSLFDProxy` has been updated to implement `SSLSocketListener` then `SSLFDProxy.c` will call `SSLSocketListener` methods to perform the above operations.


fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

siteshwar:
You may also want to remove `JSS_NSS_getEventArrayList()` function with this change as it is not used anymore. It was found through https://openscanhub.fedoraproject.org/task/7775/log/added.html.

fmarco76:
@siteshwar Thanks for the notification. I have opened a PR to remove the method.

Issue #994, opened 2024-02-08
TPS test failed on Rawhide


edewata:
The TPS test in JSS failed consistently on Rawhide:

https://github.com/edewata/jss/actions/runs/7821129758/job/21337674464#step:11:558

```
INFO: PKISocketFactory: Creating socket for pki.example.com:389
com.netscape.certsrv.base.PKIException: LDAP error (1): Operations error
	at com.netscape.certsrv.ldap.LDAPExceptionConverter.toPKIException(LDAPExceptionConverter.java:47)
	at com.netscape.cmscore.usrgrp.UGSubsystem.modifyGroup(UGSubsystem.java:1782)
	at org.dogtagpki.server.cli.SubsystemGroupMemberAddCLI.execute(SubsystemGroupMemberAddCLI.java:76)
	at org.dogtagpki.cli.CommandCLI.execute(CommandCLI.java:58)
	at org.dogtagpki.cli.CLI.execute(CLI.java:353)
	at org.dogtagpki.cli.CLI.execute(CLI.java:353)
	at org.dogtagpki.cli.CLI.execute(CLI.java:353)
	at org.dogtagpki.cli.CLI.execute(CLI.java:353)
	at org.dogtagpki.server.cli.PKIServerCLI.execute(PKIServerCLI.java:93)
	at org.dogtagpki.server.cli.PKIServerCLI.main(PKIServerCLI.java:123)
Caused by: netscape.ldap.LDAPException: Operations error (1)
	at netscape.ldap.LDAPConnection.checkMsg(LDAPConnection.java:4933)
	at netscape.ldap.LDAPConnection.modify(LDAPConnection.java:3201)
	at netscape.ldap.LDAPConnection.modify(LDAPConnection.java:3139)
	at netscape.ldap.LDAPConnection.modify(LDAPConnection.java:3149)
	at netscape.ldap.LDAPConnection.modify(LDAPConnection.java:3114)
	at com.netscape.cmscore.usrgrp.UGSubsystem.modifyGroup(UGSubsystem.java:1779)
	... 8 more
[...]

PR #1025, opened 2024-09-03
Fix TPS test failure

edewata:
PKI CA and TPS tests have been updated to use DS container from Quay instead of DS packages from Fedora to avoid JSS issue #994.

Resolves: https://github.com/dogtagpki/jss/issues/994


fmarco76 (approved):
LGTM

PR #1026, opened 2024-10-14
Remove the unused method JSS_NSS_getEventArrayList()

fmarco76:
The method was used to get SSLSoecketEvent but the mechanism has been modified in the following commit and this is method is not used.

Update SSLFDProxy to implement SSLSocketListener (0f352ecb0848ff549cf6bc255d102e1077bb1eaa)

The problem has been identified and reported as side effect of PR #1024 by @siteshwar


edewata (approved):
LGTM

fmarco76:
@edewata @siteshwar Thanks!

siteshwar:
If you may be interested in full list of findings from OpenScanHub, they can be seen through https://openscanhub.fedoraproject.org/task/9062/log/jss-5.6.0-0.20240903212537552344.pr1025.70.g73a48869/scan-results.html.

Alternatively, please review `osh-diff-scan` job for each pull request. For example https://github.com/dogtagpki/jss/pull/1022/checks?check_run_id=29079267242.

Thank you for looking into this!

PR #1029, opened 2025-02-12
Fix missing SSL Alert

fmarco76:
When trust manager fails to validate the certificate the generated SSL exception does not trigger any alert. To fix the problem the SSL alert are always verified during the wrap operation.


edewata (approved):
Nice, thanks! I've closed PR #1022.

fmarco76:
@edewata Thanks!

PR #1030, opened 2025-02-13
Update PKCS11Constants.java

edewata:
The `build_pkcs11_constants.py` has been updated to no longer exclude constants introduced in NSS 3.97 since newer NSS versions are available on all supported platforms.


fmarco76 (approved):
LGTM

PR #1031, opened 2025-02-13
Fix tomcat CI

fmarco76:
The TLS 1.2 is supported and enabled in the crypto policy so it is enabled in tomcat by default.


edewata (approved):
LGTM. If you have a link to the site policy or doc about this could you add it as a comment? Thanks.

fmarco76:
@edewata Thanks!

PR #1032, opened 2025-02-20
Add nss cert verify to JSSTrustManager

fmarco76:
Currently, JSSTrustManager does not verify if a certificate has been revoked using OCSP or CRLDP specified in the certificate. To include this verification an additional check is include to enable verification using NSS. The OCSP policy has to be enabled in the Crypto manager before it is used in the trust manager so when NSS verification is enabled revocation, along with other certificate conditions are verified.


fmarco76:
@edewata to test with a revoked certificate the NonBlockingSocketFactory has to be modified enabling the NSS check. The cahange and a  test in PKI will follow.

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
```diff
@@ -179,6 +191,9 @@ public void checkCertChain(X509Certificate[] certChain, String keyUsage) throws
 
         Enumeration<ValidityItem> reasons = status.getReasons();
         if (!reasons.hasMoreElements()) {
+            if (enabledNSSCertVerify) {
+                nssCertVerify(certChain, keyUsage);
```
edewata, line 195: In line 201 there's a code that will allow the user to manually accept the cert even though it's considered invalid. Should cert revocation be something that can be manually accepted as well? If so, the `nssCertVerify()` probably should be called before line 193.

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
```diff
@@ -428,4 +443,26 @@ public X509Certificate[] getAcceptedIssuers() {
...
+                }
+
+                cm.verifyCertificate((org.mozilla.jss.crypto.X509Certificate) certChain[i], false, usage);
+            }
+        } catch (NotInitializedException e) {
+            logger.error("JSSTrustManager: Unable to get CryptoManager: " + e, e);
+            throw new RuntimeException(e);
+        } catch (ObjectNotFoundException | InvalidNicknameException ex) {
+            logger.error("JSSTrustManager: nss certificate verification not done: " + ex, ex);
+        }
```
edewata, line 466: Not sure if these exceptions will ever happen, but if it does happen I think we should rethrow the exception (or let it be handled by the caller).

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
```diff
@@ -63,6 +66,15 @@ public class JSSTrustManager implements X509TrustManager {
 
     private String hostname;
     private boolean allowMissingExtendedKeyUsage = false;
+    private boolean enabledNSSCertVerify = false;
```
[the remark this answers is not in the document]
fmarco76: The TomcatJSS code configure how nss verification should works but do not enable/disable the verification. In this case the nss verification is disabled so the meaning is slightly different. Additionally, nss do not check only revoked certificates but it does other checks so the name `enableRevocationCheck` could be misleading. I'll remove nss from the name anyway.

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
fmarco76: This is problematic. Current JSS certificate verification with NSS throw an exception if the verification fails but the error code is reported only in the message and I do not like the message parsing approach to get the problem. 

We can implement a new JSS function to get the code instead of the exception but  NSS verification report the first problem with the certificate so it is possible to fail the verification before it checks if they are revoked. Not sure if this can work.

edewata: Yeah, internally JSS should implement native methods that can fully control NSS so we can get everything that NSS provides, but externally JSS should provide standard Java security API so eventually it can be used without JSS-specific code.

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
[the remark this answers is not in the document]
edewata: Or should we always enable revocation check in JSSTrustManager so it will work just like TomcatJSS? Eventually TomcatJSS probably should use JSSTrustManager or be replaced with it anyway.

fmarco76:
I have updated the code so the Revoke problem is handled like other errors and it is possible to overwrite the revocation acceptign the certificate.
There is a limitation I cannot overcome. The NSS code to verify the certificate checks the key extended usage before the revocation and if it is not correct I get the error before revocation is verified. This is something to consider for the following callback which could accept the certificate.

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
```diff
@@ -63,6 +65,15 @@ public class JSSTrustManager implements X509TrustManager {
...
     private boolean allowMissingExtendedKeyUsage = false;
+    private boolean enabledCertRevokeVerify = false;
+
+    public boolean isEnabledCertRevokeVerify() {
+        return enabledCertRevokeVerify;
+    }
+
+    public void setEnabledCertRevokeVerify(boolean enabledCertRevokeVerify) {
+        this.enabledCertRevokeVerify = enabledCertRevokeVerify;
+    }
```
edewata, line 76: Could we replace `enabled` with `enable` so it's more consistent with `allowMissingExtendedKeyUsage`?

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
```diff
@@ -428,4 +438,25 @@ public X509Certificate[] getAcceptedIssuers() {
 
         return caCerts.toArray(new X509Certificate[caCerts.size()]);
     }
+
+    private void certChainRevokeVerify(X509Certificate[] certChain, String KeyUsage, ValidityStatus status) {
+        logger.debug("JSSTrustManager: certChainRevokeVerify()");
+        
+        //NOTE: if the usage is not correct the verification stops and the revocation is not verified
+        int leafUsage = KeyUsage.equals(CLIENT_AUTH_OID) ? CertificateUsage.SSLClient.getUsage() : CertificateUsage.SSLServer.getUsage();
```
edewata, line 446: I wonder if we could pass a value that NSS will always consider valid so basically we're bypassing the key usage verification in NSS, then we implement our own key usage verification in PKI. We can deal with this later if it becomes a problem.

edewata (approved):
Thanks for the update! I have a minor suggestion, but everything looks good. Feel free to update/merge.

base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
fmarco76: I have done a lot of tests with many combination but I cannot get this passed. I have done also some investigation on NSS code but I have not found alternative APIs performing only the revocation check. We can implement this in JSS using lower NSS APIs but this can be evaluated in separate PR.

fmarco76:
@edewata Thanks!

PR #1033, opened 2025-02-24
Add test for HTTPS with TLS 1.3 in Tomcat

edewata:
The current test for HTTPS with default settings has been moved into a new GH workflow. A new test has been added for HTTPS with TLS 1.3.


edewata:
@fmarco76 Thanks!

PR #1035, opened 2025-02-25
JSSTrustManager check revokation  only if trusted issuer

fmarco76:
If the certificate is a trusted peer there is not check of the issuer and/or the certificate.

If the issuer is not available, or it is a trusted peer, the revocation is not verified because it cannot verify the full chain in any case. The issue is already reported.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
+++ b/base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
@@ -221,15 +221,28 @@ public class JSSTrustManager implements X509TrustManager {
 
         if (!isTrustedPeer(certChain)) {
             checkIssuerTrusted(certChain, status);
+
+            if (enableCertRevokeVerify) {
+                // If the certificate is a trusted peer there is not check of the issuer and/or the certificate.
+                // If the issuer is not available, or it is a trusted peer, the revocation is not verified because 
+                // it cannot verify the full chain in any case. The issue is already reported.
+                // See: https://github.com/dogtagpki/jss/blob/18e97f6fe56ac89627979614f35c84b50630dcf3/native/src/main/native/org/mozilla/jss/ssl/common.c#L1141
+                Enumeration<ValidityItem> reasons = status.getReasons();
+                boolean issuerTrusted = true;
+                while (reasons.hasMoreElements() && issuerTrusted) {
+                    ValidityItem reason = reasons.nextElement();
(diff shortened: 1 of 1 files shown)


base/src/main/java/org/mozilla/jss/provider/javax/crypto/JSSTrustManager.java
```diff
@@ -219,17 +219,14 @@ public void checkCertChain(X509Certificate[] certChain, String keyUsage, Validit
 
         checkHostname(certChain, status);
 
-        if (!isTrustedPeer(certChain)) {
-            checkIssuerTrusted(certChain, status);
+        if (!isTrustedPeer(certChain) && isIssuerTrusted(certChain, status) && enableCertRevokeVerify) {        
+            certChainRevokeVerify(certChain, keyUsage, status);
```
edewata, line 223: Generally I'd avoid creating a function with side effects (e.g. adding a reason into `status`) to make it easier to understand and reuse. This code probably could be written like this:
```
if (!isTrustedPeer(certChain)) {
    checkIssuerTrusted(certChain, status);
    if (!status.contains(UNKNOWN_ISSUER) && enableCertRevokeVerify) {
        certChainRevokeVerify(certChain, keyUsage, status);
    }
}
```
or this:
```
if (!isTrustedPeer(certChain)) {
    reason = checkIssuerTrusted(certChain);
    if (reason != null) {
        status.add(reason);
    } else if (enableCertRevokeVerify) {
        certChainRevokeVerify(certChain, keyUsage, status);
    }
}
```

edewata (approved):
Could you include this explanation as a comment in the code?

> If the certificate is a trusted peer there is not check of the issuer and/or the certificate.
> 
> If the issuer is not available, or it is a trusted peer, the revocation is not verified because it cannot verify the full chain in any case.

Is this actually an NSS issue? If you have the link to the issue please include it as well.

I also have a suggestion below, but I'll let you decide. Feel free to merge as is. Thanks!

BTW, it would be nice if we have tests for this to ensure that it's working as expected.

fmarco76:
> BTW, it would be nice if we have tests for this to ensure that it's working as expected.

This require to be enabled in PKI. I have identified the issue while I was testing the change in PKI.

fmarco76:
> If you have the link to the issue please include it as well.

https://github.com/dogtagpki/jss/blob/18e97f6fe56ac89627979614f35c84b50630dcf3/native/src/main/native/org/mozilla/jss/ssl/common.c#L1141

fmarco76:
@edewata Thanks!

PR #1037, opened 2025-02-26
Add test for reveked check in JSSTrustManager

fmarco76:
To work properly the JSSTrustManager has to enable the revocation check. This is in the PR [pki#4968](https://github.com/dogtagpki/pki/pull/4968) so this has to wait the merge.

### DIFF
--- a/.github/workflows/pki-ca-test.yml
+++ b/.github/workflows/pki-ca-test.yml
@@ -112,10 +112,10 @@ jobs:
 
       - name: Initialize PKI client
         run: |
-          docker exec pki pki-server cert-export ca_signing --cert-file ca_signing.crt
+          docker exec pki pki-server cert-export ca_signing --cert-file $SHARED/ca_signing.crt
 
           docker exec pki pki nss-cert-import \
-              --cert ca_signing.crt \
+              --cert $SHARED/ca_signing.crt \
               --trust CT,C,C \
               ca_signing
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pki-ca-test.yml
```diff
@@ -527,6 +527,165 @@ jobs:
...
+              -U https://pki.example.com:8443 \
+              info \
+              > >(tee stdout) 2> >(tee stderr >&2) || true
+
+          # check stdout
+          cat > expected << EOF
+            Server URL: https://pki.example.com:8443
+            Server Name: Dogtag Certificate System
+            Server Version: 11.7.0
+          EOF
```
edewata, line 600: I suppose it should have failed. If this is a known bug could you add a link to the issue? Or if it requires further investigation please add a TODO.

edewata (approved):
LGTM, but could you add a brief description about the expected result in the validation step (e.g. operation should work, cert should be rejected)? Feel free to merge. Thanks!

.github/workflows/pki-ca-test.yml
fmarco76: Not really. [Default OCSP policy in CryptoManager](https://github.com/dogtagpki/jss/blob/31699ded207d4c3798aaeb4a05ba9a0cf236b240/base/src/main/java/org/mozilla/jss/CryptoManager.java#L1399) is to not check certificate verification. In tomcatJSS there is the `revocationCheck` (formerly `enableOCSP`) to enable the check. I do not see it enabled  in the DefaultSocketFactory  so it works as expected.

fmarco76:
@edewata Thanks!

.github/workflows/pki-ca-test.yml
edewata: Hmm.. but this is on the client side. Maybe we should always enable revocation check on the client side so the behavior is consistent?

fmarco76: We are migrating to JSSSocket and the JSSTrustManager where OCSP check can be enabled independently of CryptoManager configuration. PKI client code can be modified to enable OCSP check but is there a reason to modify the current implementation? My understanding was to replace SSLSocket.

PR #1038, opened 2025-03-04
Add test for JSSTrutManager ignoring revocation

fmarco76:


edewata (approved):
As discussed below, here we're still calling the OCSP responder but we ignore the result:
https://github.com/dogtagpki/pki/pull/4976#discussion_r1979945181

I'm not sure if in practice that would be different from disabling revocation checking completely (i.e. not calling OCSP responder at all), but that can be discussed separately later. Feel free to merge. Thanks!

fmarco76:
@edewata Thanks! I am merging this but the discussion  on the opportunity to add an additional option to skip the checks can continue.

PR #1042, opened 2025-03-17
Update CA test to ignore PKI version number

edewata:
The CA test has been updated to no longer expect a specific PKI version number to make it more reliable.


edewata:
@fmarco76 Thanks!

PR #1043, opened 2025-03-20
Fix write problem with blocking channel

fmarco76:
JSSSocket can be used with blocking and non blocking channel. When used with blocking channel the write has to send all data in a single request but if the sending buffer is full only that is sent and the remaining data will be send in following write operations.

This commit fix the write adding consecutive operations if the buffer is full in order to send all application data.

As en example of the problem, if JSSSocket is used in the pki CLI, which uses blocking channels, some command with large input send only a partial amount of data and the server will timeout waiting for the remaining (this is visible in [this test](https://github.com/fmarco76/pki/actions/runs/13791116504/job/38571769253)).


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #1044, opened 2025-03-21
Make JSSSocket close() idempotent

fmarco76:
Invoking multiple times the close() method generate an exception because it calls the init() which tries to access closed channels.

The close() method is defined in Closeable interface and it has to be idempotent so it should be possible to call again without errors.

This is a problem with third party libraries not checking if the socket has been closed but call multiple times the close() method generating error.

An example of the error is in [this test](https://github.com/fmarco76/pki/actions/runs/13988696953/job/391679594390. The socket is closed by the server and and the client gets a notify which will close also the client side. Then the httpclient will close again the socket generating the error.


edewata (approved):
LGTM

edewata:
BTW, for trivial or straight forward changes feel free to skip the review process so you are not blocked.

fmarco76:
@edewata Thanks!

PR #1047, opened 2025-06-27
Add EC public key generation from spec

fmarco76:
Public key generation has been updated to allow the creation of ECPublicKey from the ECPublicKeySpec.
The code was requiring the encoded public key to generate a new key to use for signing verification but this is a limitation when the encoded key is not available, as for JWS validation using ES256 algorithm.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/provider/java/security/KeyFactorySpi1_2.java
+++ b/base/src/main/java/org/mozilla/jss/provider/java/security/KeyFactorySpi1_2.java
@@ -6,21 +6,25 @@ package org.mozilla.jss.provider.java.security;
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.math.BigInteger;
+import java.security.AlgorithmParameters;
 import java.security.InvalidKeyException;
 import java.security.Key;
 import java.security.NoSuchAlgorithmException;
 import java.security.PublicKey;
 import java.security.spec.DSAPrivateKeySpec;
 import java.security.spec.DSAPublicKeySpec;
+import java.security.spec.ECPublicKeySpec;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.KeySpec;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.RSAPrivateCrtKeySpec;
 import java.security.spec.RSAPublicKeySpec;
 import java.security.spec.X509EncodedKeySpec;
+import java.util.Arrays;
 
 import org.mozilla.jss.asn1.ASN1Util;
 import org.mozilla.jss.asn1.BIT_STRING;
+import org.mozilla.jss.asn1.CHOICE;
 import org.mozilla.jss.asn1.INTEGER;
 import org.mozilla.jss.asn1.OBJECT_IDENTIFIER;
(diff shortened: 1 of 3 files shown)


ladycfu (approved):
@fmarco76 Please note that I have employed Copilot to assist in the review of this PR.  Personally, I do not see anything really wrong other than perhaps adding more comments (which Copilot made some good suggestions too).  I am approving this PR, but still hope that you pick and choose whatever recommendations that you see fit.

base/src/main/java/org/mozilla/jss/util/ECCurve.java
```diff
@@ -170,4 +170,13 @@ public static ECCurve fromName(String name) {
 
         return null;
     }
+
```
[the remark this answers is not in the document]
jmagne: Since the list is of size 3, I'm not sure we would be saving that much. It would also depend on how often this gets called.

base/src/main/java/org/mozilla/jss/provider/java/security/KeyFactorySpi1_2.java
```diff
@@ -77,45 +83,46 @@ protected PublicKey engineGeneratePublic(KeySpec keySpec)
...
+            }
+			
+            AlgorithmIdentifier algID =
+                new AlgorithmIdentifier(oid, curve.getOIDs()[0]);
+
+            byte[] publicValueX = convertIntToBigEndian(ecPKSpec.getW().getAffineX(), pointSize);
+            byte[] publicValueY = convertIntToBigEndian(ecPKSpec.getW().getAffineY(), pointSize);
+
+            byte[] encodedPublicValue = new byte[1 + publicValueX.length + publicValueY.length];
+            encodedPublicValue[0] = 0x04; //EC_UNCOMPRESSED_POINT
```
jmagne, line 114: I agree that this block could be improved with ArrayCopy since we seem to be concatenating two byte arrays together with the tag byte at the front.

base/src/main/java/org/mozilla/jss/provider/java/security/KeyFactorySpi1_2.java
```diff
@@ -253,4 +260,21 @@ protected Key engineTranslateKey(Key key)
...
+
+    private byte[] convertIntToBigEndian(BigInteger elem, int size) {
+        byte[] arrayElem = elem.toByteArray();
+        if(arrayElem.length > size) {
+            arrayElem = Arrays.copyOfRange(arrayElem, arrayElem.length - size, arrayElem.length);
+        }
+        if(arrayElem.length < size) {
+            byte[] temp = new byte[size];
+            Arrays.fill(temp, (byte) 0);
+            for (int i=0; i < arrayElem.length; i++) {
```
jmagne, line 273: Agree with the ArrayCopy here.

jmagne (approved):
Looks good. I think the main savers would be the ArrayCopy's instead of copying one by one. The input validation suggestions seems fine, but I think the hashmap of 3 could be over kill.  Approving.

edewata:
Per discussion with @ladycfu I'll merge this PR as is now, and the suggested improvements can be done separately later. Thanks everyone!

PR #1049, opened 2025-07-07
Cleanup EC public key generation

fmarco76:
EC key size has been moved into the ECCurve Enum. The key byte arrays are handled with System.arraycopy instead of custom for loops. New private method to convert BigInteger to byte array has been documented.


fmarco76:
@ladycfu Thanks!

PR #1051, opened 2025-07-07
Update version number to 5.9.0-alpha1

edewata:
The build scripts have been modified to generate RPM version numbers more compliant with Fedora Packaging Guidelines.

https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning

### DIFF
--- a/jss.spec
+++ b/jss.spec
@@ -2,33 +2,41 @@
 Name:           jss
 ################################################################################
 
+Summary:        Java Security Services (JSS)
+URL:            https://github.com/dogtagpki/jss
+License:        (MPL-1.1 OR GPL-2.0-or-later OR LGPL-2.1-or-later) AND Apache-2.0
+
 %global         vendor_id dogtag
 %global         product_id %{vendor_id}-jss
 
 # Upstream version number:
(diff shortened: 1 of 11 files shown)


jss.spec
```diff
@@ -2,33 +2,41 @@
...
+
+# RPM version number:
+# - development:   <major>.<minor>.<update>~<phase>^<timestamp>.<commit_id>
+# - stabilization: <major>.<minor>.<update>~<phase>
+# - GA/update:     <major>.<minor>.<update>
+#
+# https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning
+
+Version:        %{major_version}.%{minor_version}.%{update_version}%{?phase:~}%{?phase}%{?timestamp:^}%{?timestamp}%{?commit_id:.}%{?commit_id}
+Release:        %autorelease
```
fmarco76, line 39: Using the macro all the COPR builds from master will increment the release number until we modify the version so it should be possible to match the package name with the commit. Is this correct?

edewata:
@fmarco76 Thanks! I also removed the redundant `--version` and `--release` options.

PR #1053, opened 2025-07-18
Adding option for ENABLE_NSS_VERSION_PQC_DEF

ladycfu:
Note that we still need jss.spec to differentiate NSS versions to decide whether to set ENABLE_NSS_VERSION_PQC_DEF to on or off.

Fix as part of the build task:
https://issues.redhat.com/browse/IDM-2360

### DIFF
--- a/cmake/JSSConfig.cmake
+++ b/cmake/JSSConfig.cmake
@@ -140,6 +140,10 @@ macro(jss_config_cflags)
         list(APPEND JSS_RAW_C_FLAGS "-O2")
     endif()
 
+    if(ENABLE_NSS_VERSION_PQC_DEF)
+        list(APPEND JSS_RAW_C_FLAGS "-DNSS_VERSION_PQC_DEF")
+    endif()
+
     list(APPEND JSS_RAW_C_FLAGS "-Wall")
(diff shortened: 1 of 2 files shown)


cmake/JSSConfig.cmake
```diff
@@ -1,3 +1,10 @@
+##
+# NSS V3.112 have some PQC defs in auth_alg_defs of ssl3con.c
+# that need to be reflected into JSS SSLCipher.c (auth_alg_defs[])
+# The upstream NSS (currently v3.113) apparently has no such defs.
+#
+option(ENABLE_NSS_VERSION_PQC_DEF "Enable PQC DEF to match NSS" OFF)
```
edewata, line 6: Usually the options are defined in the top-level `CMakeLists.txt`.

edewata (approved):
Just one suggestion, but it looks good to me.

cmake/JSSConfig.cmake
ladycfu: > Usually the options are defined in the top-level `CMakeLists.txt`.

OK, I will move the Option back to CMakeLists.txt.  That's where I put it yesterday.

PR #1055, opened 2025-07-22
fix build failure in dist-git centos caused by udpated NSS version wi…

ladycfu:
…th new pqc defs int auth_alg_defs[]

### DIFF
--- a/jss.spec
+++ b/jss.spec
@@ -28,6 +28,10 @@ License:        (MPL-1.1 OR GPL-2.0-or-later OR LGPL-2.1-or-later) AND Apache-2.
 # - GA/update:                 <major>.<minor>.<update>
 %global         full_version %{major_version}.%{minor_version}.%{update_version}%{?phase:-}%{?phase}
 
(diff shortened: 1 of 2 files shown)


jss.spec
```diff
@@ -324,7 +324,7 @@ touch %{_vpath_builddir}/.targets/finished_generate_javadocs
     --lib-dir=%{_libdir} \
     --sysconf-dir=%{_sysconfdir} \
     --share-dir=%{_datadir} \
-    --cmake=%{__cmake} \
+    --cmake="%{__cmake} -DENABLE_NSS_VERSION_PQC_DEF=ON" \
```
edewata, line 327: This option is only needed on CentOS/RHEL 10, but the CI is running on Fedora 42 so it's failing. We need to use a conditional like this:
```
%if 0%{?rhel} && 0%{?rhel} >= 10
... specify the option ...
%endif
```

edewata (changes requested):

PR #1057, opened 2025-07-25
Add matching PQC algs to auth_alg_defs[] with NSS

ladycfu:
Fix build failure caused by updated NSS version with new pqc defs int auth_alg_defs[] This requires the previous commit:
8cbe1faa8aa0cd0c6cf91ed04e5d06678d114722

This patch makes cognizant of the os version.

fixes https://issues.redhat.com/browse/IDM-2428


edewata (approved):
LGTM

PR #1060, opened 2025-07-31
Reintroduce SHA-1 references from JSSProvider

vzlamal:
Adding SHA1/RSA alias to JSSProvider.

Recently I run into issue during update of PKI infrastructure where some old systems are sending CSRs with `SHA1withRSA` singing alg. This fails with:

```
SEVERE: Unable to parse PKCS #10 request: no such algorithm: SHA1/RSA for provider Mozilla-JSS
java.security.NoSuchAlgorithmException: no such algorithm: SHA1/RSA for provider Mozilla-JSS
```

Since I need to support this old CSRs I dig into the code and found out that SHA1 algs was reintroduced to the JSS (here: https://github.com/dogtagpki/jss/commit/11ecabc9086f8b797b7545356a7b2fb8c7e94677) after removal (here: https://github.com/dogtagpki/jss/commit/f9d83e27ff6d22c9cefe4a7d1f651498782bb8cb).

In this commit  SHA1 alias was removed https://github.com/dogtagpki/jss/commit/038a3a5bb7ae92890ad1a08a2f20bea416d1da95 but never reintroduced.

So here is a version that should fix my issue.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/JSSProvider.java
+++ b/base/src/main/java/org/mozilla/jss/JSSProvider.java
@@ -140,6 +140,16 @@ public final class JSSProvider extends java.security.Provider {
         /////////////////////////////////////////////////////////////
         // Signature
         /////////////////////////////////////////////////////////////
+        put("Signature.SHA1withDSA",
+            "org.mozilla.jss.provider.java.security.JSSSignatureSpi$DSA");
+
+        put("Alg.Alias.Signature.DSA", "SHA1withDSA");
+        put("Alg.Alias.Signature.DSS", "SHA1withDSA");
+        put("Alg.Alias.Signature.SHA/DSA", "SHA1withDSA");
+        put("Alg.Alias.Signature.SHA-1/DSA", "SHA1withDSA");
+        put("Alg.Alias.Signature.SHA1/DSA", "SHA1withDSA");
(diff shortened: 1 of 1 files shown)


vzlamal:
I'm still testing this.
```
$ pki --version
PKI Command-Line Interface 11.6.0-1

$ dnf list --installed | grep jss
idm-jss.x86_64                                 5.6.0-1.el9                      @appstream
idm-jss-tomcat.x86_64                          5.6.0-1.el9                      @appstream
```

I've build RPM on the system and hot swapped the `jss.jar`

rjrelyea:
Having these is RHEL-8 makes sense, but do we really want to have them going forward. SHA1 should not be used in modern systems, and certain should never be used in signatures anymore... and certificates.
The issues with SHA1 in other cases (like HMAC) are theoretical, but the known weaknesses of SHA1 directly apply to it's use in certificates... and in particular certificate issuance. In fact, even after adding this code back, it should not work on a normally configured RHEL-9 and RHEL-10 system because we turn off SHA-1 signatures by policy (we don't create them and we don't trust them).

fmarco76:
SHA-1 was intentionally removed for its weakness. Current policies in modern system do not support it so I do not think it is OK to put it back. A better approach should be to move old legacy system to current standards.
If this is is a problem in many deployments then this could be take in consideration.

PR #1061, opened 2025-07-31
Add matching PQC algs to auth_alg_defs[] with NSS

ladycfu:
Fix build failure caused by updated NSS version with new pqc defs int auth_alg_defs[]
For reference, this change corresponds to an NSS patch that currently exists on CentOS 10 only:
    https://gitlab.com/redhat/centos-stream/rpms/nss/-/blob/c10s/nss-3.112-add-ml-dsa-ssl-support.patch?ref_type=heads#L965
    * Adding option for ENABLE_NSS_VERSION_PQC_DEF (#1053)
    * cognizant of the os version.

fixes https://issues.redhat.com/browse/IDM-2428

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -46,7 +46,7 @@ option(TEST_WITH_INTERNET "When enabled, runs various tests which require an int
 ##
 # NSS V3.112 have some PQC defs in auth_alg_defs of ssl3con.c
 # that need to be reflected into JSS SSLCipher.c (auth_alg_defs[])
-# The upstream NSS (currently v3.113) apparently has no such defs.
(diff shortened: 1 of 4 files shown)


rjrelyea:
This code looks very brittle, We should probably figure a way to make it usable for you without breaking everytime we add new identifiers to SSL.

One obvious thing is to at least check that you aren't referencing beyond your array bounds.  That way you won't crash every time a new algorithm is added (you may incorrectly disable it). 

For the rest, it probably makes sense to ask for a function in NSS to do these checks for you. You can write an upstream bug here:
https://bugzilla.mozilla.org
(you can find NSS under 'Other Projects -> Components').

PR #1065, opened 2025-09-17
Update PKCS11 constants

fmarco76:
NSS has deprecated some constants and modified their values. The macro used to deprecate the value generate problem with the constant generator implemented in JSS so it has been fixed and a new file with updated values has been generated.

The constants creating problem where:

- CKT_NSS_UNTRUSTED
- CKT_NSS_VALID
- CKT_NSS_MUST_VERIFY

The new definition is in

https://github.com/nss-dev/nss/blob/bb3335174cfdf764cfe3b6806c3ad45c1b3d8d0b/lib/util/pkcs11n.h#L621-L632

Note: The script has been modified with the help of Claude Code (https://claude.com/product/claude-code).


tools/build_pkcs11_constants.py
```diff
@@ -320,14 +321,33 @@ def get_source_content(self):
...
     # we don't care about whitespace at the end. We care about prefixed
     # whitespace when reading the copyright header though, hence rstrip()
     # and not strip().
-    return list(map(lambda x: x.rstrip(), file_handle.readlines()))
+    raw_lines = list(map(lambda x: x.rstrip(), file_handle.readlines()))
+    
+    # Handle line continuation with backslash
+    merged_lines = []
+    i = 0
+    while i < len(raw_lines):
```
jmagne, line 338: Maybe pre set len(rawlines) outside the loop?

jmagne (approved):
This looks fine. I made a simple suggestion , but feel free to ignore since this doesn't run that often.

fmarco76:
@jmagne Thanks!

PR #1075, opened 2026-02-17
Update password callback in TomcatJSS connector

edewata:
Previously during PKI server initialization if the token password is not available in `password.conf` or if the entire `password.conf` does not exist (e.g. when Nuxwdog is enabled) TomcatJSS would ask for the password on the console.

The `TomcatJSS` class has been updated to try to get the password from the password store configured in `server.xml` first, then if it's not available it will ask for the password on the console.

In regular PKI servers the password store in `server.xml` will get the passwords from `password.conf`. If Nuxwdog is enabled the password store will get the passwords from user keyring.


fmarco76 (commented):
With this change in TomcatJSS if the password file is missed or it miss a token password tomcat will wait for a user to write in the console. Since this should be in a service daemon waiting for user input could be not ideal in some cases. Should this be configurable?
[Nuxwdog startup script](https://github.com/dogtagpki/pki/blob/master/base/server/scripts/pki-server-nuxwdog#L140) already ask for user input if password cannot be found from keyring, is this intended to replace the current approach?

PR #1076, opened 2026-02-18
Fix application data send with multiple packets

fmarco76:
The additional check for buffer size to avoid looping in case of big handshake packets was blocking also the application data to be sent.

To avoid limiting the application data an additional check on the handshake status is included. If the handshake is complete then the buffer size can contain the handshake packets and there is no need to perform additional checks.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -1220,7 +1220,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
         // Check if the incoming packet is larger than our buffer capacity.
         // This prevents silent performance degradation from looping when
         // processing large TLS packets (e.g., ML-DSA certificates).
-        if (src_capacity > bufferSize) {
+        if (src_capacity > bufferSize && !handshake_already_complete) {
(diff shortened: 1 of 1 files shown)


fmarco76:
@jmagne The sequence of operation is like:
- `wrap()` or `unwrap()` methods are calling depending is client or server;
- these method invoke the `updateHandshakeState()` in the loop which, if the handshake has not been completed try to advance it with the method  `SSL.ForceHandshake()`;
- the method will write in the buffer  the certificates, ciphers, or any relevant message for the henshake;
- the initial method will get the control and send the data;
- when all the information have been exchanged and the secure communication can be established the handshake is completed and only application data are sent received.

The problem I fix with the previous PR is that when NSS try to put the certificate in the buffer it return error if the buffer if smaller and following call try agin the same operation of sending the certificate. This will create a an infinite loop.
To stop the problem I have included a check so if the buffer is full it should stop because there is a problem.

What was identified now is that when a big amount of application data are sent the buffer could get full and the exception is raised. However, in this case there is no problem because the loop will drain the buffer without creating new data. To avoid the problem I have added the condition that the full buffer is a problem only if the handhake has not been completed, this imply that the certificate chain is too big for the buffer.

jmagne (approved):
Thanks for the info. The part I was most interesting in was what happens when the handshake is completed and this overrun is detected.  If there are no negative consequences to this then will approve.

PR #1080, opened 2026-03-20
Fix race condition segfault in JSSEngineReferenceImpl cleanup

edewata:
The finalizer thread was experiencing segfaults during `PR.Shutdown()` calls due to a race condition where multiple threads could execute cleanup operations simultaneously on the same object.

Root cause:
- `cleanup()`, `closeInbound()`, `closeOutbound()`, and `tryCleanup()` methods were not synchronized
- Multiple threads could simultaneously check and modify the boolean flags (`closed_fd`, `is_inbound_closed`, `is_outbound_closed`)
- One thread could close/free `ssl_fd` while another thread was still using it in `PR.Shutdown()`, causing `SIGSEGV` in NSS `memcpy`

Changes:
- Add `synchronized` modifier to `closeInbound()`, `closeOutbound()`, `cleanup()`, and `tryCleanup()` methods
- The `synchronized` keyword provides both mutual exclusion (only one thread can execute these methods at a time) and memory visibility (changes to fields are visible to other threads)

This prevents concurrent cleanup operations from corrupting the native `PRFileDesc` pointer and eliminates the segfault during finalization.

Assisted-by: Claude Sonnet 4.5

**Note:** This PR is originally generated by Claude and updated based on reviews. There's no guarantee it will fully fix the issue. We'll need to keep monitoring for crashes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Connection shutdown and cleanup now run serially to avoid interleaving during concurrent teardowns. This reduces race conditions during SSL connection close, improving stability, preventing sporadic errors or resource leaks, and making simultaneous connection terminations more reliable.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -736,7 +736,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
     }
 
     @Override
-    public void closeInbound() {
+    public synchronized void closeInbound() {
         debug("JSSEngine: closeInbound()");
 
         if (!is_inbound_closed && ssl_fd != null && !closed_fd) {
@@ -750,7 +750,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
     }
 
     @Override
-    public void closeOutbound() {
+    public synchronized void closeOutbound() {
         debug("JSSEngine: closeOutbound()");
 
         if (!is_outbound_closed && ssl_fd != null && !closed_fd) {
@@ -1675,7 +1675,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
      * connection.
      */
     @Override
-    public void tryCleanup() {
+    public synchronized void tryCleanup() {
         debug("JSSEngine: tryCleanup()");
         if (is_inbound_closed && is_outbound_closed) {
             // throw new RuntimeException("Probably shouldn't be here!");
@@ -1688,7 +1688,7 @@ public class JSSEngineReferenceImpl extends JSSEngine {
      * data streams if still open.
      */
     @Override
-    public void cleanup() {
+    public synchronized void cleanup() {
         debug("JSSEngine: cleanup()");
 
         if (!is_inbound_closed) {


base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngine.java
```diff
@@ -181,7 +181,7 @@ public abstract class JSSEngine extends javax.net.ssl.SSLEngine {
     /**
      * Whether or not the outbound portion of this connection is closed.
      */
-    protected boolean is_outbound_closed;
+    protected volatile boolean is_outbound_closed;
```
fmarco76, line 185: Since all method accessing this variable are now syncronized, is this still to be `volatile`?

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1730,7 +1730,7 @@ private void cleanupLoggingSocket() {
         }
     }
 
-    private void cleanupSSLFD() {
+    private synchronized void cleanupSSLFD() {
```
fmarco76: Is this needed? It is invoked only from syncronised methods.

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1192,7 +1192,7 @@ private void logUnwrap(ByteBuffer src) {
     }
 
     @Override
-    public SSLEngineResult unwrap(ByteBuffer src, ByteBuffer[] dsts, int offset, int length) throws IllegalArgumentException, SSLException {
+    public synchronized SSLEngineResult unwrap(ByteBuffer src, ByteBuffer[] dsts, int offset, int length) throws IllegalArgumentException, SSLException {
```
fmarco76: `wrap()` and `unwrap()` may be concurrent for `SSSLEngine` specification. The operation should not interfere beside the possible cleanup.

fmarco76 (commented):
I have several doubt on few extra syncronisation added (see the comments). Could we limit to syncronise only the methods `closeInbound()`, `closeOutbound()`, `tryCleanup()` and `cleanup()`?
It is not sure they are the reasons for these segmentation fault but the race condition is present so this fix is OK in any case.

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngine.java
edewata: That change was suggested by Gemini (see above). I've reverted it now.

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
edewata: This change was originally generated by Claude. I see that right now it's only called by `cleanup()` which is already `synchronized`, but the `cleanupSSLFD()` is reading and modifying `closed_fd` and `ssl_fd` so it would make sense to make it `synchronized` in case we accidentally call it from a non-`synchronized` method. The extra `synchronized` will basically be ignored if the thread already have the lock, so it shouldn't cause any problem, but we can drop it if you prefer.

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
edewata: This change was also suggested by Gemini (see above). I've reverted it now.

edewata:
Thanks for the comments. I've dropped the changes suggested by Gemini, so the PR is reverted back to the original one generated by Claude. However, it looks like crashes are still happening despite the fixes made by Claude and/or Gemini, so we need to keep looking.

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
fmarco76: For what I see in the code `cleanupSSLFD()` is private and invoked only from `cleanup()`. If it get invoked in other places without a cleanup I am expecting failure because we are closing underlying sockets but still using it.  So it is not harmful but quite useless IMHO and I think has not to be included.

By the way, I have also used calude to verify the changes and my comments.

fmarco76 (approved):
I have a small doubt for `cleanupSSLFD` (see the comment) but the remaining syncronisation are OK. Thanks for the update!

 Feel free to update or merge.

For the segmentation fault I was expecting the error was not related. This code has not been modified to justify these error. Will continue the investigation.

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
edewata: OK, let me drop it then.

edewata: It's updated now.

edewata:
@fmarco76 Thanks! I'll merge once the CI completes.

edewata:
@coderabbitai resume

PR #1082, opened 2026-03-26
Allow for encrypting with CKM_AES_KEY_WRAP_KWP

Freddy113-x:
## Summary by CodeRabbit

* **New Features**
  * Added AES Key Wrap with Padding (AES-KWP) encryption algorithm with 256-bit key strength to expand cryptographic options for secure key wrapping
  * Updated algorithm type mappings to provide full integration of the new AES-KWP variant across key management operations and library functionality

### DIFF
--- a/base/src/main/java/org/mozilla/jss/crypto/EncryptionAlgorithm.java
+++ b/base/src/main/java/org/mozilla/jss/crypto/EncryptionAlgorithm.java
@@ -442,7 +442,7 @@ public class EncryptionAlgorithm extends Algorithm {
     public static final EncryptionAlgorithm AES_192_KEY_WRAP_KWP = new EncryptionAlgorithm(SEC_OID_AES_192_KEY_WRAP_KWP,
             Alg.AES, Mode.NONE,
             Padding.PKCS5, IVParameterSpecClasses, 16,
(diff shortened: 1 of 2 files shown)


jmagne:
Hi: I've checked into this by trying it out locally. It seems to build fine. I did even try to add some tests to the JSS build time test suite to see if thes KWP based Encryption class registrations can actually be used.  The attempts to encrypt with the EncryptionAlgorithm.AES_KEY_WRAP_PAD_KWP algorithm added by this fix. Unfortunately the test doesn't work. From initial research I "think" it's because we don't support a single pass encryption from JSS. We have update and doFinal. NSS may allow the one shot encryption for this algorithm possibly. This would take more investigation. Anway, the the question this, even though this fix is good, what use have you guys for this? Thanks.

Freddy113-x:
Yes I noticed that as well with AES-256 after I submitted this and can confirm that the fix is to do a one shot encryption for it. We've tested it with other pkcs11 tools for HSMs and the single part encryption works for it. The use case for this fix is to help with migrating or backing up your symmetric/asymmetric keys and getting a known symmetric key in both HSMs while in FIPS mode. If AES KWP encryption is implemented, we can encrypt and unwrap a known key into both HSMs and then leverage them for use.

fmarco76 (approved):

fmarco76:
@Freddy113-x Thanks for the PR.

PR #1083, opened 2026-04-01
Update Password.readPasswordFromConsole()

edewata:
The `Password.readPasswordFromConsole()` has been modified to read the password from the standard input in case the system console is not available (e.g. in CI environment).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved password input handling to work reliably when console access is unavailable (e.g., non-interactive or IDE environments); preserves prior behavior when a console is present.
  * Better handles input termination and errors, avoiding silent empty passwords and surfacing failures appropriately.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/util/Password.java
+++ b/base/src/main/java/org/mozilla/jss/util/Password.java
@@ -286,7 +286,37 @@ public class Password implements PasswordCallback, Cloneable,
     public static Password readPasswordFromConsole() throws PasswordCallback.GiveUpException {
 
         Console console = System.console();
-        char[] password = console.readPassword();
+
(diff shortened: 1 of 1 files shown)


fmarco76 (commented):
The Scanner close the underlying readable, maybe it is not a problem in this specific case since there should be no other input but it is better to modify to be safe.

edewata:
@fmarco76 Thanks. I've updated the code as suggested by Gemini. It does seem to be able to keep the `System.in` open for reading multiple passwords. I'm working a PR for PKI that will need that.

fmarco76 (approved):
LGTM

edewata:
@fmarco76 Thanks!

PR #1084, opened 2026-04-06
Fix TOCTOU race condition in JSSEngineReferenceImpl cleanup

edewata:
The finalizer thread was still experiencing segfaults after the initial synchronization fix (7eb0818c) due to a time-of-check to time-of-use race condition in `cleanupSSLFD()`.

Root cause:
- The `closed_fd` flag was set in the finally block AFTER native resources were freed (`ssl_fd.close()`)
- Another thread could call `closeInbound()`/`closeOutbound()`, check `!closed_fd` (still `false`), and attempt `PR.Shutdown()` on the already-freed `ssl_fd` pointer, causing SIGSEGV in NSS `memcpy`

The fix:
- Set `closed_fd = true` BEFORE freeing native resources
- This ensures the guard checks in `closeInbound()`/`closeOutbound()` will prevent any concurrent `PR.Shutdown()` calls on `ssl_fd` that is being freed
- Remove finally block since `closed_fd` is now set before exception can occur

This completes the fix for the race condition segfault in cleanup.

Assisted-by: Claude Sonnet 4.5

### DIFF
--- a/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -1732,14 +1732,18 @@ public class JSSEngineReferenceImpl extends JSSEngine {
 
     private void cleanupSSLFD() {
         if (!closed_fd && ssl_fd != null) {
+            // Set closed_fd BEFORE freeing native resources to prevent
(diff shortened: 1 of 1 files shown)


fmarco76 (approved):
LGTM but the `ssl_fd` should be in finally so it cannot be used even an exception occur avoiding potential segfault. Feel free to update and merge.

edewata:
@fmarco76 Thanks! I've updated the PR as suggested.

PR #1094, opened 2026-04-28
Add ML-KEM support for key archival and unwrapping

ladycfu:
This patch adds support for ML-KEM private key unwrapping needed for key archival operations in Dogtag PKI's KRA (Key Recovery Authority).

Changes:

1. JSSProvider.java: Add generic ML-KEM and ML-DSA KeyFactory aliases
   - Added KeyFactory.ML-KEM mapping to existing ML-KEM-512/768/1024
   - Added KeyFactory.ML-DSA mapping to existing ML-DSA-44/65/87
   - Enables KeyFactory.getInstance("ML-KEM") without specifying variant

2. PrivateKey.java: Fix CKK_ML_KEM constant and add ML-KEM types
   - Fixed CKK_ML_KEM from 0x4b to 0x49 (matches PKCS#11 v3.2)
   - Added MLKEM512, MLKEM768, MLKEM1024, MLKEM type constants
   - Enables PrivateKey.Type.MLKEM for generic ML-KEM key handling

3. PK11KeyWrapper.java: Add ML-KEM public key extraction workaround
   - Implemented DER parser to extract ML-KEM public key from SPKI
   - TODO: Replace with PK11MLKEMPublicKey.getKeyByteArray() when available
   - Added helper methods getLength() and getLengthBytes() for DER parsing
   - Required for baseUnwrapPrivate() to get public key bytes for unwrap

4. PK11KeyWrapper.c: Add CKK_ML_KEM support in nativeUnwrapPrivWithSym
   - Added CKK_ML_KEM case with CKA_DECAPSULATE and CKA_EXTRACTABLE
   - Enables unwrapping ML-KEM private keys from wrapped format

5. PK11MLKEMPublicKey.java.proposed: Reference implementation
   - Shows proposed PK11MLKEMPublicKey class following EC/RSA pattern
   - Includes native getKeyByteArray() method signature
   - For discussion: should JSS provide this class?

Known Issues:
- NSS 3.123 PK11_GetKeyType() doesn't map CKM_ML_KEM → CKK_ML_KEM
- Causes unwrap to fail with "key does not support operation"
- Verified missing in NSS_3_123_RTM tag in pk11wrap/pk11mech.c
- Need to add CKM_ML_KEM_KEY_PAIR_GEN and CKM_ML_KEM cases to NSS
- Unwrapping of ml-kem private key fails

Testing:
- Used by Dogtag PKI hsmCompatVerifyServ for ML-KEM key archival
- Unwraps ML-KEM private keys encrypted with AES-KWP

IDM-5475
Assisted-by: Claude

### DIFF
--- a/native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyWrapper.c
+++ b/native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyWrapper.c
@@ -425,7 +425,14 @@ Java_org_mozilla_jss_pkcs11_PK11KeyWrapper_nativeUnwrapPrivWithSym
         goto finish;
     }
 
-    keyType = PK11_GetKeyType(keyTypeMech, 0);
+    /* Workaround: NSS PK11_GetKeyType() doesn't map PKCS#11 v3.2 ML-KEM mechanisms
+     * Without this, it returns CKK_GENERIC_SECRET (0x10) which also works,
+     * but using the proper CKK_ML_KEM (0x49) is clearer and more correct */
+    if (keyTypeMech == CKM_ML_KEM_KEY_PAIR_GEN || keyTypeMech == CKM_ML_KEM) {
+        keyType = CKK_ML_KEM;  /* 0x49 from PKCS#11 v3.2 */
+    } else {
+        keyType = PK11_GetKeyType(keyTypeMech, 0);
+    }
 
     /* special case nethsm and lunasa*/
     if( isNethsm ) {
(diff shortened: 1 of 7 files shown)


native/src/main/native/org/mozilla/jss/pkcs11/PK11KeyWrapper.c
```diff
@@ -466,6 +466,11 @@ Java_org_mozilla_jss_pkcs11_PK11KeyWrapper_nativeUnwrapPrivWithSym
         attribs[0] = CKA_DERIVE;
         numAttribs = 1;
 	break;
+    case CKK_ML_KEM:
+        /* ML-KEM is a KEM (Key Encapsulation Mechanism), not a signature key */
+        attribs[0] = CKA_DERIVE;
+        numAttribs = 1;
```
fmarco76, line 479: Should this be `CKA_DECAPSULATE`? See [nss pkcs11t](https://github.com/nss-dev/nss/blob/97c0216f88e044e4ab31a37f13754f2a48198103/lib/util/pkcs11t.h#L751). It should be a different mechanism 

Is the extrable needed?

PR #1097, opened 2026-05-13
Update java build to java jdk 25

fmarco76:
KEM algorithms are supported from Java 23 so the jdk has to be updated in order to build the code.

## Chores

* Updated Java Development Kit version from 17/21 to 25 across all continuous integration and code analysis pipelines.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dogtagpki/jss/pull/1097)

### DIFF
--- a/.github/workflows/build-tests.yml
+++ b/.github/workflows/build-tests.yml
@@ -41,7 +41,7 @@ jobs:
     - name: Set up Java
       uses: actions/setup-java@v4
       with:
-        java-version: '21'
+        java-version: '25'
         distribution: 'adopt'
 
     - name: Install dependencies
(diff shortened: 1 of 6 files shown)


edewata (approved):
I think the Java versions in `jss.spec` need to be updated as well in case we need to build for other platforms, but this PR LGTM.

fmarco76:
> I think the Java versions in `jss.spec` need to be updated as well in case we need to build for other platforms, but this PR LGTM.

`jss.spec` requires java 25 for new fedora. Older version have to be removed as well as PQC flags and other option no more supported. This clean-up will be in a separate PR.

fmarco76:
@edewata Thanks!

PR #1103, opened 2026-05-19
Fix default ML-DSA signing

fmarco76:
According to JEP 497 [1], the ML-DSA signature algorithm should behave as follows:
- When instantiated as "ML-DSA" (generic), it accepts any ML-DSA-44, ML-DSA-65, or ML-DSA-87 key and uses the key's variant
- When instantiated with a specific variant (e.g., "ML-DSA-65"), it only accepts keys of that variant, throwing InvalidKeyException otherwise

Previously, the generic "ML-DSA" was hardcoded to "ML-DSA-65", which worked in practice because NSS uses the key's variant regardless of the signature algorithm parameter. However, this violated the JEP 497 specification.

This commit implements the correct behavior:
- Generic ML-DSA signature extracts the variant from the key (NamedParameterSpec for private keys, OID from X.509 encoding for public keys)
- Specific ML-DSA variants validate that the key matches

1. https://openjdk.org/jeps/497

## Summary by CodeRabbit

* **New Features**
  * Added support for the ML-DSA-87 signature variant.
  * Signature variant is now auto-detected from key parameters or encoded public key.

* **Breaking Changes**
  * The generic "ML-DSA" name no longer resolves; callers must use ML-DSA-44, ML-DSA-65, or ML-DSA-87 explicitly.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dogtagpki/jss/pull/1103?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/base/src/main/java/org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
+++ b/base/src/main/java/org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
@@ -13,13 +13,17 @@ import java.security.PublicKey;
 import java.security.SecureRandom;
 import java.security.SignatureException;
 import java.security.spec.AlgorithmParameterSpec;
+import java.security.spec.NamedParameterSpec;
 import java.security.spec.X509EncodedKeySpec;
+import org.mozilla.jss.asn1.ASN1Util;
 
 import org.mozilla.jss.crypto.CryptoToken;
 import org.mozilla.jss.crypto.PrivateKey;
 import org.mozilla.jss.crypto.SignatureAlgorithm;
 import org.mozilla.jss.crypto.TokenException;
 import org.mozilla.jss.crypto.TokenSupplierManager;
+import org.mozilla.jss.pkcs11.KeyType;
+import org.mozilla.jss.pkix.primitive.SubjectPublicKeyInfo;
 
 public class JSSSignatureSpi extends java.security.SignatureSpi {
 
@@ -74,10 +78,28 @@ public class JSSSignatureSpi extends java.security.SignatureSpi {
             throw new InvalidKeyException();
         }
         privk = (PrivateKey)privateKey;
-
+        SignatureAlgorithm effectiveAlg = this.alg;
+        if (privk.getAlgorithm().equals(KeyType.MLDSA.toString()) &&
+                privk.getParams() instanceof NamedParameterSpec param) {
+            if (effectiveAlg == null) {
+                effectiveAlg = switch(param.getName()) {
+                    case "ML-DSA-44" -> SignatureAlgorithm.MLDSA44;
+                    case "ML-DSA-65" -> SignatureAlgorithm.MLDSA65;
+                    case "ML-DSA-87" -> SignatureAlgorithm.MLDSA87;
+                    default -> throw new NoSuchAlgorithmException("Signature algorithm not defined for " + param.getName());
+                };
+            } else {
+                if (!effectiveAlg.toString().equals(param.getName())) {
+                    throw new InvalidKeyException("Invalid " + param.getName() + " parameter for " + effectiveAlg);
+                }
+            }
+        }
         token = privk.getOwningToken();
 
-        return token.getSignatureContext(alg);
+        if (effectiveAlg == null) {
+            throw new NoSuchAlgorithmException("Signature algorithm not defined");
+        }
+        return token.getSignatureContext(effectiveAlg);
     }
 
     @Override
@@ -87,8 +109,6 @@ public class JSSSignatureSpi extends java.security.SignatureSpi {
         try {
             CryptoToken token =
               TokenSupplierManager.getTokenSupplier().getThreadToken();
-            sig = token.getSignatureContext(alg);
-
             // convert the public key into a JSS public key if necessary
             if( ! (publicKey instanceof org.mozilla.jss.pkcs11.PK11PubKey) ) {
                 if( ! publicKey.getFormat().equalsIgnoreCase("X.509") ) {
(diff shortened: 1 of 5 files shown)


base/src/main/java/org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
```diff
@@ -74,10 +78,28 @@ public void engineInitSign(java.security.PrivateKey privateKey)
...
-
+        SignatureAlgorithm effectiveAlg = this.alg;
+        if (privk.getAlgorithm().equals(KeyType.MLDSA.toString()) &&
+                privk.getParams() instanceof NamedParameterSpec param) {
+            if (effectiveAlg == null) {
+                effectiveAlg = switch(param.getName()) {
+                    case "ML-DSA-44" -> SignatureAlgorithm.MLDSA44;
+                    case "ML-DSA-65" -> SignatureAlgorithm.MLDSA65;
+                    case "ML-DSA-87" -> SignatureAlgorithm.MLDSA87;
+                    default -> throw new NoSuchAlgorithmException("Signature algorithm not defined");
```
edewata: Could we append the `param.getName()` into the exception message to help troubleshooting?

base/src/main/java/org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
```diff
@@ -74,10 +78,28 @@ public void engineInitSign(java.security.PrivateKey privateKey)
...
+                };
+            } else {
+                if (!effectiveAlg.toString().equals(param.getName())) {
+                    throw new InvalidKeyException();
```
edewata: Could we add an exception message to help troubleshooting, e.g. `Invalid <param> parameter for <algorithm>`?

base/src/main/java/org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
```diff
@@ -189,6 +222,22 @@ public void engineSetParameter(String param, Object value)
...
+            throw new InvalidKeyException("Unable to determine key algorithm");
+        }
+        try {
+            SubjectPublicKeyInfo spki = (SubjectPublicKeyInfo)
+                          ASN1Util.decode(new SubjectPublicKeyInfo.Template(), encoded);
+             SignatureAlgorithm keyAlg = SignatureAlgorithm.fromOID(
+                     spki.getAlgorithmIdentifier().getOID());
+             return keyAlg;
+        } catch (Exception ex) {
+            throw new InvalidKeyException("Unable to determine key algorithm");
```
edewata: Let's chain the original exception to the new exception and append the original exception message to help troubleshooting.

base/src/main/java/org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
```diff
@@ -102,11 +122,24 @@ public void engineInitVerify(PublicKey publicKey)
                     publicKey.getAlgorithm(), "Mozilla-JSS");
                 publicKey = fact.generatePublic(encodedKey);
             }
-
+            SignatureAlgorithm effectiveAlg = this.alg;
+            if (publicKey.getAlgorithm().equals(KeyType.MLDSA.toString())) {
+                SignatureAlgorithm keyAlg = getKeyAlg(publicKey.getEncoded());
```
edewata, line 127: This is fine, but is it possible to determine the signature algorithm from the `publicKey` directly instead of parsing the key info? I'm thinking maybe we should have something like `PK11PrivKey.getType()` or `getKeyType()` but for `PK11PubKey`.

edewata (approved):
I have some comments below but in general it LGTM. Feel free to update/merge.

base/src/test/java/org/mozilla/jss/tests/GenerateTestCert.java
```diff
@@ -109,7 +109,7 @@ private void setSigAlg(String alg) {
         } else if (alg.equalsIgnoreCase("SHA-512/EC")) {
             sigAlg = SignatureAlgorithm.ECSignatureWithSHA512Digest;
         } else if (alg.equalsIgnoreCase("ML-DSA")) {
-            sigAlg = SignatureAlgorithm.MLDSA;
+            sigAlg = SignatureAlgorithm.MLDSA65;
```
ladycfu, line 112: how about tests for other strengths?

base/src/main/java/org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
```diff
@@ -102,11 +122,24 @@ public void engineInitVerify(PublicKey publicKey)
...
+                if (effectiveAlg == null) {
+                    effectiveAlg = keyAlg;
+                } else {
+                    if (!effectiveAlg.equals(keyAlg)) {
+                        throw new InvalidKeyException("Key variant " + keyAlg +
+                                  " does not match signature algorithm " + alg);
```
ladycfu: maybe it's the same thing, but maybe it'd be more clear if you use "effectiveAlg" here instead of "alg" in the exception message.

base/src/main/java/org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
ladycfu: perhaps add the "param.getName()" to the exception message

base/src/main/java/org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
fmarco76: According to the Java specification the same approach based on named parameter should be valid for private and public key. However, `PuiblicKey` interface does not have the `getParams()` in java 21. It was introduced in java23. As a result I am leaving this for the moment and postpone the change at when java 25 will be adopted (this version has a native support to ML-KEM), unless it is required to get the correct strength from the public key in other places.

base/src/test/java/org/mozilla/jss/tests/GenerateTestCert.java
fmarco76: This test is for generating certificates to be used in other tests. For the signature there is a different test file [JCASigTest.java](https://github.com/dogtagpki/jss/blob/master/base/src/test/java/org/mozilla/jss/tests/JCASigTest.java). I'll add some additional tests to verify the case of key not matching the signature algorithm.
Do we also need additional keys in this test?

base/src/main/java/org/mozilla/jss/provider/java/security/JSSSignatureSpi.java
edewata: We might be able to move the current `PK11PrivKey.getParams()` into `PK11Key` (or maybe create a new `PK11AsymmetricKey`), then whenever we need to use the method we can downcast the key into `PK11PubKey`. Basically we should create a class hierarchy in JSS that mimics the standard classes. We can do this separately later.

ladycfu (approved):
LGTM

fmarco76:
@edewata @ladycfu Thanks!

PR #1104, opened 2026-05-21
Force SHA-512 digest algorithm for ML-DSA in CMS SignerInfo

fmarco76:
Per RFC 9882 Section 4 [1], ML-DSA signatures in CMS must use SHA-512 as the digest algorithm identifier in the SignerInfo structure. This applies both when signed attributes are absent (pure signatures) and when they are present.

ML-DSA (FIPS 204) is a pure signature algorithm that does not have an associated digest algorithm like traditional composite signature algorithms (e.g., RSASignatureWithSHA256). However, RFC 9882 mandates that the digestAlgorithm field in SignerInfo MUST be set to id-sha512 to ensure proper CMS structure and interoperability.

This change adds exception handling to catch NoSuchAlgorithmException when getDigestAlg() is called on ML-DSA signature algorithms (ML-DSA-44, ML-DSA-65, ML-DSA-87), and explicitly sets the digest algorithm to SHA-512 in these cases.

1. RFC 9882 - Using ML-DSA in the Cryptographic Message Syntax (CMS) (https://www.rfc-editor.org/rfc/rfc9882)

Assisted-By: Claude Sonnet 4.5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved digest algorithm selection for certain signing methods to ensure the correct hash is used and avoid signing failures or invalid signatures in edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/dogtagpki/jss/pull/1104?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/base/src/main/java/org/mozilla/jss/pkix/cms/SignerInfo.java
+++ b/base/src/main/java/org/mozilla/jss/pkix/cms/SignerInfo.java
@@ -225,9 +225,21 @@ public class SignerInfo implements ASN1Value {
         } else {
             throw new IllegalArgumentException("Unexpected SignerIdentifier type");
         }
-        this.digestAlgorithm =
-                new AlgorithmIdentifier(signingAlg.getDigestAlg().toOID(),null);
-
+        if(signingAlg.equals(SignatureAlgorithm.MLDSA44) ||
+                signingAlg.equals(SignatureAlgorithm.MLDSA65) ||
+                signingAlg.equals(SignatureAlgorithm.MLDSA87)) {
+            // Per RFC 9882 Section 4, for ML-DSA the digestAlgorithm MUST be
+            // id-sha512 and the parameters MUST be absent.
+            this.digestAlgorithm =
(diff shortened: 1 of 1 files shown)


ladycfu (commented):
LGTM.  Just one minor thing regarding the discrepancy between your commit message and the code. In the commit message you mentioned "This change adds exception handling to catch NoSuchAlgorithmException " but the code actually does if/else condition (which I think is better).

ladycfu (approved):
feel free to fix the commit message and merge.

fmarco76:
> LGTM. Just one minor thing regarding the discrepancy between your commit message and the code. In the commit message you mentioned "This change adds exception handling to catch NoSuchAlgorithmException " but the code actually does if/else condition (which I think is better).

Yes, I have updated the implementation after the PR was created but forget to update the commit message. I'll update, thanks!

fmarco76:
@ladycfu Thanks!

PR #1107, opened 2026-06-01
Fix Intermittent SIGSEGV in __memmove_avx_unaligned_erms during pki C…

jmagne:
Fix Intermittent SIGSEGV in __memmove_avx_unaligned_erms during pki CLI commands.

This commit is to test a proposed fix for JSS by running the CI suite and observing the behavior.

Asisted-by: Claude Sonnet 4.5.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved native resource cleanup during garbage collection finalization to ensure proper handling and prevent resource leaks.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -1758,12 +1758,21 @@ public class JSSEngineReferenceImpl extends JSSEngine {
         }
     }
 
+    // Old comment, leaving for reference...
     // During testing with Tomcat 8.5, most instances did not call
     // cleanup, so all the JNI resources end up getting leaked: ssl_fd
     // (and its global ref), read_buf, and write_buf.
+    //
+    // Latest findings...
+    // Make this method empty. Each native resource (ssl_fd, read_buf, write_buf) has
+    // its own NativeProxy finalizer that handles individual cleanup.
+    // Do not call cleanup() here — GC finalization order is undefined,
+    // so BufferProxy objects may already be finalized, making
+    // PR.Close() or PR.Shutdown() crash with use-after-free (SIGSEGV
+    // in memcpy).
+
     @Override
     protected void finalize() {
-        cleanup();
     }
 
     private class CertValidationTask extends CertAuthHandler {


base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1758,12 +1758,21 @@ private void cleanupSSLFD() {
...
+    // its own NativeProxy finalizer that handles individual cleanup.
+    // Do not call cleanup() here — GC finalization order is undefined,
+    // so BufferProxy objects may already be finalized, making
+    // PR.Close() or PR.Shutdown() crash with use-after-free (SIGSEGV
+    // in memcpy).
+
     @Override
     protected void finalize() {
-        cleanup();
     }
```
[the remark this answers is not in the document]
jmagne: The empty override is intentional. Removing it would cause NativeProxy.finalize() to execute, which calls releaseNativeResources() on the SSLFDProxy, potentially racing with BufferProxy finalization — the same use-after-free this fix addresses.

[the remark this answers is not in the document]
edewata: @jmagne Should we drop `NativeProxy.finalize()` too then? So the caller will be responsible to call `close()` explicitly.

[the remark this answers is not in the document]
jmagne: That's a possibility . I would have to look at it. I guess we can observe this one for now, since the end result of that would most likely be the same.

edewata:
We actually have a few tests with PKI CLI and one of them crashed:
https://github.com/dogtagpki/jss/actions/runs/26769451749/job/78907888569?pr=1107#step:18:27

edewata:
Just FYI, you can add something like this to capture JVM error log:
https://github.com/dogtagpki/pki/blob/master/.github/workflows/ca-api-test.yml#L305-L309

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
```diff
@@ -1758,12 +1758,21 @@ private void cleanupSSLFD() {
...
+    // Make this method empty. Each native resource (ssl_fd, read_buf, write_buf) has
+    // its own NativeProxy finalizer that handles individual cleanup.
+    // Do not call cleanup() here — GC finalization order is undefined,
+    // so BufferProxy objects may already be finalized, making
+    // PR.Close() or PR.Shutdown() crash with use-after-free (SIGSEGV
+    // in memcpy).
+
     @Override
     protected void finalize() {
-        cleanup();
```
fmarco76, line 1766: If resources would always be removed the segmentation fault should be always present. Looking at the error and the solution I do not think the cleanup should be removed from `finalize()` (or its modern equivalent since this approach is deprecated). The problem is in the `clenupSSLFD()`. I think the comment in line [1735](https://github.com/dogtagpki/jss/blob/a653cf48819fdad522ac62e58989e0924bcc9765/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java#L1735) is not respected. The `closed_fd = true` should be in the `cleanup()` before the operation of closing the channel to skip the `PR.shutdown()` and the `cleanSSLFD()` fixed accordingly.

fmarco76 (commented):
I am not sure about the approach. Look at the comment

base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
jmagne: You could be correct but it seems that the finalize order can't be predicted which can cause things to happen out of order, causing the crash. The separate resources have finalize methods which were observed by me to fire off in my testing. Also I was able to force a similar crash by modifying the SSLEngine test by not calling the testClose method. I just threw up this PR to see how it would behave , not as a be all end all solution.

jmagne: OK, I was able to finagle the SSLEngine test locally to only run one test that I modified to create the problem. I was able to see the problem happen and after putting the fix back in, the problem didn't happen. The caveat is that if you look at the original CI run for this PR, we got a very similar error when running a PKI cmd, WITH the fix. This tells us that the situation is more complex and there could be other resources at risk. I think the only quick solution may be to put the proposed fix in the PKI cmd class to close out resources correctly, which it's not currently doing. I will create a quick PKI PR with this very short fix and see how it goes.

fmarco76: @jmagne the fix for PKI is definitely needed because the resources have to be used correctly but the library should not generate the segmentation fault. I have created the PR #1108  with the fix I was proposed. Since you have reproduced the error in your CI, could you test it to understand if it solves the problem.

The finalize method should be replaced with a cleaner to have better control but from the comments it seems this was needed in some circumstances so I would prefer not remove without a deep analysis.

jmagne:
Closing since we have another PR that is looking more promising.

PR #1108, opened 2026-06-08
Fix use-after-free crash in JSSEngineReferenceImpl finalizer

fmarco76:
Set `closed_fd` before calling `closeInbound()/closeOutbound()` in cleanup() to prevent them from calling `PR.Shutdown()` on `ssl_fd` that is being freed. This fixes a SIGSEGV crash where `PR.Shutdown()` attempted to write TLS close_notify alerts to NSS buffers during cleanup.

The crash occurred because `cleanup()` called `closeOutbound()` which checked `!closed_fd` and invoked `PR.Shutdown()`, but `cleanupSSLFD()` set `closed_fd` too late. Moving `closed_fd = true` to the start of `cleanup()` prevents the native `PR.Shutdown()` call while still freeing all resources properly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a potential crash during SSL connection cleanup that could occur in background processing threads.

* **Tests**
  * Enhanced PKI test diagnostics to automatically collect and display core dump information when test failures occur.

### DIFF
--- a/.github/workflows/pki-ca-test.yml
+++ b/.github/workflows/pki-ca-test.yml
@@ -616,6 +616,12 @@ jobs:
           docker exec pki pki-server start --wait
           docker exec pki pkidestroy -i pki-tomcat -s CA -v
 
+      - name: Check for PKI core dumps
+        if: failure()
+        run: |
+          docker exec pki ls -l
+          docker exec pki find / -path /proc -prune -o -name "hs_err_pid*.log" -exec cat {} \;
+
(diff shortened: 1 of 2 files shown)


edewata (approved):
It's hard to be sure whether these changes will completely fix the issue (Claude has tried to fix it several times without success), but I don't have any objection. Feel free to merge assuming it doesn't introduce new issues.

fmarco76:
@edewata Thanks!

PR #1110, opened 2026-07-24
Validate PK11Context before use in PK11Cipher native calls

fmarco76:
JSS_PK11_getCipherContext extracted the PK11Context pointer from a CipherContextProxy without checking it for NULL, unlike the analogous JSS_PK11_getSigContext used by PK11Signature. If the underlying context was ever invalid (e.g. already released), the NULL pointer was passed straight into PK11_CipherOp/PK11_DigestFinal, crashing the whole JVM with a SIGSEGV instead of failing the single operation.

Now a NULL context throws a catchable TokenException at the JSS/NSS boundary, matching PK11Signature's existing behavior.

Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid or expired cipher contexts.
  * Reports a clear token-related error instead of continuing with an unavailable context.


edewata (approved):
LGTM

fmarco76:
@edewata Thanks!

PR #1115, opened 2026-08-17
Fix TLS 1.3 post-handshake authentication hanging at 100% CPU

t-woerner:
Fix TLS 1.3 post-handshake authentication hanging at 100% CPU
    
    When a TLS 1.3 server sends a post-handshake CertificateRequest to a
    JSS client, the unwrap() do-while loop spins indefinitely. PR.Read()
    processes the request and queues the Certificate response in write_buf,
    but the loop never breaks to let the caller call wrap() to flush it.
    
    In JSSEngineReferenceImpl, detect pending write data after a completed
    handshake and signal NEED_WRAP to break the loop.
    
    In JSSSocketChannel, add flushPostHandshake() to drive wrap() and
    send the Certificate response. Handle non-blocking channels by
    preserving pending output when writeChannel.write() returns 0 and
    retrying on the next read() or write() call.
    
    Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=2490607
    
    Signed-off-by: Thomas Woerner <twoerner@redhat.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TLS 1.3 post-handshake authentication handling.
  * Ensured generated authentication responses are sent correctly during secure channel reads.
  * Improved handshake processing for socket-based connections requiring additional outbound records.

### DIFF
--- a/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
+++ b/base/src/main/java/org/mozilla/jss/ssl/javax/JSSEngineReferenceImpl.java
@@ -1382,6 +1382,16 @@ public class JSSEngineReferenceImpl extends JSSEngine {
                     seen_exception = true;
                 }
             }
+
+            // TLS 1.3 post-handshake auth: PR.Read() may have processed a
+            // CertificateRequest, putting the Certificate response in write_buf.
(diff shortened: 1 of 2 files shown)


fmarco76:
This commit has been integrated in #1116 which is already merged.

